PR #1, opened 2016-11-05
Improve use of UpdateVCS with git

raimue:
These changes improve handling of a ports tree with git. Exporting HOME and SSH_AUTH_SOCK ensures that git can read its configuration and is able to access SSH keys if necessary.

Review-Request: @jmroot @neverpanic

### DIFF
--- a/src/pextlib1.0/uid.c
+++ b/src/pextlib1.0/uid.c
@@ -202,6 +202,107 @@ int setegidCmd(ClientData clientData UNUSED, Tcl_Interp *interp, int objc, Tcl_O
     return TCL_OK;
 }
 
+/**
+ * wrapper around getpwuid(3)
+ *
+ * getpwuid <uid> [<field>]
+ *
+*/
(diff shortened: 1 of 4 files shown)


src/pextlib1.0/uid.c
```diff
@@ -202,6 +202,107 @@ int setegidCmd(ClientData clientData UNUSED, Tcl_Interp *interp, int objc, Tcl_O
...
 
+/**
+ * wrapper around getpwuid(3)
+ *
+ * getpwuid <uid> [<field>]
+ *
+*/
+int getpwuidCmd(ClientData clientData UNUSED, Tcl_Interp *interp, int objc, Tcl_Obj *CONST objv[]) {
+    uid_t uid;
+    char *field = NULL;
```
jmroot: field should be const?

raimue: Fixed and squashed into commit introducing this function.

PR #3, opened 2016-11-06
Support multi-valued maintainers

neverpanic:
We have two equal implementations of unobscure_maintainers, but only
a single one should exist. Move the implementation to the macports_util
package where it can be used by port(1) and port1.0, adjust uses in
port(1), drop the implementation from port1.0, because it is unused.

Additionally, add a test that verifies that @github-username works as
expected in base. Note that no changes in the code are required to
support this.

Review request: @macports/portmgr, @mojca, @larryv.

### DIFF
--- a/src/macports1.0/macports_util.tcl
+++ b/src/macports1.0/macports_util.tcl
@@ -38,6 +38,56 @@ namespace eval macports_util {
     ###################
     # Private methods #
     ###################
+
+    ##
+    # Given a list of maintainers as recorded in a Portfile, return a list of
+    # lists in [key value ...] format describing all maintainers. Valid keys
+    # are 'email' which denotes a maintainer's email address, 'github', which
+    # preceeds the GitHub username of the maintainer and 'keyword', which
+    # contains a special maintainer keyword such as 'openmaintainer' or
+    # 'nomaintainer'.
+    #
+    # @param list A list of obscured maintainers
+    # @return A list of associative arrays in serialized list format
+    proc unobscure_maintainers {list} {
+        set result {}
+        foreach sublist $list {
+            array set maintainer {}
+            foreach token $sublist {
+                if {[string index $token 0] eq "@"} {
+                    # Strings starting with @ are GitHub usernames
+                    set maintainer(github) [string range $token 1 end]
+                } elseif {[string first "@" $token] >= 0} {
+                    # Other strings that contain @ are plain email addresses
+                    set maintainer(email) $token
+                    continue
+                } elseif {[string first ":" $token] >= 0} {
+                    # Strings that contain a colon are obfuscated email
+                    # addresses
+
+                    # Split at :, assign the first part to $domain, re-assemble
+                    # the rest and assign it to $localpart
+                    set localpart [join [lassign [split $token ":"] domain] ":"]
+                    set maintainer(email) "${localpart}@${domain}"
+                } elseif {$token in {"openmaintainer" "nomaintainer"}} {
+                    # Filter openmaintainer and nomaintainer
+                    set maintainer(keyword) $token
+                } else {
+                    # All other entries must be MacPorts handles
+                    set maintainer(email) "${token}@macports.org"
+                }
+            }
+            set serialized [array get maintainer]
+            array unset maintainer
+            if {[llength $serialized]} {
+                # Filter empty maintainers
+                lappend result $serialized
+            }
+        }
+
+        return $result
+    }
+
     proc method_wrap {name} {
         variable argdefault
     
--- a/src/macports1.0/tests/macports_util.test
+++ b/src/macports1.0/tests/macports_util.test
@@ -30,6 +30,49 @@ test method_wrap {
 } -result "Method wrap successful."
 
 
+test unobscure_maintainers_github {
+    Test macports::unobscure_maintainers with a GitHub handle
+} -body {
+    return [macports::unobscure_maintainers {@github-user}]
+} -result {{github github-user}}
+
+test unobscure_maintainers_email {
+    Test macports::unobscure_maintainers with unobfuscated email addresses
+} -body {
+    return [macports::unobscure_maintainers {localpart@example.com}]
+} -result {{email localpart@example.com}}
+
+test unobscure_maintainers_obfuscated_email {
+    Test macports::unobscure_maintainers with obfuscated email addresses
+} -body {
+    return [macports::unobscure_maintainers {example.com:localpart}]
+} -result {{email localpart@example.com}}
+
+test unobscure_maintainers_handle {
+    Test macports::unobscure_maintainers with MacPorts handles
+} -body {
+    return [macports::unobscure_maintainers {handle}]
+} -result {{email handle@macports.org}}
+
+test unobscure_maintainers_keywords {
+    Test macports::unobscure_maintainers with nomaintainer and openmaintainer keywords
+} -body {
+    return [macports::unobscure_maintainers {nomaintainer openmaintainer}]
+} -result {{keyword nomaintainer} {keyword openmaintainer}}
+
+test unobscure_maintainers_multiple {
+    Test macports::unobscure_maintainers with multiple maintainers
+} -body {
+    return [macports::unobscure_maintainers {@github-user example.com:localpart}]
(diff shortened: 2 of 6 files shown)


neverpanic:
I'm trying to avoid adding more code to macports.tcl where I can. That file is already too long.

raimue:
It was more about the namespace than the file it is in, as this moves macports_util from a collection of internal helpers of macports1.0 to the public API.

neverpanic:
Good point, I guess I should move it then.

jmroot:
Splitting some of macports.tcl into different files that each contain some logical subset isn't a bad idea. Even if they're all just sourced by the main file it would probably still improve readability. But that's a bigger project than this PR.

src/macports1.0/macports_util.tcl
```diff
@@ -35,6 +35,36 @@ package provide macports_util 1.0
...
+            if {[string first "@" $maintainer] < 0} {
+                # No "@" in the maintainer, it's either an obfuscated email, or
+                # a keyword
+
+                set colonpos [string first ":" $maintainer]
+                if {$colonpos >= 0} {
+                    # Cut and re-assemble
+                    set maintainer "[string range $maintainer $colonpos+1 end]@[string range $maintainer 0 $colonpos-1]"
+                } elseif {$maintainer ni {"openmaintainer" "nomaintainer"}} {
+                    set maintainer "$maintainer@macports.org"
```
larryv: Could be
```tcl
append maintainer @macports.org
```

src/macports1.0/macports_util.tcl
```diff
@@ -35,6 +35,36 @@ package provide macports_util 1.0
...
+    ##
+    # Given a list of maintainers as recorded in a Portfile, return an
+    # unobscured version that uses proper email addresses.
+    #
+    # @param list A list of obscured maintainers
+    # @return An unobscured list of maintainers
+    proc unobscure_maintainers {list} {
+        set result {}
+        foreach maintainer $list {
+            if {[string first "@" $maintainer] < 0} {
```
larryv: I would prefer
```tcl
if {[string first @ $maintainer] >= 0]} {
    continue
}
```
so that the rest of the function could be outdented.

src/macports1.0/tests/macports_util.test
```diff
@@ -30,6 +30,25 @@ test method_wrap {
 } -result "Method wrap successful."
 
 
+test unobscure_maintainers {
+    Unobscure maintainers unit test.
+} -body {
+    if {[macports_util::unobscure_maintainers [list "port"]] ne "port@macports.org"} {
```
larryv: Is it necessary to force a pure list here? It's a little noisy 😕

src/macports1.0/macports_util.tcl
```diff
@@ -35,6 +35,36 @@ package provide macports_util 1.0
...
+                # No "@" in the maintainer, it's either an obfuscated email, or
+                # a keyword
+
+                set colonpos [string first ":" $maintainer]
+                if {$colonpos >= 0} {
+                    # Cut and re-assemble
+                    set maintainer "[string range $maintainer $colonpos+1 end]@[string range $maintainer 0 $colonpos-1]"
```
larryv: Could be
```tcl
lassign [split $maintainer :] domain user
if {$user ne ""} {
    set maintainer ${user}@${domain}
}
```
mainly because I think fussing with indices is noisy.

src/macports1.0/macports_util.tcl
jdberry: But in this case that would result in the maintainer not getting into the result list. You could of course add it to the result before continuing.

neverpanic:
I'll refactor this to include @larryv's comments and the discussion from macports/macports-guide#1.

src/macports1.0/macports_util.tcl
neverpanic: `[split ...]` does not have a maximum number of splits, though, so if you want to get this right it would be

```tcl
set user [join [lassign [split $maintainer :] domain] :]
```

which isn't any shorter either. Nevertheless, that's going to be in the next version of this PR.

neverpanic:
Code now offers a public API in `macports::unobscure_maintainers` with the implementation living in `macports_util`. I've also adjusted `port(1)` to deal with the new format.

As an example:
```tcl
maintainers     {@github-1 handle1} \
                {example.com:email2 @github-2} \
                example.org:email3 \
                @github-4 \
                openmaintainer \
                nomaintainer
```

now produces:
```
Maintainers:          Email: handle1@macports.org, GitHub: github-1
                      Email: email2@example.com, GitHub: github-2
                      Email: email3@example.org
                      GitHub: github-4
                      open maintainer
                      no maintainer
```

src/port/port.tcl
```diff
@@ -2171,9 +2155,40 @@ proc action_info { action portlist opts } {
                 set label "$opt: "
             }
 
+            if {$ropt in {"description" "long_description"}} {
```
raimue: This hunk seems unrelated to unobscuring maintainers.

src/port/port.tcl
```diff
@@ -2171,9 +2155,40 @@ proc action_info { action portlist opts } {
...
+                    if {[info exists maintainer(github)]} {
+                        lappend parts "GitHub: $maintainer(github)"
+                    }
+                    if {[info exists maintainer(keyword)]} {
+                        switch $maintainer(keyword) {
+                            nomaintainer {
+                                lappend parts "no maintainer"
+                            }
+                            openmaintainer {
+                                lappend parts "open maintainer"
```
raimue: I am not sure about these. Maybe I am just used to read them as keywords instead of multiple words… Also seems odd that they have no label.

src/port/port.tcl
neverpanic: It is an artifact of the removal of `[join $portinfo($ropt)]` above, which is required because it would otherwise flatten the maintainer list. Not running the `[join ...]` for descriptions breaks ports that use newlines in their descriptions, e.g. `barcode`.

src/port/port.tcl
neverpanic: I don't like the current form either, suggestions welcome. `none` and `Policy: openmaintainer` looks much better to me already.

raimue:
We might also need to change the `maintainer:` pseudo-port selector to search in lists.

neverpanic:
Pushed a version with `Maintainer: none` for nomaintainer and `Policy: openmaintainer` for openmaintainer.

I don't agree with changing the `port info` output when `-v` is passed, because that flag is currently meaningless for `port info`, and I don't feel like this is a compelling case to change that.

As for the pseudo-port selector, we have the exact same problem of searching in lists with other fields already, e.g. `depends_*` or `license`. Here are a couple of corner cases to illustrate what would happen:
```
$ rlwrap port-tclsh
% regexp -- "cal" {{@neverpanic cal} {raimue example.com:localpart}}
1
% regexp -- "@neverpanic" {{@neverpanic cal} {raimue example.com:localpart}}
1
% regexp -- "neverpanic" {{@neverpanic cal} {raimue example.com:localpart}}
1
% regexp -- "raimue" {{@neverpanic cal} {raimue example.com:localpart}}
1
% string match "*raimue*" {{@neverpanic cal} {raimue example.com:localpart}}
1
% string compare "{{@neverpanic cal} {raimue example.com:localpart}}" {{@neverpanic cal} {raimue example.com:localpart}}
1
```

raimue:
`port -v info` already changes the dependency information, although this is something most users would not be interested in.

I still do not like the verbosity in the output of `port info`, especially if the only listed entry is an email address, where the "Email:" label looks out of place.

However, as we agreed on the Portfile  syntax, I think we should merge this now and tune the output later.

mojca:
Will we do a mass commit for this change? I would expect a mass commit that adds github usernames to all developers where we know that info.

raimue:
I would like to see it changed at once, but mass commits are a problem for the buildbot, as we discussed when we wanted to remove the `$Id$` tags.

neverpanic:
Remember that this requires a new base release plus a week that we allow for users to install the new base version, so we have time to figure out how to do this, e.g. by implementing the successcache.

mojca:
We need to fix the parser on the buildbot setup.

PR #5, opened 2016-11-13
Remove svn.macports.org as fallback for distfiles

raimue:
If a port really still needs to use this, they can specify this
explicitly with `master_sites macports`.

Closes: https://trac.macports.org/ticket/52626

Review request: @ryandesign


ryandesign:
Thanks. If you believe this works, I trust you. I don't have the energy to test it.

raimue:
Please squash the commits before merging.

PR #7, opened 2016-11-19
Use a 'catch' wrapper that can ignore signals

jmroot:
The addition of Tclx signal handlers means that SIGINT and SIGTERM
now result in an error being raised. In most places where the catch
command is used, we don't actually check if the error is a signal, and
in most of those cases can't do anything sensible except re-raise the
error anyway.

Thus we replace 'catch' with our own version which encapsulates
checking if the error is a signal and (by default) re-raises it if so.
The -signal flag can be passed as the first argument to the new catch
to make it catch signals too.

The code is sourced from its own file in both the main interpreter
and the slave interpreters that execute Portfiles. It can't simply be
aliased in to the slaves like we normally do for shared commands
because it uses uplevel and aliases add an extra call level.

### DIFF
--- a/src/macports1.0/macports.tcl
+++ b/src/macports1.0/macports.tcl
@@ -41,6 +41,9 @@ package require reclaim 1.0
 package require selfupdate 1.0
 package require Tclx
 
+# catch wrapper shared with port1.0
+package require mpcommon 1.0
+
 namespace eval macports {
     namespace export bootstrap_options user_options portinterp_options open_mports ui_priorities
     variable bootstrap_options "\
--- a/Makefile.in
+++ b/Makefile.in
@@ -17,7 +17,7 @@ include Mk/macports.autoconf.mk
 
 all:: Mk/macports.autoconf.mk
 
-Mk/macports.autoconf.mk: Mk/macports.autoconf.mk.in src/config.h.in Makefile.in doc/Makefile.in src/Makefile.in src/cflib1.0/Makefile.in src/cregistry/Makefile.in src/darwintracelib1.0/Makefile.in src/machista1.0/Makefile.in src/macports1.0/Makefile.in src/package1.0/Makefile.in src/pextlib1.0/Makefile.in src/port/Makefile.in src/port1.0/Makefile.in src/programs/Makefile.in src/programs/daemondo/Makefile.in src/registry2.0/Makefile.in tests/Makefile.in config.status
+Mk/macports.autoconf.mk: Mk/macports.autoconf.mk.in src/config.h.in Makefile.in doc/Makefile.in src/Makefile.in src/cflib1.0/Makefile.in src/cregistry/Makefile.in src/darwintracelib1.0/Makefile.in src/machista1.0/Makefile.in src/macports1.0/Makefile.in src/mpcommon1.0/Makefile.in src/package1.0/Makefile.in src/pextlib1.0/Makefile.in src/port/Makefile.in src/port1.0/Makefile.in src/programs/Makefile.in src/programs/daemondo/Makefile.in src/registry2.0/Makefile.in tests/Makefile.in config.status
 	./config.status
 	${MAKE} clean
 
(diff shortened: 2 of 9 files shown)


jmroot:
@raimue @neverpanic Requesting review please.

src/common/catch.tcl
```diff
@@ -0,0 +1,25 @@
...
+    rename catch builtin_catch
+
+    proc catch {args} {
+        set catch_signal no
+        if {[lindex $args 0] eq "-signal"} {
+            set catch_signal yes
+            set args [lrange $args 1 end]
+        }
+        set err [uplevel 1 [list builtin_catch {*}$args]]
+        if {$err == 1 && !$catch_signal} {
```
raimue: `if {$err && ...}`? It will probably always be TCL_ERROR (== 1), but just to be sure...

src/common/catch.tcl
```diff
@@ -0,0 +1,25 @@
...
+            set savedErrorCode $::errorCode
+            set savedErrorInfo $::errorInfo
+            set sigstring "POSIX SIG "
+            set len [string length $sigstring]
+            if {[string equal -length $len $sigstring $savedErrorCode]} {
+                return -code error -errorinfo $savedErrorInfo -errorcode $savedErrorCode "Aborted: [lindex $savedErrorCode end] signal received"
+            }
+        }
+        return $err
```
raimue: Can we be sure nothing modified `$::errorCode` and `$::errorInfo` (if so, why did we save them)? Should we restore these variables before returning?

src/common/catch.tcl
```diff
@@ -0,0 +1,25 @@
...
+            set args [lrange $args 1 end]
+        }
+        set err [uplevel 1 [list builtin_catch {*}$args]]
+        if {$err == 1 && !$catch_signal} {
+            set savedErrorCode $::errorCode
+            set savedErrorInfo $::errorInfo
+            set sigstring "POSIX SIG "
+            set len [string length $sigstring]
+            if {[string equal -length $len $sigstring $savedErrorCode]} {
+                return -code error -errorinfo $savedErrorInfo -errorcode $savedErrorCode "Aborted: [lindex $savedErrorCode end] signal received"
```
raimue: Will we need the information where the error happened? The values in the `optionsVarName` argument of [catch](https://www.tcl.tk/man/tcl8.5/TclCmd/catch.htm) would no longer match the error returned here. It is probably harmless as we would abort or re-throw the error, but not exactly the behavior of the original catch.

neverpanic:
No comments other than @raimue's.

src/macports1.0/macports.tcl
```diff
@@ -41,6 +41,9 @@ package require reclaim 1.0
 package require selfupdate 1.0
 package require Tclx
 
+# catch wrapper shared with port1.0
+source [file join [file dirname [info script]] ../common/catch.tcl]
```
jmroot: Maybe. What would be the benefit?

src/common/catch.tcl
jmroot: Well if err can be a number other than 1 or 0 then it is not a boolean, so that probably won't work. And we certainly don't want to execute this code if err is something like TCL_BREAK.

src/common/catch.tcl
jmroot: Actually they should match as we saved the info earlier. Catch is executed with uplevel so it sets optionsVarName in the calling context. (Unless I'm missing your point here entirely?)

src/common/catch.tcl
jmroot: Hmm, maybe. I don't think any of the code in between can error though. I saved the variables purely for use in the return of the signal case.

src/common/catch.tcl
raimue: The documentation isn't that clear what the possible return values are, but mentions `::errorInfo` will only be set for TCL_ERROR. Then `1` would be correct.

jmroot:
I've been using this for a while and it seems to work, so I'm going to go ahead and merge.

PR #9, opened 2016-12-05
Merge master into gsoc15-dependency branch using git merge

JacksonIsaac:


raimue:
It looks like you first did a rebase and then a merge? This is really hard to follow and duplicates most of master since the last merge between the branches. It would be better to do this only with a single merge instead of a rebase to avoid the mass amount of "new" commits.

Also the last commit ddc845b seems to be completely new and is not actually part of the merge.

JacksonIsaac:
In this branch, I did a git merge master. 

gsoc15-dependency had 57 commits that were uncommitted to master that were added along with the merge making it 58 commits ahead of master (57 + 1 merge commit).

Last commit ddc845b was because I missed a \ while fixing the merge conflicts, making the current branch ahead by 59 commits. When I tried to run rebase I was getting more commits that this. https://github.com/JacksonIsaac/macports-base/commits/gsoc15-dependency (which I think I might have done incorrectly).

raimue:
Ah, on closer inspection I see that this just adds a merge commit. Very confusing view by GitHub.

PR #12, opened 2016-12-18
Add missing quotes around some Makefile variables

vvs31415:
Put quotes around some variables which can contain special characters. These are quoted in other makefiles.


raimue:
Patch looks good to me. The first line of the commit message could be more telling.

PR #14, opened 2017-01-03
Improve description of -N switch

smithsp:
Note that the use of the phrase "non-interactive mode" for the -N switch is in contradiction to the documentation of the command argument, which it claims if I give the command argument, then I am not in interactive mode already...  Also, searching for "yes" in the man page yields nothing.  I am here providing a pull request with my suggestion for improvements to the description of this switch.

### DIFF
--- a/doc/port.1.txt
+++ b/doc/port.1.txt
@@ -129,7 +129,7 @@ The port command recognizes several global flags and options.
     Quiet mode, suppress informational messages to a minimum, implies -N
 
 -N::
-    Non-interactive mode, interactive questions are not asked 
+    Non-question mode, interactive questions are not asked (avoid typing yes)
 
 .Installation and upgrade
 -n::
--- a/doc/port.1.txt
+++ b/doc/port.1.txt
@@ -129,7 +129,7 @@ The port command recognizes several global flags and options.
     Quiet mode, suppress informational messages to a minimum, implies -N
 
 -N::
-    Non-question mode, interactive questions are not asked (avoid typing yes)
+    No-questions mode, interactive questions are not asked (avoid typing yes)
 
 .Installation and upgrade
 -n::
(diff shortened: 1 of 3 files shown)


doc/port.1.txt
```diff
@@ -129,7 +129,7 @@ The port command recognizes several global flags and options.
     Quiet mode, suppress informational messages to a minimum, implies -N
 
 -N::
-    Non-interactive mode, interactive questions are not asked 
+    Non-question mode, interactive questions are not asked (avoid typing yes)
```
raimue: Shouldn't that be "no-questions"?

The phrase "avoid typing yes" is misleading, as in many cases the operation is aborted instead of choosing the "yes" option.

smithsp: I added the "yes" to the line so that the option could be found.

jmroot:
I'd rather change the existing references to interactive mode, as that is not truly interactive, it's really just batch mode reading from stdin. Maybe call it something like "shell mode".

smithsp:
@jmroot There are 26 instances of "interactive mode" in the base repo.  That is more than I am willing to delve into today.

jmroot:
Resolved by 62b1d86.

doc/port.1.txt
smithsp: @jmroot Thank you for resolving the documentation conflict between shell mode and interactive mode.  However, can you also include the change from the end of this line `(avoid typing yes)`?  I searched `man port` for `yes` when looking for this option, and was unable to find it.

jmroot: I agree with Rainer, this phrasing is not good. As well as the behaviour not being the same as answering yes all the time, there are other types of question than just yes/no.

smithsp: @jmroot Thank you for your reasoning.  I understand better why my suggestion is not OK.

PR #15, opened 2017-01-28
Explain that interactions, and no initial change.

dershow:
No changes are made to the files until the user selects them after running it.


neverpanic:
Thanks.

PR #16, opened 2017-01-31
Fix the detection of /opt/local/bin in user's PATH

barry-scott:
The postflight script was using the SHELL variable, which is root's shell
and not USHELL, which is the user's shell.

The test for MacPorts paths in the users PATH can be made more robust.
The old code will match on /opt/locale for example.

Spit the PATH into one path per line with tr and then grep for an exact match
with /opt/local/bin.


barry-scott:
@jmroot I think you after something like the second commit.

jmroot:
Thanks.

PR #17, opened 2017-02-02
Never run installing user's code as root 

barry-scott:
Replace all cases of ${USHELL} -l with /usr/bin/su commands.
First use is commented explaining each steps purpose.

Use printenv VAR | tail -n 1 to make sure that only the VAR's value is passed on.
It is not uncommon for profiles to print information to stdout.

Refactor the MANPATH logic to use su and the tr|grep idium to look at for a path.

Delete all the logic for DISPLAY as its not required in any supported macOS
version that MacPorts supports.

Replace hardtab with spaces

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -2,6 +2,17 @@
 # This is a log of major user-visible changes in each MacPorts release.
 ###
 
+Release 2.4.1 (unreleased)
+    - Avoid an infinite loop when livecheck.regex is empty.
+      (raimue in d33a8a0)
+
+    - Fixed detection of /opt/local/bin already in the user's PATH in the
+      installer postflight script. (barry-scott in dcb0788)
+
+    - Fixed the display of default variants by 'port variants' when the
+      defaults are affected by variants present in variants.conf.
+      (jmr in 9e63a61)
+
 Release 2.4 (2017-01-27 by jmr)
     - New action 'port reclaim' to clean inactive ports and unnecessary
       distfiles to get back disk space, developed during GSoC 2014.
--- a/ChangeLog
+++ b/ChangeLog
@@ -13,6 +13,10 @@ Release 2.4.1 (unreleased)
       defaults are affected by variants present in variants.conf.
       (jmr in 9e63a61)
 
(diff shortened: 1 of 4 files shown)


jmroot:
It's still possible to build an installer for Tiger (and Ryan may yet do so.)

barry-scott:
Oh. I took that MacPorts web site making avalable only 10.9 and later kits to mean the 10.5 was not needed.

Do you need a change?

mojca:
I cannot comment the contents, but you should avoid merging and use rebase instead.

jmroot:
Yes, please restore the DISPLAY handling. There are packages available for OS versions back to 10.5 (I didn't build one for 10.4 because it stopped working in VirtualBox.)

barry-scott:
@mojca I take it you mean that I should not use:
$ git fetch
$ git merge upstream/master

What do you recommend?

jmroot:
https://trac.macports.org/wiki/WorkingWithGit#updating

tl;dr: We normally do `git pull --rebase`.

mojca:
And potentially
```
[pull]
  rebase = true
```
in `.git/config`.

barry-scott:
DISPLAY handling is back but with the su idiom.

mojca:
... and then you do `git rebase -i HEAD~6` to reorder and/or squash commits, followed by a force push to branch once you are happy (this can in principle also be done by reviewer).

barry-scott:
@mojca Are you suggesting I do that now to affect this pull request?

neverpanic:
@barry-scott Yes, we usually bring the PR to the exact state of commits we want to apply on master before merging.

barry-scott:
I'm not winning against git. I will close this PR and setup my git clone following your recommendations. Then generate a new PR.

neverpanic:
@barry-scott I can also rebase the patches for you, if that helps?

barry-scott:
@neverpanic Thanks for the offer. I did the fork again and setup up to use git merge --rebase so I should have PRs in a cleaner state from now on.

PR #18, opened 2017-02-03
Never run installing user's profile code as root

barry-scott:
Replace all cases of ${USHELL} -l with /usr/bin/su commands.
First use is commented explaining each steps purpose.

Use printenv VAR | tail -n 1 to make sure that only the VAR's value is passed on.
It is not uncommon for profiles to print information to stdout.

Refactor the MANPATH logic to use su and the tr|grep idium to look at for a path.

Delete all the logic for DISPLAY as its not required in any supported macOS
version that MacPorts supports.

Replace hardtab with spaces


neverpanic:
Thank you! Merged.

PR #19, opened 2017-02-16
misc: let 'port provides' and trace mode honor the FS case-sensitivity.

Ionic:
While updating `ghc` in trace mode, I noticed it failing while including `"GC.h"`. `ghc` actually meant to include a file named `GC.h` from its own source tree, but due to providing `-I${prefix}` before include directives for its own source tree, `/opt/local/include/gc.h` was picked by the compiler.

On a case-sensitive system, this wouldn't have happened.

Ultimately, the build process was able to read `/opt/local/include/GC.h`, even though `ghc` did *not* depend upon `boehmgc`, but eventually failed to build as it wasn't able to read `/opt/local/include/gc/gc.h`. Curiously, both files *should* have belonged to the same port (`bohemgc`.) However, `port provides /opt/local/include/GC.h` returned no registered port for this file - which was also queried as such by trace mode and thus allowed access. Checking `port provided /opt/local/include/gc.h` correctly returned `boehmgc`.

This PR tried to fix up this shortcoming by respecting the FS's underlying case-sensitivity.

`ghc` built fine for my in trace mode with it.

Because this changeset is a bit bigger than others and there's breaking potential, please review.

### DIFF
--- a/src/cregistry/entry.c
+++ b/src/cregistry/entry.c
@@ -1,7 +1,7 @@
 /*
  * entry.c
  *
- * Copyright (c) 2010-2011, 2014 The MacPorts Project
+ * Copyright (c) 2010-2011, 2014, 2017 The MacPorts Project
  * Copyright (c) 2007 Chris Pickel <sfiera@macports.org>
  * All rights reserved.
  *
@@ -39,6 +39,7 @@
 #include <sqlite3.h>
 #include <stdlib.h>
 #include <string.h>
+#include <stdio.h>
 
 /*
  * TODO: possibly, allow reg_entry_search to take different matching strategies
@@ -517,16 +518,31 @@ int reg_entry_installed(reg_registry* reg, char* name, reg_entry*** entries,
  *
  * @param [in] reg     registry to search in
  * @param [in] path    path of the file to check ownership of
+ * @param [in] cs      false if check should be performed case-insensitive,
+ *                     true otherwise
  * @param [out] entry  the owner, or NULL if no active port owns the file
  * @param [out] errPtr on error, a description of the error that occurred
  * @return             true if success; false if failure
  */
-int reg_entry_owner(reg_registry* reg, char* path, reg_entry** entry,
+int reg_entry_owner(reg_registry* reg, char* path, int cs, reg_entry** entry,
         reg_error* errPtr) {
     int result = 0;
     sqlite3_stmt* stmt = NULL;
-    char* query = "SELECT id FROM registry.files WHERE actual_path=? AND active";
     int lower_bound = 0;
+
+    /* Make sure query_template is as big as the buffer might get. */
+    const char* query_template = "SELECT id FROM registry.files WHERE (actual_path = ? COLLATE NOCASE) AND active";
+    const int query_max_size = strlen(query_template) + 1;
+
+    char* query = malloc(query_max_size);
+
+    if (!query) {
+        return 0;
+    }
+
+    snprintf(query, query_max_size, "SELECT id FROM registry.files WHERE (actual_path = ? %s) AND active",
+             cs ? "" : "COLLATE NOCASE");
+
     if ((sqlite3_prepare_v2(reg->db, query, -1, &stmt, NULL) == SQLITE_OK)
             && (sqlite3_bind_text(stmt, 1, path, -1, SQLITE_STATIC)
                 == SQLITE_OK)) {
@@ -555,6 +571,7 @@ int reg_entry_owner(reg_registry* reg, char* path, reg_entry** entry,
     if (stmt) {
         sqlite3_finalize(stmt);
     }
+    free(query);
     return result;
 }
 
@@ -568,12 +585,27 @@ int reg_entry_owner(reg_registry* reg, char* path, reg_entry** entry,
  *
  * @param [in] reg  registry to find file in
  * @param [in] path path of file to get owner of
+ * @param [in] cs   false if check should be performed case-insensitive,
+ *                  true otherwise
  * @return          id of owner, or 0 for none
  */
-sqlite_int64 reg_entry_owner_id(reg_registry* reg, char* path) {
+sqlite_int64 reg_entry_owner_id(reg_registry* reg, char* path, int cs) {
     sqlite3_stmt* stmt = NULL;
     sqlite_int64 result = 0;
-    char* query = "SELECT id FROM registry.files WHERE actual_path=? AND active";
+
+    /* Make sure query_template is as big as the buffer might get. */
+    const char* query_template = "SELECT id FROM registry.files WHERE (actual_path = ? COLLATE NOCASE) AND active";
+    const int query_max_size = strlen(query_template) + 1;
+
+    char* query = malloc(query_max_size);
+
+    if (!query) {
+        return 0;
+    }
+
+    snprintf(query, query_max_size, "SELECT id FROM registry.files WHERE (actual_path = ? %s) AND active",
+             cs ? "" : "COLLATE NOCASE");
+
     if ((sqlite3_prepare_v2(reg->db, query, -1, &stmt, NULL) == SQLITE_OK)
             && (sqlite3_bind_text(stmt, 1, path, -1, SQLITE_STATIC)
                 == SQLITE_OK)) {
@@ -588,6 +620,7 @@ sqlite_int64 reg_entry_owner_id(reg_registry* reg, char* path) {
     if (stmt) {
         sqlite3_finalize(stmt);
     }
+    free(query);
     return result;
 }
 
--- a/src/pextlib1.0/Pextlib.c
+++ b/src/pextlib1.0/Pextlib.c
@@ -4,7 +4,7 @@
  * Copyright (c) 2002 - 2003 Apple Inc.
  * Copyright (c) 2004 - 2005 Paul Guyot <pguyot@kallisys.net>
  * Copyright (c) 2004 Landon Fuller <landonf@macports.org>
- * Copyright (c) 2007 - 2016 The MacPorts Project
+ * Copyright (c) 2007 - 2017 The MacPorts Project
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -52,6 +52,7 @@
 #include <sys/socket.h>
 #include <sys/stat.h>
 #include <sys/wait.h>
+#include <sys/param.h>
 #include <ctype.h>
 #include <errno.h>
 #include <fcntl.h>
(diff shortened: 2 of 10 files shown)


src/pextlib1.0/tracelib.c
```diff
@@ -614,10 +616,48 @@ static void dep_check(int sock, char *path) {
         answer(sock, "#");
     }
 
+    if (-1 == fs_cs) {
```
neverpanic: This may differ per path. Some people may have random mount points within `/opt/local`, so it's probably not a very good idea to cache this.

src/pextlib1.0/tracelib.c
```diff
@@ -614,10 +616,48 @@ static void dep_check(int sock, char *path) {
...
+            if (path_stat.st_ino == lowercase_path_stat.st_ino) {
+                /* Truly case-insensitive! */
+                fs_cs = 0;
+            }
+            else {
+                /* Just similar-named, still a case-sensitive FS. */
+                fs_cs = 1;
+            }
+        }
+    }
```
neverpanic: I don't like the idea of manually fiddling with the case here. Couldn't we use `getattrlist(2)` to do this, maybe using this:

```c
#include <stdio.h>
#include <string.h>
#include <sys/attr.h>
#include <sys/mount.h>
#include <sys/param.h>
#include <unistd.h>

typedef struct volcaps {
	u_int32_t size;
	vol_capabilities_attr_t volcaps;
} volcaps_t;

int fs_case_sensitive(char *path) {
	struct statfs f;
	struct attrlist attrlist;
	volcaps_t volcaps;

	if (-1 == statfs(path, &f)) {
		/* failed to get mountpoint for a file; maybe the file doesn't exist */
		return 1;
	}

	memset(&attrlist, 0, sizeof(attrlist));
	attrlist.bitmapcount = ATTR_BIT_MAP_COUNT;
	attrlist.volattr = ATTR_VOL_CAPABILITIES;

	if (-1 == getattrlist(f.f_mntonname, &attrlist, &volcaps, sizeof(volcaps), 0)) {
		/* we don't know anything, assume case sensitive */
		return 1;
	}

	if ((attrlist.volattr & ATTR_VOL_CAPABILITIES) == 0) {
		/* failed to get mountpoint capabilities */
		return 1;
	}
	if ((volcaps.volcaps.valid[VOL_CAPABILITIES_FORMAT] & VOL_CAP_FMT_CASE_SENSITIVE) == 0) {
		/* capabilities bit for case-sensitivity not valid */
		return 1;
	}
	return (volcaps.volcaps.capabilities[VOL_CAPABILITIES_FORMAT] & VOL_CAP_FMT_CASE_SENSITIVE) > 0;
}

```
and then just use its return value?

src/registry2.0/registry.tcl
```diff
@@ -236,7 +236,33 @@ proc open_file_map {args} {
...
+		# Look at the inode number to distinguish these files.
+		if {![catch {::file lstat $file lstatret_orig}]} {
+			if {![catch {::file lstat $lowercase_file lstatret_lower}]} {
+				if {$lstatret_orig(ino) == $lstatret_lower(ino)} {
+					# File systems not supporting inodes, like FAT,
+					# will also land here. Luckily, FAT is mostly cis...
+					# If necessary, we could implement further checks
+					# based upon [file system], but refrain from this
+					# for now.
+					set cs false
```
neverpanic: Not quite sure what to do here. We could add the function to pextlib and export it to Tcl, and just call it from here (and then just have it return 1 on !macOS)

src/cregistry/entry.c
```diff
@@ -517,16 +517,45 @@ int reg_entry_installed(reg_registry* reg, char* name, reg_entry*** entries,
...
+        strncat(query, cis_operator, query_size - strlen(query) - 1);
+    }
+    else {
+        strncat(query, cs_operator, query_size - strlen(query) -1);
+    }
+    strncat(query, base_query_mid, query_size - strlen(query) - 1);
+    strncat(query, base_query_end, query_size - strlen(query) - 1);
+    if (!cs) {
+        strncat(query, extend_query, query_size - strlen(query) - 1);
+    }
```
neverpanic: Well great, I had a comment here that somehow vanished.

I don't like the `strncat` funny business. Wouldn't using `snprintf` or `asprintf` be much shorter and more readable?

Consider:
```c
char *query;
if (-1 == asprintf(&ret, "SELECT id FROM registry.files WHERE actual_path %s ? AND active", cs ? "=" : "LIKE")) {
    return 0;
}

/* or with COLLATE NOCASE */
asprintf(&ret, "SELECT id FROM registry.files WHERE actual_path %s ? AND active %s", cs ? "=" : "LIKE", cs ? "" : "COLLATE NOCASE");

/* sqlite3_prepare_v2(...) */

free(query);
```

Or, if you want to avoid the non-standard `asprintf` (even though I think we already use it in MacPorts) or the `malloc`:
```c
char query[strlen("SELECT id FROM registry.files WHERE actual_path LIKE ? AND active COLLATE NOCASE") + 1];
snprintf(query, sizeof(query), "SELECT id FROM registry.files WHERE actual_path %s ? AND active %s",
         cs ? "=" : "LIKE", cs ? "" : "COLLATE NOCASE");
```

src/pextlib1.0/tracelib.c
Ionic: I thought that maybe caching it once would be good enough to not slow down stuff each time a file is checked. You're right, though, there could be random mountpoints. Even worse - what if we're checking a file installed in the MacPorts applications dir, which might be on a separate partition than `$[prefix}`? Seems like there's no way around doing the check each time a file is accessed in the first place - short of caching it per-file. Since caching is generally missing in tracelib, I guess checking every time is the best we can do for now.

Speaking of memory leaks, here's another...

src/cregistry/entry.c
```diff
@@ -517,16 +517,45 @@ int reg_entry_installed(reg_registry* reg, char* name, reg_entry*** entries,
...
+    const char* base_query_start = "SELECT id FROM registry.files WHERE actual_path";
+    const char* base_query_mid = "?";
+    const char* base_query_end = " AND active";
+    const char* cis_operator = " LIKE ";
+    const char* cs_operator = "=";
+    const char* extend_query = " COLLATE NOCASE";
+    const int query_size = strlen(base_query_start) + strlen(cs_operator)
+                           + strlen(cis_operator) + strlen(base_query_mid)
+                           + strlen(base_query_end) + strlen(extend_query) + 1;
+    char* query = malloc(query_size);
```
Ionic: I did, thanks.

src/cregistry/entry.c
```diff
@@ -517,16 +517,45 @@ int reg_entry_installed(reg_registry* reg, char* name, reg_entry*** entries,
...
     int result = 0;
     sqlite3_stmt* stmt = NULL;
-    char* query = "SELECT id FROM registry.files WHERE actual_path=? AND active";
     int lower_bound = 0;
+    const char* base_query_start = "SELECT id FROM registry.files WHERE actual_path";
+    const char* base_query_mid = "?";
+    const char* base_query_end = " AND active";
+    const char* cis_operator = " LIKE ";
+    const char* cs_operator = "=";
+    const char* extend_query = " COLLATE NOCASE";
```
Ionic: It works, but it doesn't help either. For some reason `COLLATE NOCASE` doesn't seem to change anything.

I've played around a bit with the sqlite3 binary on my 10.6 VM directly and it seems that "COLLATE NOCASE" must be carefully placed to have an effect.

Case in point:
```
sqlite> SELECT id from files WHERE actual_path="/opt/local/bin/gsed" AND active COLLATE NOCASE;
50
sqlite> SELECT id from files WHERE actual_path="/opt/local/bin/gseD" AND active COLLATE NOCASE;
sqlite> SELECT id from files WHERE actual_path="/opt/local/bin/gseD" COLLATE NOCASE AND active;
50
sqlite> SELECT id from files WHERE UPPER(actual_path)="/opt/local/bin/gseD" COLLATE NOCASE AND active;
50
sqlite> SELECT id from files WHERE (UPPER(actual_path)="/opt/local/bin/gseD" COLLATE NOCASE) AND active;
50
```

But if we're using `LIKE` anyway, I can drop the COLLATE NOCASE thingy.

Ionic: Oh, apropos `LIKE`, do we actually want to use it? What if a path actually contains characters like `%` and `_`? Won't that lead to "interesting behavior"?

Given that both `COLLATE NOCASE` and `LIKE` perform a case-insensitive check for ASCII characters only, there isn't a lot of functional difference. If the ICU extension is available and loaded, `LIKE` can also lowercase non-ASCII characters, but I doubt Apple would build sqlite3 in that way. And since we don't have any control about the system sqlite3, it's probably better to assume that both `COLLATE NOCASE` and `LIKE` will only mangle pure ASCII characters.

src/cregistry/entry.c
```diff
@@ -517,16 +517,45 @@ int reg_entry_installed(reg_registry* reg, char* name, reg_entry*** entries,
...
+    const int query_size = strlen(base_query_start) + strlen(cs_operator)
+                           + strlen(cis_operator) + strlen(base_query_mid)
+                           + strlen(base_query_end) + strlen(extend_query) + 1;
```
Ionic: Yeah, but a few more chars in the buffer won't make a huge difference. And it's a bit less error-prone if someone messes this stuff up...

src/cregistry/entry.c
Ionic: We're using `asprintf` alright, but only in `pextlib` and `machista`. We can't do that in `cregistry`, because it has to work cross-platform (think Linux.)

Ported it over to `snprintf` instead.

src/cregistry/entry.c
Ionic: Also, that part will be gone in the next iteration.

src/pextlib1.0/tracelib.c
Ionic: Yeah, could do. I found `getattrlist` while searching around and it might be good enough for `pextlib` and anything that only ever gets built and executed on OS X, but that's about it. Lowercasing the path and checking... which reminds me, we also have to check the uppercase version in case the original path happens to be lowercased in the first place (which is mostly the case!)

The best way to do this is probably by using a combination of `getattrlist` on Darwin and (either if it's not Darwin or `getattrlist` failed) a fallback that does the lowercase-uppercase dance otherwise.

Ionic:
New changes:
  - memory leak fixes
  - replaced `strcat`/`strcpy` indirection with `snprintf`
  - spawned off a new function in tracelib `fs_case_sensitive` (also in the TCL part)
  - checks for uppercase variant as well
  - faulty one-time caching of FS CS value

Not fixed yet/pending discussion:
  - `LIKE` vs. `COLLATE NOCASE` (currently using both)
  - duplication in TCL and C code for FS CS detection
  - not using `getattrlist`

src/pextlib1.0/tracelib.c
```diff
@@ -584,6 +585,104 @@ static void sandbox_violation(int sock UNUSED, const char *path, sandbox_violati
...
+    struct stat path_stat = { 0 },
+           lowercase_path_stat = { 0 },
+           uppercase_path_stat = { 0 };
+
+    if (-1 == lstat(path, &path_stat)) {
+        free(lowercase_path);
+        free(uppercase_path);
+
+        ui_error(interp, "unable to get information on file %s", path);
+        answer(sock, "#");
```
neverpanic: I think you should answer `?` for non-existant files. `#` will cause the client to `abort(3)`, which may have strange side effects.

Ionic:
New changes:
  - checking functions moved into `Pextlib.c`, with a TCL interface that can be called using `[fs_case_sensitive $file]`
  - split up into TCL command function, `fs_case_sensitive_darwin()` that uses `getattrlist()` and `fs_case_sensitive_fallback()` that uses multiple `lstat()` calls
  - TCL command function and tracelib default to `fs_case_sensitive_darwin()` if `__APPLE__`, then call `fs_case_sensitive_fallback()` if either the first check failed or not `__APPLE__`
  - removed TCL duplication of `lstat()` algorithm
  - introduced new sqlite index on `actual_path` with `COLLATE NOCASE`, bumped up internal database version number - operations finally are FAST again!
  - removed `LIKE` operator usage, as we don't want to go out of our ways to escape `%` and `_` in path values... now always using the equality operator with `COLLATE NOCASE`
  - some minor copyright notice updates

Ionic:
I'm pondering whether it would be beneficial to implement some sort of caching. Not exactly per-file, but at least per-mount-point. That would help to avoid the probably costly `getattrlist()` and `lstat()` calls on subsequent calls for files on the same volume.

Doing this with a static/fixed cache of 50 entries and non-eviction would be rather easy, but I'm not sure how to free memory again at the end of execution. Do you think `atexit()` would be good enough for that? Linux/glibc allow using `atexit()` in a shared library and functions will be called when the shared library is unloaded, which is the case when tclsh shuts down (hopefully not earlier...) - but I don't know how OS X handles that.

I'll probably give that a shot locally and see if it works.

Ionic:
New changes:

  - don't error out in flatfile registry mode, only issue a warning that case-sensitive search is not implemented
  - added proper description to first commit implementing the case-sensitivity feature
  - add mount point case-sensitivity caching
  - caching only active on BSD-like systems (check commit message why not on Linux)

Ionic:
Rebased commits a tiny bit to add a few string checks where necessary and make it more resilient - after Pextlib blew up during a cmake run.

src/pextlib1.0/Pextlib.c
```diff
@@ -649,10 +651,149 @@ int SetMaxOpenFilesCmd(ClientData clientData UNUSED, Tcl_Interp *interp, int obj
...
+
+        if (!new_data) {
+            ui_info(interp, "pextlib: unable to reallocate FS cache entries list, leaving untouched.");
+            return ret;
+        }
+
+        cache->entries = new_data;
+
+        mount_cs_cache_entry_t *new_entry = malloc(sizeof(mount_cs_cache_entry_t));
+        new_entry->mountpoint = strdup(mountpoint);
```
neverpanic: Strictly speaking, you should probably also have error handling for this allocation.

src/pextlib1.0/Pextlib.c
```diff
@@ -663,38 +804,52 @@ typedef struct volcaps {
...
     }
 
+    struct attrlist attrlist = { 0 };
+    volcaps_t volcaps = { 0 };
+
     attrlist.bitmapcount = ATTR_BIT_MAP_COUNT;
     attrlist.volattr = ATTR_VOL_CAPABILITIES;
 
-    if (-1 == getattrlist(f.f_mntonname, &attrlist, &volcaps, sizeof(volcaps), 0)) {
+    if (-1 == getattrlist(mntpoint, &attrlist, &volcaps, sizeof(volcaps), 0)) {
```
neverpanic: Shouldn't you first check `lookup_mount_cs_cache()` before calling the potentially more expensive `getattrlist(2)`?

src/pextlib1.0/Pextlib.c
```diff
@@ -648,6 +651,424 @@ int SetMaxOpenFilesCmd(ClientData clientData UNUSED, Tcl_Interp *interp, int obj
...
+
+        free(cache->entries);
+        cache->entries = NULL;
+    }
+}
+
+/**
+ * Rollback mount cache object.
+ */
+static void rollback_mount_cs_cache(Tcl_Interp *interp, mount_cs_cache_t *cache) {
```
neverpanic: I wouldn't have gone the extra mile to shrink this again on failed allocations. The bit of extra unused memory is not a problem (it's neither lost nor unreachable, nor will it be used uninitialized) when `cache->count` isn't increased.

Now that you have it, let's keep it.

src/pextlib1.0/Pextlib.c
```diff
@@ -648,6 +651,424 @@ int SetMaxOpenFilesCmd(ClientData clientData UNUSED, Tcl_Interp *interp, int obj
...
+
+    attrlist.bitmapcount = ATTR_BIT_MAP_COUNT;
+    attrlist.volattr = ATTR_VOL_CAPABILITIES;
+
+    if (-1 == getattrlist(mntpoint, &attrlist, &volcaps, sizeof(volcaps), 0)) {
+        free(mntpoint);
+        return ret;
+    }
+
+    if (-1 == ret) {
```
neverpanic: `ret` will always be -1 here. There is no code path that could make it a different value.

Ionic:
I kind of lost track of this. However, I also had the opportunity to test this stuff with thousands of ports.

"Merging".

PR #20, opened 2017-02-17
Make uninstalling unnecessary ports easy

danchr:
`port reclaim` is quite nice for cleaning up waste, but it doesn't do anything about ports you no longer need, such as build dependancies — like `apt-get` and its `autoremove`. To work around this, I'd occasionally do repeated `port uninstall leaves` until no such leaves were left — which is a bit silly…

This pull request attempts to address this in two different ways. The first commit adds an `rleaves` alias which identifies all unrequested ports. The seconds commit achieves the same within `port reclaim`.

I considered extending `leaves` to list _all_ proper leaves — but opted not to, in the interest of backwards compatibility.

### DIFF
--- a/doc/port-reclaim.1
+++ b/doc/port-reclaim.1
@@ -1,5 +1,5 @@
 '\" t
-.TH "PORT\-RECLAIM" "1" "2017\-01\-28" "MacPorts 2\&.4\&.99" "MacPorts Manual"
+.TH "PORT\-RECLAIM" "1" "2017\-01\-27" "MacPorts 2\&.4\&.99" "MacPorts Manual"
 .\" -----------------------------------------------------------------
 .\" * Define some portability stuff
 .\" -----------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


danchr:
> Looks good to me at a first glance.
>
> I assume you tested this?

Yeah, I did my best to optimise `reclaim` and ensure that it works as expected. `rleaves` isn't quite as fast, due to lacking access to `sort_portlist_by_dependendents`. From my perspective, the main risk is that it'll expose bugs like this:

* User installs some port A depending on port B.
* User installs port C, from source, which has no declared dependency on port B.
* During its `configure` stage, port C picks up port B and links against it during build.
* User uninstalls port A.
* User later runs `port reclaim` which uninstalls port B, breaking port C.
* `port rev-upgrade` notices the breakage and reinstalls port C.

I didn't actually test this, or run into it, since I tend to install all ports using trace mode. (Shouldn't that be the default by now, anyway?) Still, this is likely to expose bugs like that — and speaking of bugs, I should probably add a run of `rev-upgrade` within `reclaim` to catch them.

ryandesign:
That's fine. If it exposes dependency bugs in ports, then those bugs can be fixed.

Trace mode didn't really work until the release of MacPorts 2.4.0 a few weeks ago. Assuming we get no bug reports about it now, it could be come default, but last I heard it still had a 50% performance penalty, so maybe that's not so great. In any case, not relevant to this pull request and can be discussed elsehwere.

PR #21, opened 2017-02-27
livecheck: add -nocase to ignore all cases in regexp

kurthindenburg:
For livecheck regex all we are interested in is the version numbers.  To avoid to have to match all cases in livecheck.regex, this patch adds -nocase.  There are many ports where the name/distname/website have different cases which makes livecheck time-consuming with adding non-default regexps.


raimue:
Should we make this configurable with the default being case-insensitive? If the `(?-i)` modifier works to get case-sensitive matching back if ever needed, that should also be enough.

PR #25, opened 2017-07-06
New argument port fetch --no-mirrors

raimue:
This flag avoids adding the mirrors defined by global_mirror_site to the
list of fetch URLs. This is meant to aid in development of port updates.
No mirror will have a distfile for a new port version, so there is no
need to try.

### DIFF
--- a/src/port1.0/portfetch.tcl
+++ b/src/port1.0/portfetch.tcl
@@ -272,12 +272,17 @@ proc portfetch::get_full_mirror_sites_path {} {
 # Perform the full checksites/checkpatchfiles/checkdistfiles sequence.
 # This method is used by distcheck target.
 proc portfetch::checkfiles {urls} {
-    global global_mirror_site
+    global global_mirror_site ports_fetch_no-mirrors
     upvar $urls fetch_urls
 
-    checksites [list patch_sites [list $global_mirror_site PATCH_SITE_LOCAL] \
-                master_sites [list $global_mirror_site MASTER_SITE_LOCAL]] \
-               [get_full_mirror_sites_path]
+    set sites [list patch_sites {} \
+                    master_sites {}]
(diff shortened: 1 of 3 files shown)


src/port1.0/portfetch.tcl
```diff
@@ -272,12 +272,17 @@ proc portfetch::get_full_mirror_sites_path {} {
...
     upvar $urls fetch_urls
 
-    checksites [list patch_sites [list $global_mirror_site PATCH_SITE_LOCAL] \
-                master_sites [list $global_mirror_site MASTER_SITE_LOCAL]] \
-               [get_full_mirror_sites_path]
+    set sites [list patch_sites {} \
+                    master_sites {}]
+    if {![info exists ports_fetch_no-mirrors] || ${ports_fetch_no-mirrors} eq "no"} {
+        set sites [list patch_sites [list $global_mirror_site PATCH_SITE_LOCAL] \
+                        master_sites [list $global_mirror_site MASTER_SITE_LOCAL]]
```
raimue: Should PATCH_SITE_LOCAL and MASTER_SITE_LOCAL still be kept?

neverpanic: We've never documented them to my knowledge, and they're probably not used very often. On the other hand, we did add an `ARCHIVE_SITE_LOCAL` when we implemented archive support, so be sure to also remove that if you're going to drop them.

neverpanic: And just in case you mean keeping them when `--no-mirrors` is set, I'd say let's keep them – I'd expect local mirrors to be very fast.

mf2k:
I have a test case where I would like to use this feature. Please merge!

neverpanic:
I'm merging this. We can still change behavior later if we decide it should be different.

PR #26, opened 2017-07-15
portlint: add warning(s) about modeline

kurthindenburg:
closes https://trac.macports.org/ticket/52767


raimue:
The guide says "Its use is optional and up to the port maintainer.". If we always expect the modeline, we should change the policy first.

kurthindenburg:
See #33

PR #28, opened 2017-07-21
Abort on unknown subport

kurthindenburg:
Check that given command line subport exists, otherwise abort.

Example: sudo port install zlib subport=foo
         Error: Port zlib does not have a subport foo

closes https://trac.macports.org/ticket/34619

### DIFF
--- a/src/macports1.0/macports.tcl
+++ b/src/macports1.0/macports.tcl
@@ -1761,6 +1761,29 @@ proc mportopen {porturl {options {}} {variations {}} {nocache {}}} {
 
     $workername eval {source Portfile}
 
+    set subport [$workername eval {set subport}]
+    set name [$workername eval {set name}]
+
+    ui_debug "name=$name; subport=$subport"
+    # Verify subport is actually a 'subport' of port
+    if {$name ne $subport} {
+        if {[catch {mportlookup $name} result]} {
+            ui_debug "$::errorInfo"
+            error "lookup of portname $name failed: $result"
+        }
+        if {[llength $result] < 2} {
+            error "$name not found"
+        }
+        array set portinfo [lindex $result 1]
+        if {[info exists portinfo(subports)]} {
+            if {[lsearch $portinfo(subports) $subport] < 0} {
+                error "$name does not have a subport $subport"
+            }
+        } else {
+            error "$name does not have any subports"
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


ryandesign:
Cool! Thanks!

raimue:
Could you add the same check also to action_info, action_notes, action_deps, etc.? Or even better, let's move this into its own proc so we can avoid the code duplication.

kurthindenburg:
% sudo port fetch zlib subport=zlib
Error: Port zlib does not have a subport zlib

Yes, I saw this as well - however, 'port info --subports zlib' shows no subports, so I think this is correct behavoir.

Also, you are correct, it doesn't work when not specifying the port (ie using folder for port).

kurthindenburg:
I'll update my 2nd attempt shortly for comments - I will have to test more - some of this I'm unclear if this is the best way to do this.

Currently, "port info zlib subport=zlib" will work even though technically zlib has no subports

Also, "subports=" appear to be ignored for uninstall

jmroot:
With uninstall the correct subport is always taken from the registry.

Getting mportlookup involved is not viable, because Portfiles don't have to be indexed (and in fact indexing them involves running mportopen on them.) I would do the check like this:
```diff
diff --git a/src/macports1.0/macports.tcl b/src/macports1.0/macports.tcl
index a7223fff..45b5c015 100644
--- a/src/macports1.0/macports.tcl
+++ b/src/macports1.0/macports.tcl
@@ -1773,7 +1773,19 @@ proc mportopen {porturl {options {}} {variations {}} {nocache {}}} {
 
     $workername eval {port::run_callbacks}
 
-    ditem_key $mport provides [$workername eval {set subport}]
+    set actual_subport [$workername eval {set PortInfo(name)}]
+    if {[$workername eval {info exists user_options(subport)}]} {
+        # The supplied subport may have been set on the command line by the
+        # user, or simply obtained from the PortIndex or registry. Check that
+        # it's valid in case the user made a mistake.
+        set supplied_subport [$workername eval {set user_options(subport)}]
+        if {$supplied_subport ne $actual_subport} {
+            set portname [$workername eval {set name}]
+            mportclose $mport
+            error "$portname does not have a subport '$supplied_subport'"
+        }
+    }
+    ditem_key $mport provides $actual_subport
 
     return $mport
 }
```

PR #29, opened 2017-07-29
Remove actions that make no sense on command line

kurthindenburg:
Remove cd, quit, exit from command line actions

closes https://trac.macports.org/ticket/32866

### DIFF
--- a/src/port/port.tcl
+++ b/src/port/port.tcl
@@ -4368,7 +4368,14 @@ array set action_array [list \
 # Expand "action".
 # Returns an action proc, or a list of matching action procs, or the action passed in
 proc find_action { action } {
-    global action_array
+    global action_array ui_options
+
+    # Remove commands that have no use on command line
+    if {![info exists ui_options(ports_commandfiles)]} {
+        array unset action_array cd
+        array unset action_array exit
+        array unset action_array quit
+    }
 
     if { ! [info exists action_array($action)] } {
         set guess [guess_action $action]


jmroot:
This will run every time find_action is called, and modifies the global action_array which is used in other places like print_help. I would do it like this instead:
```diff
diff --git a/src/port/port.tcl b/src/port/port.tcl
index a70c975d..8a1494d7 100755
--- a/src/port/port.tcl
+++ b/src/port/port.tcl
@@ -4371,7 +4371,17 @@ proc find_action { action } {
     global action_array
 
     if {![info exists action_array($action)]} {
-        return [array names action_array [string tolower $action]*]
+        # list of actions that are valid for this mode
+        global action_list
+        if {![info exists action_list]} {
+            global ui_options
+            if {![info exists ui_options(ports_commandfiles)]} {
+                set action_list [lsearch -regexp -all -inline -not [array names action_array] {^quit$|^exit$|^cd$}]
+            } else {
+                set action_list [array names action_array]
+            }
+        }
+        return [lsearch -glob -inline -all $action_list [string tolower $action]*]
     }
 
     return $action

```

raimue:
I think it would make more sense to have a flag in `action_array` for that instead of hardcoding the names somewhere else.

jmroot:
Fair point, though storing the information as a flag in action_array may make it more complicated to extract the action names we want. Would a shellmode_action_list defined next to action_array be an acceptable compromise?

PR #30, opened 2017-07-30
Add port provides option -q to only output port

kurthindenburg:
closes: https://trac.macports.org/ticket/38428


raimue:
Looks good.

If this is meant for scripting, then the other case "… is not provided by a MacPorts port." should maybe print an empty line with `-q`?

jmroot:
This branch seems to have some duplicate and unrelated commits.

kurthindenburg:
Yea, I'm having issues w/ github and PRs :-(  - I'll commit from a fresh patch.

PR #31, opened 2017-08-02
Don't abort if the current folder doesn't exist, use user home.

kurthindenburg:
closes https://trac.macports.org/ticket/40035


kurthindenburg:
I notice none of the other uses of [pwd] are "caught" for errors.

jmroot:
Also the original pwd (if valid) needs to be saved for use in get_current_port. That's another good reason not to use the home dir, just in case it contains a Portfile.

raimue:
macports1.0 should not mess with the current working directory on loading. For commands like `portindex` that would be unexpected and could make them work, when they should have failed. I think it is better to solve this in the port client.

In most cases, we do not need to initialize `$current_portdir` on startup and that would avoid the problem. We should delay getting cwd until it is actually needed in `proc process_command_file` or `proc process_cmd`). However, we need to make sure `[pwd]` is only evaluated once to get the path the command was started in.

jmroot:
I see your point, but as Kurt pointed out, there are a lot of places throughout the code base where [pwd] is used assuming it won't throw an error (usually in order to cd somewhere else and then cd back afterwards). All of these need to be fixed to really address the issue the ticket brings up.

pmetzger:
Should we close this?

kurthindenburg:
Well I don't think there was a consensus on how to proceed with these [pwd] issues.

pmetzger:
@kurthindenburg Well, given the lack of consensus and the passage of eight months, perhaps this should be closed? The PR queue is really for things that are imminently awaiting integration. If something hasn't gotten consensus after a long time it probably is not going to be integrated soon.

PR #32, opened 2017-08-04
Add checks for valid epoch and revision in mportopen

kurthindenburg:
Mimic port lint to check epoch and revision are integers.  Note that
this will cause portindex to fail.

closes https://trac.macports.org/ticket/26486

### DIFF
--- a/src/macports1.0/macports.tcl
+++ b/src/macports1.0/macports.tcl
@@ -1787,6 +1787,17 @@ proc mportopen {porturl {options {}} {variations {}} {nocache {}}} {
     }
     ditem_key $mport provides $actual_subport
 
+    set epoch [$workername eval {set PortInfo(epoch)}]
+    if {![string is wideinteger -strict $epoch]} {
+        mportclose $mport
+        error "$actual_subport epoch is not numeric: $epoch"
+    }
+    set revision [$workername eval {set PortInfo(revision)}]
+    if {![string is wideinteger -strict $revision]} {
+        mportclose $mport
+        error "$actual_subport revision is not numeric: $revision"
+    }
+
     return $mport
 }
 


jmroot:
Any reason not to do this with an option_proc in the portfile context? Also, since an error is raised by this new check, the lint check can be removed.

kurthindenburg:
Getting back around to this; I'm not sure I understand what you mean.  Could you explain it a bit more please?

ryandesign:
An `option_proc` is a procedure that can be attached to an "option" (like epoch and revision and most of the other things you can set in a portfile), which runs when that option is modified or accessed. The procedure can make changes to the value being set, can prevent the value from being set, or do anything else really. There are various `option_proc`s in various files in src/port1.0. Maybe portmain.tcl would be the appropriate place to do this?

jmroot:
Like this:
```diff --git a/src/port1.0/portmain.tcl b/src/port1.0/portmain.tcl
index 96f5202c..7e12d465 100644
--- a/src/port1.0/portmain.tcl
+++ b/src/port1.0/portmain.tcl
@@ -58,12 +58,23 @@ options prefix name version revision epoch categories maintainers \
         compiler.cpath compiler.library_path \
         add_users
 
+proc portmain::check_option_integer {option action args} {
+    if {$action eq "set"} {
+        if {![string is wideinteger -strict $args]} {
+            return -code error "$option must be an integer"
+        }
+    }
+}
+
 # Order of option_proc and option_export matters. Filter before exporting.
 
 # Assign option procedure to default_variants
 option_proc default_variants handle_default_variants
 # Handle notes special for better formatting
 option_proc notes handle_option_string
+# Ensure that revision and epoch are integers
+option_proc epoch portmain::check_option_integer
+option_proc revision portmain::check_option_integer
 
 # Export options via PortInfo
 options_export name version revision epoch categories maintainers platforms description long_description notes homepage license provides conflicts replaced_by installs_libs license_noconflict patchfiles
```

PR #33, opened 2017-08-25
portlint: add warning(s) about modeline

kurthindenburg:
Only warn if a modeline is found that is not on the first line.
The guide has conflicting info as to whether a modeline is required
and what it should be.

"This should be the first line of a Portfile. It sets the correct editing options for vim and emacs. See Port Style for more information. Its use is optional and up to the port maintainer."
"By default, the top line of all Portfiles should use a modeline that defines soft tabs for the vim and emacs editors as shown."

closes https://trac.macports.org/ticket/52767

### DIFF
--- a/src/port1.0/portlint.tcl
+++ b/src/port1.0/portlint.tcl
@@ -197,9 +197,20 @@ proc portlint::lint_main {args} {
             incr warnings
         }
 
-        if {($lineno == $topline_number) && [string match "*-\*- *" $line]} {
-            ui_info "OK: Line $lineno has emacs/vim Mode"
-            incr topline_number
+        if {($lineno == $topline_number)} {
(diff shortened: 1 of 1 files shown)


grimreaper:
LGTM but maybe make
```
+                ui_info "OK: Line $lineno does not have a modeline; consider adding a modeline, preferability the standard MacPort's modeline"
```

output something to copy/paste?

raimue:
@jmroot requested changes to this in the corresponding [ticket](https://trac.macports.org/ticket/52767#comment:3). To make this copy & paste ready as suggested by @grimreaper also seems useful.

As we already have a ticket to track this, I am closing this stale pull request until we get it worked out. Please reopen with an updated patch or create a new pull request.

PR #35, opened 2017-10-04
startupitem_type: remove SystemStarter

kurthindenburg:
Tiger introduced launchd and is the earliest OS that macports supports.
This removes the obsolete startupitem_type SystemStarter.

### DIFF
--- a/src/port1.0/portstartupitem.tcl
+++ b/src/port1.0/portstartupitem.tcl
@@ -55,12 +55,10 @@
 #   startupitem.logfile     logpath
 #       Log to the specified file -- if not specified then output to /dev/null
 #       - for launchd, just set this as the standard out key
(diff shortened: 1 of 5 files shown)


src/port1.0/portstartupitem.tcl
```diff
@@ -660,12 +373,7 @@ proc portstartupitem::startupitem_create {args} {
     if {${startupitem.type} eq "default" || ${startupitem.type} eq ""} {
         switch -exact ${os.platform} {
             darwin {
-                set haveLaunchd ${portutil::autoconf::have_launchd}
```
raimue: I guess this means we can also remove the check from configure and the variable in `src/port1.0/port_autoconf.tcl.in`.

PR #36, opened 2017-10-06
Remove configure checks for launchd

kurthindenburg:
Macports supports Tiger and later which all have launchd.

### DIFF
--- a/configure
+++ b/configure
@@ -693,7 +693,6 @@ CFLAGS_QUICHEEATERS
 TAR_CMD
 TAR_Q
 SED_EXT
-HAVE_LAUNCHD
 ZIP
 XZ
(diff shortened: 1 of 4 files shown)


ryandesign:
Will this be a problem for building on non-Mac systems that might not have launchd?

kurthindenburg:
I don't believe so; the only use for HAVE_LAUNCH was under startupitem_create_darwin_launchd()

There is a check in portstartupitem for non-darwin system to use rcng; however that will be removed on next pull (unless there is objection - see https://trac.macports.org/ticket/36770).

ryandesign:
Ok thanks. And thanks for all the cleanup work you've been doing! 👍

PR #37, opened 2017-10-06
port lint: allow spaces before PortGroup lines

kurthindenburg:
Currently, port lint only finds PortGroups that start on first column.
A large number of ports have PortGroup lines indented due to ifs.

Note: lint displays multiple identical PortGroup groups as an error.

### DIFF
--- a/src/port1.0/portlint.tcl
+++ b/src/port1.0/portlint.tcl
@@ -227,8 +227,8 @@ proc portlint::lint_main {args} {
             set require_blank true
             set require_after "PortSystem"
         }
-        if {[string match "PortGroup*" $line]} {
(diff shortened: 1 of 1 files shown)


src/port1.0/portlint.tcl
```diff
@@ -227,8 +227,8 @@ proc portlint::lint_main {args} {
             set require_blank true
             set require_after "PortSystem"
         }
-        if {[string match "PortGroup*" $line]} {
-            regexp {PortGroup\s+([A-Za-z0-9_]+)\s+([0-9.]+)} $line -> portgroup portgroupversion
+        if {[string match "*PortGroup*" $line]} {
+            regexp {\s*PortGroup\s+([A-Za-z0-9_]+)\s+([0-9.]+)} $line -> portgroup portgroupversion
```
jmroot: Regex needs to be anchored to the start of the line.

PR #39, opened 2017-10-10
base: tracelib: Use binary search in dep_check()

neverpanic:
Keeping the list of dependencies sorted allows using binary search when
determining whether a file is provided by a port in the dependency tree.

Depending on the length of the list of dependencies, this can yield
build time improvements of around 10 %.

Additionally, use Tcl's list functions to interpret the list of
dependencies rather than splitting at whitespaces manually.

### DIFF
--- a/src/pextlib1.0/tracelib.c
+++ b/src/pextlib1.0/tracelib.c
@@ -585,6 +585,13 @@ static void sandbox_violation(int sock UNUSED, const char *path, sandbox_violati
 }
 
 /**
+ * Internal helper function to compare two strings.
+ */
+static int pointer_strcmp(const char** a, const char** b) {
+    return strcmp(*a, *b);
+}
+
+/**
  * Check whether a path is in the transitive hull of dependencies of the port
  * currently being installed and send the result of the query back to the
  * socket.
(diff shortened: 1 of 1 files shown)


neverpanic:
Here's some rough testing:
```
baseline:
  source-highlight: 79.35 real       171.27 user        23.24 sys
  gimp2:           445.11 real       832.40 user       301.46 sys
binary search:
  source-highlight: 78.77 real       168.45 user        22.88 sys
  gimp2:           403.69 real       725.90 user       265.76 sys

```

src/pextlib1.0/tracelib.c
```diff
@@ -650,12 +647,23 @@ static void dep_check(int sock, char *path) {
...
+            /* found the port */
             free(port);
             answer(sock, "+");
             return;
+        } else if (result < 0) {
+            /* continue search right */
+            left = index + 1;
+        } else {
+            /* continue search left */
+            right = index;
```
raimue: Looks like this block could be replaced with POSIX bsearch(3). Untested example:
```
if (NULL != bsearch(port, depends, dependsLength, sizeof(depends[0]), (int (*)(const void*, const void*)) pointer_strcmp)) {
    free(port);
    answer(sock, "+");
    return;
}
```

src/pextlib1.0/tracelib.c
```diff
@@ -1004,28 +1016,59 @@ static int TracelibCloseSocketCmd(Tcl_Interp *interp UNUSED) {
...
-    depends = malloc(l + 2);
-    if (!depends) {
+    if (TCL_OK != Tcl_ListObjGetElements(interp, objv[2], &length, &objects)) {
+        return TCL_ERROR;
+    }
+
+    /* When called twice, do not leak memory */
+    if (depends) {
+        for (size_t i = 0; i < dependsLength; ++i) {
+            free(depends[i]);
```
Ionic: Could also set `depends[i] = NULL;` here and below, but probably not really important since the chunk of memory is unreachable afterwards.

neverpanic: Thought about that but didn't precisely for that reason.

raimue:
I just went with rebase, as the commit messages were clear enough.

PR #40, opened 2017-10-14
Remove rcng startupitem type

kurthindenburg:
Remove non-darwin startupitem_type rcng.  As far as I can tell, it hasn't worked in a long time since it was committed in 2005.    It is missing the etc/rc.subr which contains the script needed for the rc scripts to run.  If this is needed later, it would be better to start from scratch using a new version of rc scripts.

### DIFF
--- a/doc/macports.conf.5
+++ b/doc/macports.conf.5
@@ -395,7 +395,7 @@ lt lt.
 T{
 \fBSupported types:\fR
 T}:T{
-none, launchd, default, rcNG\&.
+none, launchd, default\&.
 T}
 T{
 \fBDefault:\fR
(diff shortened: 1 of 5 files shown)


jmroot:
No port using it isn't a factor as it's not something that a port would set; it's selected based on os.platform or set in macports.conf.

Are you sure it's actually broken? Sourcing ${prefix}/etc/rc.subr when it doesn't exist will fail harmlessly.

kurthindenburg:
Ok, I agree the commit message needs changed.  However, it does not work as the rc.subr should contain commands (load_rc_config, run_rc_command); it errors out when trying to run the rc.d/ file.   

Tested with  openntpd and linux.

https://trac.macports.org/ticket/36770 has more info on the subject - I believe the gist is to remove the startupitem_type

raimue:
Sorry, totally missed the part about the commit message. It's still correct, unless a user manually installed a `rc.subr` to the prefix, the generated script was useless and would only produce an error, so that is kind of broken.

PR #44, opened 2017-11-05
Use sig_atomic_t instead of int for signal handler

grimreaper:
Strictly speaking sig_atomic_t does not have to be `int` though any
system this code targets is unlikely to be affected.


Ionic:
Rebased and merged (though github didn't notice it due to the rebase.)

PR #45, opened 2017-11-08
Opportunistic HFS compression.

eborisch:
Detects (at runtime) modern bsdtar from libarchive and activates HFS+/APFS compression when available. See -dev thread https://lists.macports.org/pipermail/macports-dev/2017-November/036758.html and https://trac.macports.org/ticket/36560 for discussion.


src/registry2.0/portimage.tcl
```diff
@@ -348,9 +348,18 @@ proc extract_archive_to_tmpdir {location} {
             t(ar|bz|lz|xz|gz) {
                 set tar "tar"
                 if {[catch {set tar [macports::findBinary $tar ${macports::autoconf::tar_path}]} errmsg] == 0} {
-                    ui_debug "Using $tar"
-                    set unarchive.cmd "$tar"
-                    set unarchive.pre_args {-xvpf}
+                    # Opportunistic HFS compression. bsdtar will automatically
+                    # ignore if on non-HFS filesystem.
+                    if {![catch {macports::binaryInPath bsdtar}] &&
+                        ![catch {exec bsdtar -x --hfsCompression < /dev/null >& /dev/null}]} {
```
jmroot: Might as well just check `[file executable ${prefix}/bin/bsdtar]` (and not bother checking if --hfsCompression is supported) since that's the only place a suitably new bsdtar will be. If the next macOS release ships one, we can check for it once at configure time.

RJVB: That's exactly what I'm doing in my version of this patch as well as in a couple of ports that unpack huge source trees. Works perfectly, as long as the target filesystem supports compression.  I don't know how bsdtar handles pre-compression HFS versions but it will raise an error when unpacking to a non-HFS filesystem. That doesn't just concern older OS X versions; there is no obligation that I know of to install MacPorts to an Apple filesystem (all of ${prefix} could be on a ZFS dataset, for instance).
/Methinks that for the current application it should suffice to provide an option in macports.conf, possible setting it appropriately during the initial install. This is a local configuration aspect that won't change often. Or, as I do in my local `compress_workdir` PG: query an env. variable to opt in or out of the feature.

> What happens if ${portdbpath}/software is on a different volume than the rest of ${prefix}?

That's going to depend on how you move the unpacked (MPExtract) directory to its destination. Currently compression is lost (my /Applications/Macports is on a different volume than ${prefix}) but that can be avoided if the move is handled through `ditto` or `${prefix}/rsync`.

eborisch: >What happens on Leopard and Tiger?

I don't know with absolute certainty (haven't tested), but according to https://github.com/aosm/xnu/commit/d7b24a84012524e9dfcec55e4a6f2a5ff1f00199 the UF_COMPRESSED flag was not defined prior to 10.10, so the code which turns on the compress steps (but *not* the code that permits the --hfsCompression flag to be parsed) is disabled, so it be comes a no-op: https://github.com/libarchive/libarchive/blob/c10875db78b04fd730bb08245b7032baba906815/libarchive/archive_write_disk_posix.c#L634

> Might as well just check [file executable ${prefix}/bin/bsdtar] (and not bother checking if --hfsCompression is supported)

My thought here was that a user might install bsdtar (somewhere else in their PATH) on their own if they want to make sure it is always available; a corner case to be sure... I'm fine either way. It would be quicker to not have to check the flag, but it's also not slow. (Only done once per extraction.)

>I don't know how bsdtar handles pre-compression HFS versions but it will raise an error when unpacking to a non-HFS filesystem. 

No, it (bsdtar) silently disables the compression step. See the email discussion chain and the libarchive source.

>>What happens if ${portdbpath}/software is on a different volume than the rest of ${prefix}

> That's going to depend on how you move the unpacked (MPExtract) directory to its destination. Currently compression is lost (my /Applications/Macports is on a different volume than ${prefix}) but that can be avoided if the move is handled through ditto or ${prefix}/rsync.

Good point. That is a separate step and can be addressed separately. This change won't break that step in that case, but it will be unnecessarily slower for the (wasted) compression action. A check could certainly be added to see if our extract location and expected install locations ($prefix and /Applications/MacPorts) are on different mount points; but I suspect many (most?) users aren't in that situation.

RJVB:
On Wednesday November 08 2017 18:51:35 Eric A. Borisch wrote:

> I don't know with absolute certainty (haven't tested), but according to https://github.com/aosm/xnu/commit/d7b24a84012524e9dfcec55e4a6f2a5ff1f00199 the UF_COMPRESSED flag was not defined prior to 10.10

It is defined in /usr/include/sys/stat.h on 10.9 and probably on every OS X version that supported HFS compression.
It's used in afsctool, and AFAIK that application runs even on 10.6 .


> >I don't know how bsdtar handles pre-compression HFS versions but it will raise an error when unpacking to a non-HFS filesystem. 
> 
> No, it (bsdtar) silently disables the compression step. See the email discussion chain and the libarchive source.

Maybe the latest versions do, but earlier this year when I tested putting $prefix/var/macports/build on a ZFS dataset it failed. That's when and why I added support for opting in/out via an env. variable to my PortGroup.

Not that I'll blame anyone participating in that discussion for not having tested the --hfsCompression option on incompatible filesystems, and libarchive isn't wrong in raising an error either.

And FWIW, I said `a` non-HFS filesystem, not `all` ;)

> but I suspect many (most?) users aren't in that situation.

Anyone who has $prefix on a dedicated volume but kept /Applications/MacPorts is in this situation. The best way to handle all cases would be to move the extracted files into place using rsync which has an option to preserve HFS compression or drop it if unsupported.

R.

RJVB:
I have to correct myself. bsdtar 3.3.1 does NOT raise an error when extracting to ZFS with --hfsCompression. It does something worse: it creates empty files.

eborisch:
>It is defined in /usr/include/sys/stat.h on 10.9 and probably on every OS X version that supported HFS compression.
It's used in afsctool, and AFAIK that application runs even on 10.6 .

Apologies; I didn't dig enough. Defined since 10.6; but to the original Leopard/Tiger statement, it's not defined, so the code is not run.

>I have to correct myself. bsdtar 3.3.1 does NOT raise an error when extracting to ZFS with --hfsCompression. It does something worse: it creates empty files.

That's called a feature. ;)

I've been a big ZFS fan since OpenSolaris 2008.11, and use it extensively -- on FreeBSD and ZoL. I guess I'd say that I wouldn't put it up as a roadblock for MacPorts development. We can certainly check that the temporary extraction directory is mounted with hfs or apfs as belts-and-suspenders.

>Anyone who has $prefix on a dedicated volume but kept /Applications/MacPorts is in this situation. The best way to handle all cases would be to move the extracted files into place using rsync which has an option to preserve HFS compression or drop it if unsupported.

No doubt. That should be handled during the move-into-place step and can be a separate discussion. This (compressing the 'source' of the potential rsync) would be a necessary pre-condition to using rsync's compression preservation... for (yes, no proof, but guessing) the bulk of users with one hfs/apfs volume, however, this change gets them the compression benefits independent of an updated move-into-place step.

RJVB:
>That's called a feature. ;)

As in lossy compression?

>I've been a big ZFS fan since OpenSolaris 2008.11, and use it extensively -- on FreeBSD and ZoL. I guess I'd say that I wouldn't put it up as a roadblock for MacPorts development. We can certainly check that the temporary extraction directory is mounted with hfs or apfs as belts-and-suspenders.

OpenZFSOnOSX is getting better and better, even providing boot support nowadays. It'd be a great solution to put $prefix on a dedicated volume, with dedicated datasets with appropriate transparent compression and/or performance tweaks for the various parts of the MacPorts tree - all mounted under /opt/local without fuss or symlinks.

Anyway, I'd indeed incorporate a file system check (or make certain libarchive does this). Or simply a setting in macports.conf. Not just because of ZFS, of course.

>No doubt. That should be handled during the move-into-place step and can be a separate discussion.

Of course. I haven't even bothered with that on my own system, I just run afsctool from time to time.

>This (compressing the 'source' of the potential rsync) would be a necessary pre-condition to using rsync's compression preservation...

Separate discussion, but there's no need to make this overly complicated if rsync is used. It already does all those checks: it won't preserve compression if the target doesn't support it but also won't compress an uncompressed file.

R

eborisch:
Added check that device that extraction, prefix, and applications_dir are in. Also check that extraction directory is on apfs or hfs volume.

RJVB:
Just a heads-up: I tried `bsdtar -x --hfsCompression` on NFS and AFP shares and found the option is indeed ignored silently.

If anyone reading this is already involved with libarchive development (or just bug reporting/discussion), maybe they can forward the ultimately lossy compression effect the option has on ZFS datasets? ;)

src/registry2.0/portimage.tcl
```diff
@@ -349,9 +355,17 @@ proc extract_archive_to_tmpdir {location} {
...
+                    # The comparisons against prefix and applications_dir should be
+                    # revisted if rsync or ditto is eventually used to perform final
+                    # move into destinations.
+                    set extrfs [_getfs ${extractdir}]
                     if {![catch {macports::binaryInPath bsdtar}] &&
-                        ![catch {exec bsdtar -x --hfsCompression < /dev/null >& /dev/null}]} {
+                        ![catch {exec bsdtar -x --hfsCompression < /dev/null >& /dev/null}] &&
+                        ${extrfs} == [_getfs ${macports::prefix}] &&
+                        ${extrfs} == [_getfs ${macports::applications_dir}] &&
+                        [string length [exec mount | egrep "${extrfs} .*(hfs|apfs)"]] > 0 } {
```
raimue: The match needs to be anchored at the beginning of the line with `^` and needs to match the correct field. What if the mount point directory contains the substring "hfs" or "apfs"?

Also see above, should rather use `statfs(2)` to determine filesystem.

src/registry2.0/portimage.tcl
```diff
@@ -348,9 +348,18 @@ proc extract_archive_to_tmpdir {location} {
             t(ar|bz|lz|xz|gz) {
                 set tar "tar"
                 if {[catch {set tar [macports::findBinary $tar ${macports::autoconf::tar_path}]} errmsg] == 0} {
```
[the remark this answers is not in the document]
eborisch: It's used to set the error string if finding tar fails. Leaving it for now.

src/registry2.0/portimage.tcl
```diff
@@ -297,6 +297,12 @@ proc _activate_file {srcfile dstfile} {
     }
 }
 
+# Detect the /dev device containing path
+# returns: path of containing device
+proc _getfs {path} {
+    return [exec df [file normalize ${path}] | awk {{if (NR > 1) print $1}} ]
```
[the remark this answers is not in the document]
eborisch: I'll look into adding functions for device node and filesystem type (both via ```statfs(2)```) to pextlib this weekend and updating.

src/registry2.0/portimage.tcl
eborisch: Yep. Will look at this weekend as time permits.

src/registry2.0/portimage.tcl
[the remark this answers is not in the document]
raimue: It makes no sense that you would error out when `tar` was not found, but the code would actually never have used it.

I meant it should be like this:
```
if {![catch {macports::binaryInPath bsdtar}] && ... } {
    ... # using bsdtar --hfsCompression
} else {
    if {[catch {set tar [macports::findBinary tar ...] ... } {
        ... # using tar
    } else {
        ... # tar not found error
    }
}
```

src/pextlib1.0/statfs.c
```diff
@@ -0,0 +1,129 @@
+/*
+ * realpath.c
```
raimue: `statfs.c`

src/pextlib1.0/statfs.c
```diff
@@ -0,0 +1,129 @@
...
+ *
+ * Copyright (c) 2009 The MacPorts Project.
+ * All rights reserved.
```
raimue: `statfs.c` and 2017.

src/pextlib1.0/Pextlib.c
```diff
@@ -1133,6 +1134,9 @@ int Pextlib_Init(Tcl_Interp *interp)
 
     Tcl_CreateObjCommand(interp, "fs_case_sensitive", FSCaseSensitiveCmd, NULL, NULL);
 
+    Tcl_CreateObjCommand(interp, "devicenode", DeviceNodeCmd, NULL, NULL);
+    Tcl_CreateObjCommand(interp, "fs_type", FilesystemTypeCmd, NULL, NULL);
+
```
raimue: To keep the interface closer to the C API, I would do the same as in our implementation of [getpwuid](https://github.com/macports/macports-base/blob/master/src/pextlib1.0/uid.c#L203-L306) and either return a list of key/value pairs for the `struct statfs` members or a specific field that was requested. These two examples would then give equivalent output:

```
array set data [statfs $path]
puts $data(mntfromname)
puts $data(fs_type)
```
```
puts [statfs $path mntfromname]
puts [statfs $path fs_type]
```

Although we do not need the other entries right now, we might later need them. I think it is better to keep the abstraction to the C API as low as possible here to make it easier to understand what this is meant to retrieve.

I am sorry that I did not outline this directly that clearly when I suggested a wrapper for `statfs(2)`. If you want me to, I could also look into implementing `statfs` in the same style as `getpwuid`.

src/registry2.0/portimage.tcl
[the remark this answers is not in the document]
eborisch: The tar string set in 355 is used both in 356 (where it may get set again), and again in all the way down in 411 if finding tar fails. Yes, this could be rewritten with the throw much closer to the checks, and the "success case" outside the if - else - throw structure, but I was trying to minimize the lines of code (and whitespace from a removed indent) change... your thoughts? I can go either way.

src/pextlib1.0/compile_commands.json
```diff
@@ -0,0 +1,7 @@
+[
```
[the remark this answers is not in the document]
eborisch: That was an error on my part.

src/pextlib1.0/Pextlib.c
eborisch: I've never written these before, so I went for low complexity so I could get it done when I had a few moments last night. If you're up for putting together a lower level version, I'm all for it!

eborisch:
I've switched over to X-macros for states accessors. It is trivial to add additional ones now, and this should be less intensive than parsing the entirety of the states structure (not that it's large) when most users want a particular one or two elements. I've been using this to great effect on my own machine to reduce the footprint of MacPorts.

raimue:
I took another look at this pull request now. Thanks for investing the work, but I do not think these X-macros are how this should be solved. This requires a separate syscall to get each struct member, when one syscall already all values. Furthermore, the code for the macro generator is hard to read and understand.

I hope you do not mind that I started working on this. I took this pull request, rebased it against current master and implemented a statfs command that returns all relevant fields. I also fixed the remaining issue that we do not need to check for tar if we are going to use bsdtar anyway.

As I do not want to overwrite your master branch, I opened a new pull request instead:
https://github.com/macports/macports-base/pull/76

PR #47, opened 2017-11-11
Fix `make install DESTDIR=/foobar`

yan12125:
In the current version, trying to running `make install DESTDIR=/foobar` as a normal user (with write permissions to /foobar) results in permission errors.

The DESTDIR approach is used in my [Arch Linux personal package](https://aur.archlinux.org/packages/macports-base-git/). I'm going to run version checks as a daily cron job on my Linux server.


yan12125:
I got a notice:

> buildbot failure in MacPorts on base-10.5_ppc

Looks like related to https://trac.macports.org/ticket/51516? https://build.macports.org/builders/base-10.5_ppc/builds/268/steps/test/logs/stdio says:

> Error: Failed to fetch gcc_select: error:14077410:SSL routines:SSL23_GET_SERVER_HELLO:sslv3 alert handshake failure

jmroot:
Don't worry about that, one of the tests is broken on that platform.

PR #48, opened 2017-12-01
startupitem.type: finish removing obsolete code

kurthindenburg:
Remove all remaining code dealing with startupitem.type

If anyone modified their macports.conf to adjust startupitem.type,
it is silently ignored.  There's only one port (denyhost) using
startupitem.type; I'll fix that shortly.

closes https://trac.macports.org/ticket/36770

### DIFF
--- a/doc/macports.conf.5
+++ b/doc/macports.conf.5
@@ -385,27 +385,6 @@ T}
 .sp 1
 .RE
 .PP
-startupitem_type
-.RS 4
-Set the default type of startupitems to be generated, overridable by Portfiles that explicitly state a startupitem\&.type key\&. If set to "default", then a type will be selected that\(cqs appropriate to the OS\&.
-.TS
-tab(:);
-lt lt
-lt lt.
-T{
-\fBSupported types:\fR
-T}:T{
-none, launchd, default\&.
-T}
-T{
-\fBDefault:\fR
-T}:T{
-default
-T}
-.TE
-.sp 1
-.RE
-.PP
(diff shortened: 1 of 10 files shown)


jmroot:
Is there definitely no situation in which someone might want to not create startupitems at all, rather than just not link them into /Library/LaunchDaemons? As I think I've mentioned, I'm planning to change startupitem.install to controlling the latter.

kurthindenburg:
I don't see a reason why a port would set startupitem.create "yes" and need to set startupitem.type "none"

I thought that startupitem.install already controlled whether a symlink is created?  However, I see a number of ports that check for it and run code depending on it.

jmroot:
I'm thinking more of the case where it's set in macports.conf. Currently, `startupitem_install no` and `startupitem_type none` do the same thing, as far as I know. I can't think of any case off the top of my head where you would need to prevent generation of startupitems entirely, but it's something to consider.

kurthindenburg:
I see what you mean; using startupitem_type none seems like a misuse for that case.  Perhaps better would be allow startupitem_create to be set in macports.conf?  Although, I don't really see any real issue with just create the startup files regardless.

grimreaper:
Ref #53  in which I did some small cleanup

jmroot:
I'll deal with this as part of https://trac.macports.org/ticket/44496.

pixilla:
Late comment, there may have been a case where people used port command without sudo and can't write to /Library/LaunchDaemons.

PR #49, opened 2017-12-08
Add use_no_compression option

emcrisostomo:
`use_no_compression` configures the fetch and the extract phases to deal with a non-compressed file with no extract extension.

When creating the port for Amazon AWS `ecs-cli`, I had to deal with the case in which the port has to download an uncompressed binary file (see https://github.com/aws/amazon-ecs-cli).

The port (submitted in https://github.com/macports/macports-ports/pull/1089), works with the following configuration:

```
extract.suffix
extract.cmd             cp
extract.pre_args
extract.post_args       ${worksrcpath}
extract.mkdir           yes
```

Basically, the extract suffix is set to an empty string and the downloaded file is just copied into `${worksrcpath}`, which gets created by setting `extract.mkdir` to `yes`.

I thought it may be useful to deal with this case natively, and I added the `use_no_compression` option.

### DIFF
--- a/src/port1.0/portextract.tcl
+++ b/src/port1.0/portextract.tcl
@@ -74,7 +74,7 @@ proc portextract::disttagclean {list} {
 }
 
 proc portextract::extract_start {args} {
-    global UI_PREFIX extract.dir extract.mkdir use_bzip2 use_lzma use_xz use_zip use_7z use_lzip use_dmg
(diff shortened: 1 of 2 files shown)


emcrisostomo:
After learning a much simpler workaround, I realise this is probably an over-engineered solution for a problem that's easy to overcome. Closing.

ryandesign:
This is a pretty uncommon situation that I don't think needs special treatment in MacPorts base.

emcrisostomo:
> This is a pretty uncommon situation that I don't think needs special treatment in MacPorts base.

I agree.

PR #54, opened 2017-12-23
Update year in AUTHORS for port man page

JacksonIsaac:
Update year range from 2004-2012 to 2004-2017
under AUTHORS for port man page.

Change the from branch to separate branch instead of fork/master. 
Replaces PR #50

### DIFF
--- a/doc/port.1.txt
+++ b/doc/port.1.txt
@@ -44,7 +44,7 @@ available port tree(s). These may be used in the same way as a 'portname'.
     - 'inactive': set of installed but inactive ports
     - 'installed': set of all installed ports
     - 'uninstalled': ports in the ports tree(s) that are not installed
-    - 'outdated': installed ports that are out of date with respect to their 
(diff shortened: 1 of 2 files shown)


doc/port.1.txt
```diff
@@ -178,7 +178,7 @@ The port command recognizes several global flags and options.
 -t::
     Enable trace mode debug facilities on platforms that support it, currently
     only Mac OS X.
-    + 
+    +
```
neverpanic: I think this is a change of behavior. While trailing whitespace is uncommon, it's the only way in asciidoc to add a newline to a paragraph without ending the paragraph.

Can you check the output to make sure that the text below this linebreak is still associated with the `-t` command line option?

JacksonIsaac: I checked the man page before and after removing the trailing whitespace and it looks same to me.

Screenshot of comparison: https://imgur.com/a/u671e

neverpanic: OK, thanks for checking.

PR #55, opened 2017-12-29
If expected path of extract doesn't exist, symlink to actual path.

slewsys:
This attempts to resolve an [issue raised by Mojca](https://github.com/macports/macports-ports/pull/1075#discussion_r154493451) where worksrcdir has to be set manually if a source archive extracts to an unexpected directory.  In this particular case, for instance, the archive is named v2.4.1.tar.gz and extracts to libwebsockets-2.4.1.

Similarly, using the github.setup defaults, the archive (tagged as v2.4.1) is libwebsockets-2.4.1.tar.gz and extracts to warmcat-libwebsockets-e4e774b.

### DIFF
--- a/src/port1.0/portextract.tcl
+++ b/src/port1.0/portextract.tcl
@@ -124,7 +124,7 @@ proc portextract::extract_start {args} {
 }
 
 proc portextract::extract_main {args} {
-    global UI_PREFIX filespath worksrcpath extract.dir use_dmg
+    global UI_PREFIX filespath workpath worksrcdir worksrcpath extract.dir use_dmg
 
     if {![exists distfiles] && ![exists extract.only]} {
         # nothing to do
@@ -155,6 +155,17 @@ proc portextract::extract_main {args} {
         }
 
         chownAsRoot ${extract.dir}
+    }
+
+    # If expected path of extract doesn't exist && worksrcdir is
+    # not explicitly set to subdirectory, symlink to actual path.
+    if {![file isdirectory $worksrcpath] && [regexp {^[^/]+$} $worksrcdir]} {
+        set workdirs [glob -nocomplain -types d [file join $workpath *]]
+        if {[llength $workdirs] == 1} {
+            set dir [file tail [lindex $workdirs 0]]
+            ui_debug [format [msgcat::mc "Symlink: %s -> %s"] $worksrcpath $dir]
+            symlink $dir $worksrcpath
+        }
     }
     return 0
 }
--- a/src/pextlib1.0/fs-traverse.c
+++ b/src/pextlib1.0/fs-traverse.c
@@ -170,7 +170,7 @@ do_traverse(Tcl_Interp *interp, int flags, char * CONST *targets, Tcl_Obj *varna
     FTS *root_fts;
     FTSENT *ent;
 
-    root_fts = fts_open(targets, FTS_PHYSICAL /*| FTS_COMFOLLOW */| FTS_NOCHDIR | FTS_XDEV, &do_compare);
+    root_fts = fts_open(targets, FTS_PHYSICAL | FTS_COMFOLLOW | FTS_NOCHDIR | FTS_XDEV, &do_compare);
 
     while ((ent = fts_read(root_fts)) != NULL) {
         switch (ent->fts_info) {
--- a/src/pextlib1.0/tests/fs-traverse.tcl
+++ b/src/pextlib1.0/tests/fs-traverse.tcl
@@ -255,6 +255,10 @@ proc setup_trees {root} {
 
     set trees(sub1) "
         $root/a/c/a     {link ../d}
+        $root/a/c/a/a   file
+        $root/a/c/a/b   {link ../../b/a}
+        $root/a/c/a/c   directory
+        $root/a/c/a/d   file
     "
 
     set trees(sub2) "


slewsys:
An alternative approach might be to update global variables worksrcdir and worksrcpath, but without looking deeper, it’s not clear to me whether by the time extract is run, e.g., cmake’s configure command hasn't already been defined in terms of the previous value of worksrcpath.

jmroot:
I don't really see what problem this solves. The referenced PR doesn't seem to contain any workaround involving the directory name. If worksrcdir should be different to distname then it can just be set that way.

slewsys:
The problem is that in 900 ports, worksrcdir is explicitly defined.  I don’t claim that all of those can be eliminated, but even if only a fraction can, it seems that those proposed 6 lines of code merits consideration.
-AM

ryandesign:
It doesn't seem like it's that difficult for the port author to specify the `worksrcdir`.

> Similarly, using the github.setup defaults, the archive (tagged as v2.4.1) is libwebsockets-2.4.1.tar.gz and extracts to warmcat-libwebsockets-e4e774b

The github portgroup handles the weird github `worksrcdir` values correctly. The only reason it didn't work in the above-referenced PR was because the portfile submitted there subverted the github portgroup's correct functioning by overriding some of its variables.

slewsys:
Okay, thank you for the corrections.  In the mean time, I’ve been running the patch mostly without incident.  The one exception I’ve come across so far is pax, whose worksrcdir is a subdirectory of the extracted archive.  

I’d like to explore the issue further at some point, but until then, is there a way to mark a pull request dormant, or should I just delete it?
-AM

raimue:
This seems like a nice idea to avoid setting the `worksrcdir` option on many ports.

This should only be applied if `worksrcdir` is the default value. If `worksrcdir` contains slashes, the logic is wrong to match against the top-level.

This will break if any other directory is created inside `workpath`. We already have `.tmp` and `.home`, but they are not matched due to the leading dot. We could compare the list of directories before and after extraction to find the top-level directory from the tarball. Maybe it is not even not an issue and the complexity is not needed, I am just pointing it out.

ryandesign:
I have to admit that automatically handling an unexpected worksrcdir in a distfile is something that the github and bitbucket portgroups have had to deal with, and therefore support for that feature in MacPorts base wouldn't be entirely unwelcome since that would simplify those portgroups.

In those portgroups, the implementation is: find the weird directory name and rename it to worksrcdir. In this PR, the implementation is: find the weird directory name and symlink worksrcdir to it. A third possible implementation would be to somehow arrange for worksrcdir to be set to the weird directory name automatically. This would more closely match what other ports do when they manually set worksrcdir, but is harder to implement. It would also mean the value of worksrcdir would be undefined prior to the completion of the fetch phase, since we wouldn't know what's in the file until we've fetched it, but that's probably ok.

Automatic worksrcdir handling should be disabled if extract.mkdir is yes. If there are multiple distfiles, presumably the automatic worksrcdir handling should focus on the first distfile.

mojca:
You need to fix these commits.

slewsys:
Per Rainer’s comment, worksrcdir is now tested for slashes.  This allows xar(1) to build correctly (not pax, as I said previously).

I chose symlinks over renaming since they're non-descrutive.  However, symlinks can have unexpected side-effects.  For instance,  find(1) does not accept a symlink without a trailing slash, e.g.:

   find /ports

does  not work (where /ports is symlink).  It has to written as:

  find /ports/

Again, I chose not to set worksrcdir because that is also “destructive” like renaming.

If a symlink is created, it's at the end of the extract phase, so I would expect extract.mkdir to have already been executed?

As currently implemented, the PR is triggered only if there is one and only one extract directory not matching worksrcpath...
-AM

slewsys:
> ∑On Apr 4, 2018, at 12:34 PM, Mojca Miklavec <notifications@github.com> wrote:
> 
> You need to fix these commits.


Thank you for the heads up.  Running the following commands on my local fork seem to have repaired the problem:

  $ git checkout master
  $ git pull upstream master
  $ git checkout symlink_extracted_worksrcpath
  $ git rebase master
  $ git push -f

The problem seems to have been caused in the first place by rebasing prior to pulling from upstream.
-AM

raimue:
Any objections? This has been waiting in the queue for a while. I am in favor of merging this.

The commit message needs to be corrected, though, as it should follow our style for [CommitMessages](https://trac.macports.org/wiki/CommitMessages).

slewsys:
> On Apr 18, 2018, at 5:43 PM, Rainer Müller <notifications@github.com> wrote:
> 
> @raimue approved this pull request.


Just a heads up:  I found at least one port, audio/ID3, with a post-extract step that needs to be removed in order to compatible with #55:

# post-extract {
#     move "${workpath}/id3 tag framework" ${worksrcpath}
# }

 I also haven’t examined all the post-extract steps in the various port groups for compatibility.  Finally, Ryan mentioned extract.mkdir as a potential issue, which I haven’t explored yet.

So while I’d love to see the results from the buildbot with #55 installed, it may be a little premature...
-AM

raimue:
The buildbot is running the latest stable release, which is MacPorts 2.4.3 at the moment. Even if merged to master now, this would not appear anywhere until MacPorts 2.5.0 will be released. Developers running master have to expect breakage and are encouraged to report broken ports, so we can fix them.

That said, I am also fine if you want to continue with this on your branch until you think it is ready to be merged to master.

pmetzger:
Okay, so what remains at this point so that this can be integrated? (And does this imply any documentation changes we want to make?)

slewsys:
> 
> On Apr 19, 2018, at 11:32 AM, Perry E. Metzger <notifications@github.com> wrote:
> 
> Okay, so what remains at this point so that this can be integrated? (And does this imply any documentation changes we want to make?)

Mostly eyeballing post-extract code.   There are about 462 Portfiles with post-extract steps
according to the command:

$ cd /ports/ && git ls-files */Portfile | xargs -n30 egrep -l '^[[:space:]]*post-extract' | wc -l
462

What I’m looking for are Portfiles that create worksrcpath in the post-extract step like audio/ID3:

post-extract {
    move "${workpath}/id3 tag framework" ${worksrcpath}
}

That step becomes superfluous with the proposed patch, which creates a symlink instead.
But more significantly, the `move’ command above then fails because ${worksrcpath} already
exists (as a symlink).

Here’s a script that attempts to extract just the post-extract code from all Portfiles:

#!/usr/bin/env bash
#
# @(#) post-extract.sh
#
# This script attempts to extract the `post-extract' text from Portfiles.
# It assumes that `post-extract’  block ends with right curly bracket  (})
# with the same indentation as the `post-extract' keyword.
#
if ! cd /ports/ || test ! -d .git; then
    echo "Run this script from a MacPorts ports git repository"
    exit 1
fi

git ls-files */Portfile |
    xargs -n30 egrep -l '^[[:space:]]{0,}post-extract' |
    while read portfile; do
        echo "*** $portfile ***"
        portfile_text=$(< "$portfile")
        leading_space=$(ed -s <<EOF
a
$portfile_text
.
/^\\([[:space:]]\\{0,\\}\\)post-extract.*$/s;;\\1;p
EOF
                        )
        sed -ne "/^${leading_space}post-extract/,/^${leading_space}}[[:space:]]*\$/p" <<<"$portfile_text"
    done

pmetzger:
Perhaps The Right Thing would be to have a flag to disable the automatic code, add that flag to the existing Portfiles that try to do this on their own, merge the automation, and then incrementally remove the flag as we fix those Portfiles? Or would that be Way Too Much Work?

slewsys:
> On Apr 20, 2018, at 11:59 AM, Perry E. Metzger <notifications@github.com> wrote:
> 
> Perhaps The Right Thing would be to have a flag to disable the automatic code, add that flag to the existing Portfiles that try to do this on their own, merge the automation, and then incrementally remove the flag as we fix those Portfiles? Or would that be Way Too Much Work?

It should be easy. Searching the output of the previous script for `move.*worksrcpath’ suggests only a handful of issues, where removing the post-extract step is sufficient.  I’d like to add PRs for those ports.

And your interest helps, thank you!  I’ll plan to set aside a moment of clear-headedness, God willing, to see it through :)
-AM

jmroot:
> Perhaps The Right Thing would be to have a flag to disable the automatic code

Given that the reason for this change is to be able to avoid setting worksrcdir in some ports, having to set something else in other ports would appear to negate any benefit obtained.

slewsys:
On Apr 21, 2018, at 12:04 AM, Joshua Root <notifications@github.com> wrote:
> 
> Perhaps The Right Thing would be to have a flag to disable the automatic code
> 
> Given that the reason for this change is to be able to avoid setting worksrcdir in some ports, having to set something else in other ports would appear to negate any benefit obtained.

Agreed. That’s why I want to submit PRs for known issues, and compile a small sample of remaining ports to raise my confidence level.  So far, so good.
-AM

jmroot:
Would a smart default for worksrcdir be less problematic than creating symlinks? E.g.
```diff --git a/src/port1.0/portmain.tcl b/src/port1.0/portmain.tcl
index dbd5bbf0..4b085b51 100644
--- a/src/port1.0/portmain.tcl
+++ b/src/port1.0/portmain.tcl
@@ -95,6 +95,16 @@ proc portmain::get_subbuildpath {} {
     }
     return [file normalize [file join $portbuildpath $subdir]]
 }
+proc portmain::get_worksrcdir {} {
+    global distname workpath
+    if {![file isdirectory [file join $workpath [join $distname]]]} {
+        set candidates [glob -nocomplain -tails -types d -directory $workpath *]
+        if {[llength $candidates] == 1} {
+            return [lindex $candidates 0]
+        }
+    }
+    return $distname
+}
 default workpath {[getportworkpath_from_buildpath $subbuildpath]}
 default prefix /opt/local
 default applications_dir /Applications/MacPorts
@@ -108,7 +118,7 @@ default revision 0
 default epoch 0
 default license unknown
 default distname {${name}-${version}}
-default worksrcdir {$distname}
+default worksrcdir {[portmain::get_worksrcdir]}
 default filespath {[file join $portpath [join $filesdir]]}
 default worksrcpath {[file join $workpath [join $worksrcdir]]}
 # empty list means all archs are supported
```

raimue:
This "smart default" would be evaluated even if `${worksrcdir}` is accessed indirectly, for example with something like `build.dir ${worksrcpath}/subdir`. The main problem is that the value could change when the Portfile is evaluated again after the extract target was already executed. A Portfile could work with separate steps like `sudo port extract && sudo port build`, but not with `sudo port install`. This approach seems error-prone.

slewsys:
On Apr 21, 2018, at 7:39 AM, Rainer Müller <notifications@github.com> wrote:
> 
> This "smart default" would be evaluated even if ${worksrcdir} is accessed indirectly, for example with something like build.dir ${worksrcpath}/subdir. The main problem is that the value could change when the Portfile is evaluated again after the extract target was already executed. A Portfile could work with separate steps like sudo port extract && sudo port build, but not with sudo port install. This approach seems error-prone.

That state (e.g., the value of `worksrcdir`) is not be maintained between invocations of `port` is why `worksrcdir` should not be set, e.g., at the end of `port extract`.  But Joshua’s suggestion would initialize `worksrcdir` globally.

The main problem that I can think of is that if not overridden, the code in `worksrcdir` would be evaluated multiple times, making it inefficient without memoization.  On the other hand, it does nicely avoid cases found so far to trip up #55.

I’m in over my head without running the code, so do you have  a particular counter-example in mind?
-AM

jmroot:
It would be a problem if the value of $worksrcdir was used in setting some other variable (directly, not using `default`) before the extract phase happened. Which on reflection is probably not too uncommon.

pmetzger:
So what steps do we need to take to advance this for integration? Does it still need some time for thought?

slewsys:
On Apr 26, 2018, at 10:41 PM, Perry E. Metzger <notifications@github.com> wrote:
> 
> So what steps do we need to take to advance this for integration? Does it still need some time for thought?

I like Joshua’s idea.  I’ll submit a PR with memoization.   But I’ll also complete the symlink approach as a fallback.

It will take a little time. I may not get to it this weekend, but I'll set aside next weekend if not.  Okay?
-AM

pmetzger:
That seems great! Thank you for working on this.

slewsys:
On Apr 21, 2018, at 4:26 AM, Joshua Root <notifications@github.com> wrote:
> 
> index dbd5bbf0..4b085b51 100644
> 
> --- a/src/port1.0/portmain.tcl
> +++ b/src/port1.0/portmain.tcl
> @@ -95,6 +95,16 @@
>  proc portmain::get_subbuildpath {} {
>      }
>      return [file normalize [file join $portbuildpath $subdir]]
>  }
> 
> +proc portmain::get_worksrcdir {} {
> +    global distname workpath
> +    if {![file isdirectory [file join $workpath [join $distname]]]} {
> +        set candidates [glob -nocomplain -tails -types d -directory $workpath *]
> +        if {[llength $candidates] == 1} {
> +            return [lindex $candidates 0]
> +        }
> +    }
> +    return $distname
> +}
> 
>  default workpath {[getportworkpath_from_buildpath $subbuildpath]}
>  default prefix /opt/local
>  default applications_dir /Applications/MacPorts
> 
> @@ -108,7 +118,7 @@
>  default revision 0
>  default epoch 0
>  default license unknown
>  default distname {${name}-${version}}
> 
> -default worksrcdir {$distname}
> +default worksrcdir {[portmain::get_worksrcdir]}
> 
>  default filespath {[file join $portpath [join $filesdir]]}
>  default worksrcpath {[file join $workpath [join $worksrcdir]]}
>  # empty list means all archs are supported

Port audio/ID3 is problematic, principally because it sets build.dir:

build.dir           ${worksrcpath}/ID3

Portfile assignments such as this are evidently evaluated once and for all at the
time they're read - in particular, before any stages are run. So although `worksrcpath’
gets updated following extraction, the value of build.dir does not.

Adding memoization presents a similar hurdle because portmain::get_worksrcdir
also gets evaluated before any stages are run, and therefore initialized to the wrong
value. So although adding memoization allows audio/ID3 to build without change,
it does so for the wrong reason.

This is what I’m calling the “memoized” version:

+proc portmain::get_worksrcdir {} {
+    global __worksrcdir distname workpath
+    if {$__worksrcdir == ""} {
+        set __worksrcdir $distname
+        if {![file isdirectory [file join $workpath [join $distname]]]} {
+            set dirs [glob -nocomplain -tails -types d -directory $workpath *]
+            if {[llength $dirs] == 1} {
+                set __worksrcdir [lindex $dirs 0]
+            }
+        }
+    }
+    return $__worksrcdir
+}
+
 default workpath {[getportworkpath_from_buildpath $subbuildpath]}
 default prefix /opt/local
 default applications_dir /Applications/MacPorts
@@ -108,7 +122,10 @@ default revision 0
 default epoch 0
 default license unknown
 default distname {${name}-${version}}
-default worksrcdir {$distname}
+# __worksrcdir: memoized value of worksrcdir
+default __worksrcdir ""
+# default worksrcdir {$distname}
+default worksrcdir {[portmain::get_worksrcdir]}

-AM

jmroot:
Yeah, that's the problem I was referring to in my last comment.

slewsys:
On May 6, 2018, at 12:55 AM, Joshua Root <notifications@github.com> wrote:
> 
> Yeah, that's the problem I was referring to in my last comment.

The point, I guess, is that it’s not a shortcoming of the MacPorts implementation.
Nonetheless, PR #55 seems to be what the present design calls for.

I did naively try to prefix `build.dir’ assignment with `default’, which doesn’t
make it dynamic.  To my mind at least, your implementation opens the possibility
that such a mechanism within Portfile is be desirable.  GNU make, for instance,
provides `immediate’ and `deferred’ assignment operators `:=‘ and `=‘, respectively.
-AM

slewsys:
PR #1760 contains patches and updates for ports conflicting with PR #55.

No PRs submitted for port groups.  In the course of testing, though, I noticed that golang apps could probably benefit from their own port group.

src/port1.0/portextract.tcl
```diff
@@ -152,5 +152,17 @@ proc portextract::extract_main {args} {
...
     }
+
+    # If expected path of extract doesn't exist && worksrcdir is
+    # not explicitly set to subdirectory, symlink to actual
+    # path.
+    if {![file isdirectory $worksrcpath] && [regexp {^[^/]+$} $worksrcdir]} {
+        set workdirs [glob -nocomplain -types d [file join $workpath "*"]]
+        if {[llength $workdirs] == 1} {
+            set dir [file tail [lindex $workdirs 0]]
+            ui_info "$UI_PREFIX  [format [msgcat::mc "Symlink: %s -> %s"] $worksrcpath $dir]"
```
ryandesign: I think I would use `ui_debug` here instead of `ui_info`, and I would not use `$UI_PPREFIX`.

slewsys:
Replaced ui_info with ui_debug and removed UI_PREFIX per Ryan's comments.

slewsys:
On May 8, 2018, at 10:31 AM, Ryan Schmidt <notifications@github.com> wrote:
> However, symlinks can have unexpected side-effects. For instance, find(1) does not accept a symlink without a trailing slash, e.g.:
> 
> That's true, and it turns out to be a problem for the MacPorts equivalent procedure fs-traverse as well. It's not uncommon for a port to need to traverse all the files in ${worksrcpath} to fix their permissions, for example. With the changes from this PR, the following block only prints the symlink, rather than every file and directory inside the directory it links to:
> 
> post-extract {
>     fs-traverse item ${worksrcpath} {
>         ui_msg ${item}
>     }
> }
> 
> It works correctly if fs-traverse is invoked with ${worksrcpath}/ instead of ${worksrcpath}, but it would be better if ports didn't have to make that change. Maybe fs-traverse can be fixed to dereference the symlink first.

In fs-traverse.c, the flag to follow “root” symlinks is explicitly commented out:

  base/src/pextlib1.0/fs-traverse.c:173
  root_fts = fts_open(targets, FTS_PHYSICAL /*| FTS_COMFOLLOW */| FTS_NOCHDIR | FTS_XDEV, &do_compare);

If I’m reading the fts(3) man page correctly, uncommenting `| FTS_COMFOLLOW'  would make fs-traverse behave like find(1) with option `-H’, i.e., follow symbolic links explicitly given on the command line, as opposed to all symbolic links in the hierarchy (option -L).

But rather than uncomment FTS_COMFOLLOW  in the above code, I would argue that it’s safer to provide an explicit option, like find(1) option `-H’. In particular, if PR #55 is committed,  ${workpath} may contain two directories:  a directory and a symbolic link to it.

> Port audio/ID3 is problematic, principally because it sets build.dir:
> 
> build.dir           ${worksrcpath}/ID3
> 
> Portfile assignments such as this are evidently evaluated once and for all at the
> time they're read - in particular, before any stages are run. So although `worksrcpath’
> gets updated following extraction, the value of build.dir does not.
> 
> Exactly. This is perfectly legal and normal Portfile syntax, so my proposal of arranging for worksrcdir to be set to the weird directory name isn't workable. Which leaves us with either renaming or symlinking.

Renaming would have to be done in a particular phase (extract, in this case).  But as Rainer noted, that’s a problem if, say, someone calls `port extract’ followed by `port build’.  The extract phase, and thus renaming, wouldn't occur in the second invocation.  To support this, MacPorts would have to implement a variable cache of some sort.

> We have this in MacPorts, via our lazily-evaluated option defaults. To set an option immediately (evaluates referenced variables immediately):
> 
> configure.args --with-python=${prefix}/bin/python${python_version}
> 
> To defer setting an option until the first time it's evaluated (evaluates referenced variables later):
> 
> default configure.args {--with-python=${prefix}/bin/python${python_version}}

I meant to say that `default’ is apparently not intended for use in Portfiles themselves.  And just to be clear, I’m not advocating for this.  Only that support for Joshua’s implementation would require a design change, which would probably also require something like the caching mechanism alluded to above.

> But we should not require all portfiles to adapt to a worksrcdir whose value can change. That would be a lot of work to fix in existing ports, and it would be something we have to educate each new contributor about.

Agreed.  I think it was a conscious design choice.  The Emacs editor was designed with dynamic scoping as a feature, but today is in the throes of being rewritten with lexical scoping.  Dynamic scoping can become too much of a good thing :)

> So with all that said, and provided the fs-traverse issue is resolved first, I cautiously approve of the symlinking approach used in this PR.

Okay, thank you for your comments.
-AM

slewsys:
As Ryan noted by email, `fs-traverse` was originally designed to follow
symbolic links given on the command line (Ryan's email evidently didn't
get attached to this discussion):

>  We used to follow root symlinks, but it was commented out to fix a problem with delete, which used fs-traverse at the time, to work around a bug on Panther, but that was changed in 2473bc7 after we dropped Panther support. So perhaps that means we can enable FTS_COMFOLLOW for fs-traverse again. MacPorts base has a test suite; we should make sure this change doesn't show any new test failures.

Indeed, `fs-traverse` is still documented this way per portfile(7):

> If fs-traverse is called directly on a symbolic link, the link will be followed. All other links encountered during traversal will not be followed.

However, in the present version of MacPorts, `fs-traverse` does not
follow _symlinks_, which can be verified by adding the following
post-extract step to any port's _Portfile_:

```tcl
post-extract {
    fs-traverse dir ${portpath}/work {
        if {[file isdirectory ${dir}]} {
            puts "FS-TRAVERSE ${dir}"
        }
    }
}
```

and running the command:
```
port -v extract port-name
```
The output is a single line `FS-TRAVERSE ${portpath}/work`.

With this PR, the original behavior would be restored so that running the
above command generates `FS-TRAVERSE` lines for all subdirectories of
_${portpath}/work_.

No regressions are triggered by this PR in either macOS 12 or 13:

```
$ cd /macports/base/tests
$ make test
/opt/libexec/macports/bin/tclsh8.5 ./test.tcl -nocolor
Total:1 Passed:1 Failed:0 Skipped:0  case-insensitive-deactivate
Total:3 Passed:3 Failed:0 Skipped:0  checksums-1
Total:1 Passed:1 Failed:0 Skipped:0  dependencies-a
Total:1 Passed:1 Failed:0 Skipped:0  dependencies-b
Total:1 Passed:1 Failed:0 Skipped:0  dependencies-c
Total:1 Passed:1 Failed:0 Skipped:0  dependencies-d
Total:1 Passed:1 Failed:0 Skipped:0  dependencies-e
Total:1 Passed:1 Failed:0 Skipped:0  envvariables
Total:1 Passed:1 Failed:0 Skipped:0  setuid
Total:1 Passed:1 Failed:0 Skipped:0  site-tags
Total:2 Passed:2 Failed:0 Skipped:0  statefile-unknown-version
Total:2 Passed:2 Failed:0 Skipped:0  statefile-version1
Total:2 Passed:2 Failed:0 Skipped:0  statefile-version1-outdated
Total:2 Passed:2 Failed:0 Skipped:0  statefile-version2
Total:2 Passed:2 Failed:0 Skipped:0  statefile-version2-invalid
Total:2 Passed:2 Failed:0 Skipped:0  statefile-version2-outdated
Total:1 Passed:1 Failed:0 Skipped:0  svn-and-patchsites
Total:1 Passed:1 Failed:0 Skipped:0  trace
Total:2 Passed:2 Failed:0 Skipped:0  universal
Total:1 Passed:1 Failed:0 Skipped:0  variants
Total:2 Passed:2 Failed:0 Skipped:0  xcodeversion
$
```

There is, however, the question of whether enabling `fs-traverse` to
follow symlinks can of itself, or in interaction with the proposed
extension in macports/macports-base#55, cause conflicts.

So far, the answers appear to be _no_ and _no_ (though conflicts do exist, as
noted below). This is based on examination of all TCL code in the MacPorts base
and ports/_resources as well as sampling Portfiles that use fs-traverse, e.g., in
combination with post-extract.

Furthermore, it appears that the post-extract step of the following port
groups wouldn't be necessary following this update:

bitbucket-1.0
github-1.0
octave-1.0

PR macports/macports-base#55 extension does itself conflict with existing
ports per PR macports/macports-ports#1760 as well as `octave-1.0.tcl` 
port group. This would be an issue going forward for port and group authors, per
another of Ryan's email comments (that didn't get attached to this discussion):

> The new code in macports/macports-base#55 runs before any port post-extract blocks, right? So port post-extract blocks that rename the worksrcdir could be updated to first check whether the directory already exists at the new name. If so, we must be running on a new version of base and there's no need for the port to rename it.

In particular, code that is presently written as, e.g.:

```tcl
post-extract {
    move [glob ${workpath}/${name}-${version}-*] ${worksrcpath}
}
```

would, for backward compatibility, have to be implemented as:

```tcl
post-extract {
    if {![file exists ${worksrcpath}]} {
        move [glob ${workpath}/libao-${version}-*] ${worksrcpath}
    }
}
```

raimue:
The tests on Travis failed with your commit as the behavior changed. The tests for fs-traverse are in `src/pextlib1.0/tests/fs-traverse.tcl` and would need to be changed accordingly.

I think the post-extract phases in the port groups would still be necessary as a Portfile could add additional distfiles for download. For example if we remove it from the github port group, we will have to restore this post-extract phase in Portfiles that have additional distfiles. This would cause trouble than just leaving it in the port group in my opinion.

If you add compatibility workarounds to ports, also add a comment why you added them, so we know when they can be removed.

slewsys:
On May 16, 2018, at 11:02 AM, Rainer Müller <notifications@github.com> wrote:
> 
> The tests on Travis failed with your commit as the behavior changed. The tests for fs-traverse are in src/pextlib1.0/tests/fs-traverse.tcl and would need to be changed accordingly.

A number of these ports are actually candidates to be flagged for removal.  For instance:

* FScript - only builds on OS X Lion and earlier.

* dapplib - has compile errors, and is no longer distributed by the author.

* starfighter - is for X11 and is no longer distributed by the author.

* huevos - is almost certainly obsolete functionality and no longer distributed by the author.

* QMK-Groundstation - depends on qt-mac version 4.2


This is all documented in the original commit message, along with the fact that the updates are in preparation for macports/macports-base#55.

What’s new since the original PR is that Ryan requested the updates be made backward-compatible with pre-macports/macports-base#55 baes versions.  That has been completed.  The backward-compatiblility commits aren’t listed separately on GitHub PR macports/macports-ports#1760 because they were squashed into the original PR.  Perhaps the squash was premature, since not doing so would have helped flag the changes.
-AM

slewsys:
On May 16, 2018, at 11:02 AM, Rainer Müller <notifications@github.com> wrote:
> 
> The tests for fs-traverse are in src/pextlib1.0/tests/fs-traverse.tcl and would need to be changed accordingly.

Update for fs-traverse regression test  has been added to PR macports/macports-base#55.
It now expects symbolic links on the command line to be followed.

There are actually three commits listed on GitHub. My git skills are inadequate to combine them, so please disregard the first two…. For the record:

https://github.com/slewsys/macports-base/commit/18d200e4eb4758bb3fbc5187eb5c134095d9ac34 - is a dirty version of fs-traverse regression test update.
https://github.com/macports/macports-base/pull/55/commits/1b968eba5491b8c16dceb383f20f0c0505a7e5d5 - was supposed to be a clean version squashed onto the dirty one.
https://github.com/slewsys/macports-base/commit/8a308416c56f5de0c3891e02e87ce9109797c6b6 -  second attempt at squashing clean version onto dirty.

> I think the post-extract phases in the port groups would still be necessary as a Portfile could add additional distfiles for download. For example if we remove it from the github port group, we will have to restore this post-extract phase in Portfiles that have additional distfiles. This would cause trouble than just leaving it in the port group in my opinion.

Good point.   Port group post-extract can’t be removed as you say.

> If you add compatibility workarounds to ports, also add a comment why you added them, so we know when they can be removed.

All backward compatibility updates include the comment:

# This block can be removed once backward compatibility with MacPorts
# base versions prior to PR macports/macports-base#55 is not needed.

though I’ll remove that comment from octave port group.  Thank you!
-AM

pmetzger:
Squashing is pretty easy. All you do is use `git reset --soft [somerev]` in your branch to "undo" the commits, commit again, and then do a force push of the branch. For example, to squash the last three commits

```
git reset --soft HEAD~3
git commit -m "New message for the combined commit"
git push origin +name-of-branch
```

slewsys:
On May 17, 2018, at 8:57 AM, Perry E. Metzger <notifications@github.com> wrote:
> 
> Squashing is pretty easy. All you do is use git reset --soft [somerev] in your branch to "undo" the commits, commit again, and then do a force push of the branch. For example, to squash the last three commits
> 
> git reset --soft HEAD~3
> git commit -m "New message for the combined commit"
> git push origin +name-of-branch

I did use reset, though not with option `—soft’, perhaps that would have made the difference?  I think reset works for the last commit(s), but in this case, the bad commit was not the last.  To edit a commit other than the last, it may be that `rebase' is called for?

The end result is that only the good commit remains on my fork, whereas on GitHub there are multiple commits.  You’d think that I could use “git pull” to bring those into my branch, but git just reports “Already up to date.”
-AM

slewsys:
On May 18, 2018, at 7:23 AM, Perry E. Metzger <notifications@github.com> wrote:
> 
> If you don't know what git is doing (that is, if you only understand it quite vaguely) you can make a serious mess of things.

Yeah, that’s what I love about git :)


> I recommend taking enormous care at this point. You might want to ask for help on macports-dev.

My best guess is that it’s an interaction with GitHub.  It’d be great if I could fix the problem from my end, but I wouldn’t be surprised if this is not the case.  Live and learn.
-AM

pmetzger:
I doubt it's an interaction with GitHub, for what it's worth...

pmetzger:
At worst, extract all the changes you want into a diff, create a new branch, re-apply the changes.

mojca:
Can someone please provide a summary of what still needs to be done? If it's just a matter of squashing commits into a single one (and potentially clean up some other git mess), I happily volunteer to fix that, just please let me know what the commit message should be and I'll do it.

If there are other reservations about the functionality, can someone please provide a new summary? There was plenty of feedback, but it's not clear to me right now which part of the feedback still applies and what has been addressed.

I would like to see this resolved and would be in favour of committing it as long as setting an explicit `worksrcdir` where necessary still works (`wxWidgets-3.0` is an example of a port that needs to set that variable; I'm also sure that I recently set it somewhere in a "recursive" way, like `worksrcdir ${worksrcdir}/something`, which might also be worth checking).

Also, it would be committed to `master`, so there's still plenty of time to fix any remaining issues afterwards, it's not like with `Portfiles` which fly directly to users. Many more users test functionality once it lands in `master` compared to when the code just sits in a PR.

pmetzger:
@mojca Thank you for helping to bring this to completion. It's appreciated.

slewsys:
>On Oct 2, 2018, at 1:11 PM, Mojca Miklavec <notifications@github.com> wrote:
>Can someone please provide a summary of what still needs to be done? If it's just a matter of squashing commits into a single one (and potentially clean up some other git mess), I happily volunteer to fix that, just please let me know what the commit message should be and I'll do it.

The FScript and Octave patches of macports/macports-ports#1760 are now obsolete.
Otherwise, I'm not aware of any recent PRs or commits that conflict with either macports/macports-base#55 or macports/macports-ports#1760, which are two parts of a single commit.

@raimue 's comment about cleanup refers, I think, to https://github.com/macports/macports-base/pull/55#issuecomment-389763564 where I was unable to squash three commits.  Still not sure how to resolve that...

Having said that,  macports/macports-base#55 consists of three commits (with proposed commit message):

1. [portextract.tcl: If expected path of extract doesn't exist, symlink to actual path.](https://github.com/macports/macports-base/pull/55/commits/4386b6bd306d0246797ddccb1a14d53f900b7964)
2. [fs-traverse.c: Follow symlink on command line.](https://github.com/macports/macports-base/pull/55/commits/782cd571e1c45f5a866f2bd58c1394c118982636)
3. [fs-traverse.tcl: Update tests for fs-traverse.](https://github.com/macports/macports-base/pull/55/commits/1b968eba5491b8c16dceb383f20f0c0505a7e5d5)

portextract.tcl's commit message might add: If worksrcdir is explicitly set, nothing is done.  

Consequently, there's no conflict with, e.g., wxWidgets-3.0.

macports/macports-ports#1760 consists of seven commits (excluding FScript and Octave patches) which resolve conflicts between macports/macports-base#55 and existing ports:

1. [libao: Prepare for PR macports/macports-base#55.](https://github.com/macports/macports-ports/pull/1760/commits/779d5a5a2de690455c1cb290464b985ee4076b28)
2. [daaplib: Prepare for PR macports/macports-base#55.](https://github.com/macports/macports-ports/pull/1760/commits/30d6ba4d677a73d3ab8d2bebcd8d03ad11136b95)
3. [play: Prepare for PR macports/macports-base#55.](https://github.com/macports/macports-ports/pull/1760/commits/4f768fbc4497837e6469da1c03dad7f36369666e)
4. [starfighter: Prepare for PR macports/macports-base#55.](https://github.com/macports/macports-ports/pull/1760/commits/c410d9e3088416d31eb0b85823ef7081f76128a6)
5. [MobileDevice: Prepare for PR macports/macports-base#55.](https://github.com/macports/macports-ports/pull/1760/commits/236b2662eaebadbe4b307dc25924197b6169ffc7)
6. [huevos: Prepare for PR macports/macports-base#55.](https://github.com/macports/macports-ports/pull/1760/commits/161916e5917fbbae17d5f3478ed4d091daf39430)
7. [QMK-Groundstation: Prepare for PR macports/macports-base#55.](https://github.com/macports/macports-ports/pull/1760/commits/36e7af7cc661621df92049387989a51c572ea195)

>There was plenty of feedback, but it's not clear to me right now which part of the feedback still applies and what has been addressed.

The additional feedback was exploring alternative solutions and potential impact. The most significant, perhaps, is Ryan's observation that when updating Portfiles, temporary wrappers for backward-compatibility are needed.  So Portfiles will get bigger before they get smaller. The patches of macports/macports-ports#1760 are examples of this.

Let me know if there's anything more that I can do on my end.  Thank you Mojca and Perry!

mojca:
My question about what needs to be done was mainly referring to this particular PR. There is no reason to squash the commits in macports/macports-ports#1760, only remove those which are no longer relevant.

Also, I assume that there were a lot more commits in the beginning. What I see here are already three clean commits and I see no reason to squash them any further. It's nice if the first line is below 50/70 characters, so I can try to somehow combine that with your proposed commit messages written above, but I see no reason to do anything else.

slewsys:
On Oct 2, 2018, at 5:08 PM, Mojca Miklavec <notifications@github.com> wrote:
> 
> My question about what needs to be done was mainly referring to this particular PR. There is no reason to squash the commits in macports/macports-ports#1760, only remove those which are no longer relevant.

macports/macports-ports#1760 has been updated accordingly.

> Also, I assume that there were a lot more commits in the beginning. What I see here are already three clean commits and I see no reason to squash them any further.

One commit grew to a dozen and is now down to ten (including macports/macports-ports#1760).  Some effort was made to avoid surprises.

> It's nice if the first line is below 50/70 characters, so I can try to somehow combine that with your proposed commit messages written above, but I see no reason to do anything else.

Agreed.  I’ve since found Emacs Magit very helpful in managing commits.  Thank you again!
-AM

pmetzger:
Magit is really wonderful. It's the best front end I know of for Git.

mojca:
There was no further feedback since April/May and my understanding is that feedback to the initial version has been considered, so I'll merge this. (If there are any remaining issues left, they can still be fixed before the release.)

pmetzger:
If this is now merged, then https://github.com/macports/macports-ports/pull/1760 should probably be merged some time soon (after the conflicts are resolved.)

PR #57, opened 2018-01-12
Find port providing ccache binary for trace mode

raimue:
Instead of assuming the ccache binary is provided by the ccache port,
query the registry which port provides ${prefix}/bin/ccache if it exists
and also include all dependencies.

Closes: https://trac.macports.org/ticket/55680

### DIFF
--- a/src/port1.0/portutil.tcl
+++ b/src/port1.0/portutil.tcl
@@ -1304,7 +1304,7 @@ set ports_dry_last_skipped ""
 proc target_run {ditem} {
     global target_state_fd workpath portpath ports_trace PortInfo ports_dryrun \
            ports_dry_last_skipped worksrcpath subport env portdbpath \
-           macosx_version
+           macosx_version prefix
     set portname $subport
     set result 0
     set skipped 0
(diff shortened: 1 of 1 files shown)


ryandesign:
It looks like this changes the behavior if `configure.ccache` is true but ccache is not yet installed. Before, the dependency would be added; now, it isn't.

raimue:
Yes, you are right, there is a slight change, but that is to keep the same behavior with and without trace mode. If `configure.ccache` is set, but the ccache binary cannot be executed, it will be disabled in the configure phase.

Adding ccache to this list means it allows to access files installed by the ccache port. If the ccache port is not installed, no files will be provided by it, so the previous version effectively did not do anything in this case.

However, during testing `recursive_collect_deps` had failed if I asked it to collect dependencies of the ccache port when it was not installed. This caused the build to fail with trace mode although it continued with a warning without trace mode. Therefore, I added the check whether the ccache binary is active in the prefix to avoid the trace mode error if `configure.ccache` is enabled but no port provides the binary. Now it will always produce the same warning that ccache was unusable and disabled during configure with or without trace mode.

PR #58, opened 2018-03-04
portconfigure: fix compiler selection on PPC

kencu:
PPC cannot use clang-3.3 or clang-3.4 successfully
fallback to gcc6 or gcc7 if default compilers are
blacklisted
fixes innumerable tickets to be closed individually

I hope I've done this correctly - it's my first base PR.

### DIFF
--- a/src/port1.0/portconfigure.tcl
+++ b/src/port1.0/portconfigure.tcl
@@ -508,7 +508,7 @@ proc portconfigure::configure_get_default_compiler {} {
 
 # internal function to choose compiler fallback list based on platform
 proc portconfigure::get_compiler_fallback {} {
-    global xcodeversion macosx_deployment_target default_compilers configure.sdkroot configure.cxx_stdlib os.major
+    global xcodeversion macosx_deployment_target default_compilers configure.sdkroot configure.cxx_stdlib os.major configure.build_arch
 
     # Check our override
     if {[info exists default_compilers]} {
@@ -527,9 +527,17 @@ proc portconfigure::get_compiler_fallback {} {
                 return {gcc-4.0}
             }
         } elseif {[vercmp $xcodeversion 3.0] >= 0} {
-            return {gcc-4.2 apple-gcc-4.2 gcc-4.0 macports-clang-3.4 macports-clang-3.3}
+            if {${configure.build_arch} eq "ppc" || ${configure.build_arch} eq "ppc64"} {
+                return {gcc-4.2 apple-gcc-4.2 gcc-4.0 macports-gcc-6 macports-gcc-7}
+            } else {
+                return {gcc-4.2 apple-gcc-4.2 gcc-4.0 macports-clang-3.4 macports-clang-3.3}
+            }
         } else {
-            return {apple-gcc-4.2 gcc-4.0 gcc-3.3 macports-clang-3.3}
+            if {${configure.build_arch} eq "ppc" || ${configure.build_arch} eq "ppc64"} {
+                return {apple-gcc-4.2 gcc-4.0 macports-gcc-6 macports-gcc-7}
+            } else {
+                return {apple-gcc-4.2 gcc-4.0 macports-clang-3.3}
+            }
         }
     }
 


src/port1.0/portconfigure.tcl
```diff
@@ -527,9 +527,17 @@ proc portconfigure::get_compiler_fallback {} {
                 return {gcc-4.0}
             }
         } elseif {[vercmp $xcodeversion 3.0] >= 0} {
-            return {gcc-4.2 apple-gcc-4.2 gcc-4.0 macports-clang-3.4 macports-clang-3.3}
+            if {${configure.build_arch} eq "ppc" || ${configure.build_arch} eq "ppc64"} {
+                return {gcc-4.2 apple-gcc-4.2 gcc-4.0 macports-gcc-6  macports-gcc-7}
```
kencu: I've been thinking whether this should be based configure.build_arch or os.arch. I chose build_arch, as building PPC ports on an Intel machine doesn't work reliably with clang at present.

kencu:
If no-one has any objections here I will go ahead and merge this in later on today..

src/port1.0/portconfigure.tcl
```diff
@@ -527,9 +527,17 @@ proc portconfigure::get_compiler_fallback {} {
                 return {gcc-4.0}
             }
         } elseif {[vercmp $xcodeversion 3.0] >= 0} {
-            return {gcc-4.2 apple-gcc-4.2 gcc-4.0 macports-clang-3.4 macports-clang-3.3}
+            if {${configure.build_arch} eq "ppc" || ${configure.build_arch} eq "ppc64"} {
```
ryandesign: Should this be testing if `${os.arch} eq "powerpc"` instead?

src/port1.0/portconfigure.tcl
ryandesign: There's an extra space there. Why is gcc6 better than gcc7?

src/port1.0/portconfigure.tcl
```diff
@@ -527,9 +527,17 @@ proc portconfigure::get_compiler_fallback {} {
...
+                return {gcc-4.2 apple-gcc-4.2 gcc-4.0 macports-gcc-6  macports-gcc-7}
+            } else {
+                return {gcc-4.2 apple-gcc-4.2 gcc-4.0 macports-clang-3.4 macports-clang-3.3}
+            }
         } else {
-            return {apple-gcc-4.2 gcc-4.0 gcc-3.3 macports-clang-3.3}
+            if {${configure.build_arch} eq "ppc" || ${configure.build_arch} eq "ppc64"} {
+                return {apple-gcc-4.2 gcc-4.0 macports-gcc-6 macports-gcc-7}
```
ryandesign: Why did you remove `gcc-3.3` from the list for 10.4 PowerPC?

mojca:
Unrelated, but you can also fix https://trac.macports.org/ticket/48345 while working on this :)

No objections about that change. As a general remark I'm somewhat inclined towards moving those values to a place that won't require a new release to make the change effective, but that's a separate issue. (I guess that you'll want to add gcc 8 when that one gets released. The list of compilers for newer systems is somewhat limited as well.)

src/port1.0/portconfigure.tcl
kencu: I wondered the same thing for a bit -- but clang on an Intel machine can't make useful PPC code either, I figured. If you prefer the test to be os.arch I don't mind. Either will work on the PPC machines.

src/port1.0/portconfigure.tcl
kencu: at present, gcc7 has about 4x as many test failures on PPC as gcc6 does. Also, it's more strict and so raises more unhelpful-to-the-end-user build errors.

src/port1.0/portconfigure.tcl
kencu: just thought nobody cared about gcc-3.3 for anything anymore, and we would never want it to be called in. I can add it back if you know of something that builds with it.

kencu:
I would dearly love to find a place to do this configuration that was not in base. Is there some other place that would roll out in a selfupdate? I couldn't find it, if so.

src/port1.0/portconfigure.tcl
ryandesign: I care about the inconsistency between Intel and PowerPC. I haven't seen a ticket or mailing list thread discussing dropping gcc-3.3, but if we want to drop it, shouldn't we always drop it, and not just on PowerPC?

src/port1.0/portconfigure.tcl
ryandesign: I remember you mentioned gcc7 problems on PowerPC before, but 7.3.0 was released since then. Does it still have those failures?

Newer clang compilers are also more strict than older clang compilers, yet we don't shun them. We deal with the problems, fix the build failures, report the problems to the developers...

src/port1.0/portconfigure.tcl
ryandesign: I guess you're right, it should stay as it is.

ryandesign:
> I would dearly love to find a place to do this configuration that was not in base. Is there some other place that would roll out in a selfupdate? I couldn't find it, if so.

No; these changes must be made in base.

mojca:
To clarify: yes, the changes definitely need to be made in base. That doesn't mean that we could not move them elsewhere in the future to allow changing these values in an easier way.

ryandesign:
I don't see why they should be moved out of base. We don't need to change them that often.

kencu:
Thanks everyone. I'll make the requeste changes.

src/port1.0/portconfigure.tcl
kencu: I will admit to not having run a full test suite on gcc 7.3.0 on PPC as yet. It's a two-day affair of building and testing :>  I can do that in a bit and recheck the status of the current gcc7, as we might as well get this done right once.

src/port1.0/portconfigure.tcl
kencu: True. will fix.

kencu:
I removed the extra space, and deleted gcc-3.3 as a fallback compiler on 10.4 Intel as suggested. I think this is good to go now therefore. Testing gcc7 and deciding if it should become the default compiler on PPC will take me about three more weeks, given that I'm presently in Sunny Puerto Vallarta and have no reasonable access to machinery here.

PR #59, opened 2018-03-07
port mirror: Respect special license NoMirror

raimue:
Use this license to prevent mirroring of ports which license does not
allow to host their distfiles on our servers.

### DIFF
--- a/src/port1.0/portmirror.tcl
+++ b/src/port1.0/portmirror.tcl
@@ -51,7 +51,12 @@ namespace eval portmirror {
 # It also records the path in a database.
 
 proc portmirror::mirror_main {args} {
-    global fetch.type mirror_filemap portdbpath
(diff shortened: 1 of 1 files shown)


jmroot:
Seems reasonable for this to be the default behaviour, but it couldn't hurt to have an option to mirror everything anyway. This could be used if someone creates a mirror that will only be used internally by their company for example.

raimue:
Sounds good. How about `port mirror --ignore-license` to force mirroring? Or should it rather use the generic global `-f` or local `--force` option?

raimue:
This can be merged in this state, but please open a new ticket to remind us to add the `--ignore-license` option to the port cli.

The special license `NoMirror` also needs to be documented in the guide (with a reference that it is supported as of MacPorts 2.5.0).

neverpanic:
https://trac.macports.org/ticket/56025

PR #61, opened 2018-03-12
reclaim.tcl: don't ask for partial removal

g5pw:
This patch eliminates the possibility for the user to choose which
unrequested/inactive port to remove in the reclaim phase. Now the user
can either remove all ports on none at all.

This was done so we can safely force-remove all ports and avoid asking
confirmation to the user.

### DIFF
--- a/src/macports1.0/reclaim.tcl
+++ b/src/macports1.0/reclaim.tcl
@@ -456,20 +456,13 @@ namespace eval reclaim {
             ui_msg "Found no inactive ports."
 
         } else {
-
-            if {[info exists macports::ui_options(questions_multichoice)]} {
-                set retstring [$macports::ui_options(questions_multichoice) "Would you like to uninstall these inactive ports?" "" $inactive_names]
-
-                if {[llength $retstring] > 0} {
-                    foreach i $retstring {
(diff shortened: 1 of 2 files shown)


src/macports1.0/reclaim.tcl
```diff
@@ -527,20 +521,15 @@ namespace eval reclaim {
...
-            if {[info exists macports::ui_options(questions_multichoice)]} {
-                set retstring [$macports::ui_options(questions_multichoice) "Would you like to uninstall these ports?" "" $unnecessary_names]
-
-                if {[llength $retstring] > 0} {
-                    foreach i $retstring {
-                        set name [lindex $unnecessary_names $i]
-                        set port $ports($name)
-
-                        ui_msg "Uninstalling: $name"
+            ui_msg " [join $unnecessary_names {, }]."
```
neverpanic: This seems like a very unreadable way to print the ports. Did you test that this will actually stay readable with many ports?

g5pw: This is actually an error, that line should go.

neverpanic:
Thank you for your contribution!

PR #62, opened 2018-03-13
portconfigure: fix compiler selection on PPC

kencu:
PPC cannot use clang-3.3 or clang-3.4 successfully
fallback to gcc6 or gcc7 if default compilers are blacklisted
fixes innumerable tickets to be closed individually

Also - deleted gcc-3.3 from 10.4 Intel as a compiler fallback choice
for consistency with the PPC changes

### DIFF
--- a/src/port1.0/portconfigure.tcl
+++ b/src/port1.0/portconfigure.tcl
@@ -473,7 +473,7 @@ proc portconfigure::configure_get_default_compiler {} {
 
 # internal function to choose compiler fallback list based on platform
 proc portconfigure::get_compiler_fallback {} {
-    global xcodeversion macosx_deployment_target default_compilers configure.sdkroot configure.cxx_stdlib os.major
+    global xcodeversion macosx_deployment_target default_compilers configure.sdkroot configure.cxx_stdlib os.major configure.build_arch
 
     # Check our override
     if {[info exists default_compilers]} {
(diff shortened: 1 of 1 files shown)


ryandesign:
Doesn't this break C++ ports on PowerPC? You cannot share objects between C++ libraries; you will crash if you try. And this change makes it so that some ports (for example some library *foolib*) might compile with /usr/bin/g++-4.2 (and link with /usr/lib/libstdc++.6.dylib) as they did before, while ports that blacklist old GCC (for example a program *barprog* that uses the aforementioned library) might compile with /opt/local/bin/g++-mp-6 (and link with /opt/local/lib/libstdc++.6.dylib). When the *barprog* program allocates a C++ object using /usr/lib/libstdc++.6.dylib and passes it to the *foolib* library, and it tries to work with that object using /opt/local/lib/libstdc++.6.dylib, it will crash.

We have had a longstanding policy of not using FSF GCC as a general purpose compiler for exactly this reason and I'm very nervous not only about this change in general but also that this change was immediately merged to the stable 2.4 branch.

The only reason why we're able to use gcc6's C++ compiler in the cxx11 1.1 portgroup is that we also put `-D_GLIBCXX_USE_CXX11_ABI=0` into `CXXFLAGS`. Don't we need to do that in MacPorts base now as well, when macports-gcc-6 is chosen due to the above changes (but *not* when it is chosen in any other existing way, as that would break all the ports that explicitly use gcc6 today)? And if so, how do we do that?

PR #64, opened 2018-03-14
portindex: return >0 value if some ports failed

l2dy:
https://trac.macports.org/wiki/Meetings/MacPortsMeeting2018/GSoC#Hackingsessionideas

### DIFF
--- a/src/port/portindex.tcl
+++ b/src/port/portindex.tcl
@@ -299,3 +299,7 @@ puts "\nTotal number of ports parsed:\t$stats(total)\
       \nPorts successfully parsed:\t[expr {$stats(total) - $stats(failed)}]\
       \nPorts failed:\t\t\t$stats(failed)\
       \nUp-to-date ports skipped:\t$stats(skipped)\n"
+
+if {$stats(failed) > 0} {
+    exit 1
+}


l2dy:
@raimue Can you check the failing test? I'm still on my way home from the meeting.

raimue:
I can only assume that `portindex` in this test already failed on this port before, but the error was silently ignored, in which case this change was a good idea to detect it.

The error can be reproduced by running `portindex` on an empty Portfile.

raimue:
Should be fixed as of e97d916d.

jmroot:
I still think this may cause problems elsewhere. For example, `port sync` will now print an error saying "updating PortIndex for $source failed" if some ports failed to parse, which is not entirely true and may alarm users. The fact that some ports failed to parse is already reported by portindex itself.

I really think someone needs to review all the places that portindex is used if this is going to be the default behaviour. And we would need to inform users that their own scripts may break.

Another problem is that there's no distinction between some portfiles failing to parse and portindex failing entirely due to e.g. lack of permissions.

danielluke:
If no one wants to review every place portindex is used (and/or if we don't want to communicate this out to users and deal with any support issues from the behavior change for them) - this could be implemented with an optional command line argument (sort of the inverse of port's -p).

jmroot:
See https://lists.macports.org/pipermail/macports-users/2018-March/044842.html

l2dy:
>See https://lists.macports.org/pipermail/macports-users/2018-March/044842.html

The sync command still returns 0.

raimue:
As I understand it, @jmroot wants to be able to differentiate between fatal errors and errors while parsing some ports. This is reasonable, as in this state it would not even update `PortIndex.quick` when `portindex` failed. `port sync` not reflecting any error in the exit code is a different issue.

portindex could have different exit codes that base can interpret (for example 1 for fatal error, 2 for failed parsing of some ports).

Or we go with the suggestion made in https://github.com/macports/macports-base/pull/64#issuecomment-373532033. For example, a `portindex -e` flag that causes portindex to return a non-zero exit code when parsing of some ports failed.

l2dy:
Should I change the return code to 2 for now?

PR #66, opened 2018-03-20
Add depends_patch to handle the patch dependencies of ports

kurthindenburg:
This can be used to handle xz patches in the below trac ticket (which
requires another patch to base).

See https://trac.macports.org/ticket/52445

### DIFF
--- a/src/macports1.0/macports.tcl
+++ b/src/macports1.0/macports.tcl
@@ -3560,8 +3560,8 @@ proc macports::_deptypes_for_target {target workername} {
     switch -- $target {
         fetch       -
         checksum    {return depends_fetch}
-        extract     -
-        patch       {return "depends_fetch depends_extract"}
+        extract     {return "depends_fetch depends_extract"}
+        patch       {return "depends_fetch depends_extract depends_patch"}
         configure   -
         build       {return "depends_fetch depends_extract depends_build depends_lib"}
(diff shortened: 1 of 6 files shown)


src/macports1.0/macports.tcl
```diff
@@ -3560,8 +3560,8 @@ proc macports::_deptypes_for_target {target workername} {
     switch -- $target {
         fetch       -
         checksum    {return depends_fetch}
-        extract     -
-        patch       {return "depends_fetch depends_extract"}
+        extract     {return "depends_fetch depends_extract"}
+        patch       {return "depends_fetch depends_extract depends_patch"}
```
ryandesign: Don't you need to add `depends_patch` to all the subsequent phases as well?

src/port1.0/portutil.tcl
```diff
@@ -1405,8 +1405,8 @@ proc target_run {ditem} {
                     switch $target {
                         fetch       -
                         checksum    { set deptypes "depends_fetch" }
-                        extract     -
-                        patch       { set deptypes "depends_fetch depends_extract" }
+                        extract     { set deptypes "depends_fetch depends_extract" }
+                        patch       { set deptypes "depends_fetch depends_extract depends_patch" }
```
ryandesign: And the same thing here.

PR #67, opened 2018-03-22
Add the ability to handle xz patch files

kurthindenburg:
closes https://trac.macports.org/ticket/52445

### DIFF
--- a/src/port1.0/portpatch.tcl
+++ b/src/port1.0/portpatch.tcl
@@ -93,12 +93,20 @@ proc portpatch::patch_main {args} {
 
     set gzcat "[findBinary gzip $portutil::autoconf::gzip_path] -dc"
     set bzcat "[findBinary bzip2 $portutil::autoconf::bzip2_path] -dc"
+    catch {set xzcat "[findBinary xz $portutil::autoconf::xz_path] -dc"}
+
     foreach patch $patchlist {
         ui_info "$UI_PREFIX [format [msgcat::mc "Applying %s"] [file tail $patch]]"
         switch -- [file extension $patch] {
             .Z -
             .gz {command_exec patch "$gzcat \"$patch\" | (" ")"}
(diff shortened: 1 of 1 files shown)


src/port1.0/portpatch.tcl
```diff
@@ -93,12 +93,20 @@ proc portpatch::patch_main {args} {
...
         switch -- [file extension $patch] {
             .Z -
             .gz {command_exec patch "$gzcat \"$patch\" | (" ")"}
             .bz2 {command_exec patch "$bzcat \"$patch\" | (" ")"}
+            .xz {
+                if {[info exists xzcat]} {
+                    command_exec patch "$xzcat \"$patch\" | (" ")"
+                } else {
+                    return -code error [msgcat::mc "xz binary not found; port needs to add 'depends_patch bin:xz:xz'"]
+                }}
```
ryandesign: Are we expecting the port author to add that dependency? I would expect MacPorts base to do so, just as I would have hoped base was already adding a dependency on `bin:gzip:gzip` for .gz patches or `bin:bzip2:bzip2` for .bz2 patches.

jmroot: Would be nice to have the dependency auto-added, sure. Even without that, this is still a step forward. We don't add deps on gzip or bzip2 to ports anywhere because base already depends on them so pervasively.

kurthindenburg:
Should I commit this as-is?  or try to see if it is possible to auto-add the dep?

ryandesign:
Commit it as-is. Currently, the xz dependency for distfiles gets added when `use_xz yes` is used, and no equivalent exists for patchfiles. In future, I want to [auto-detect distfile type](https://trac.macports.org/ticket/50969) and add needed dependencies based on filename extension. When we do that for distfiles, we can also do it for patchfiles.

PR #68, opened 2018-03-24
Add check that port can be used on current platform

kurthindenburg:
Implement a check that the current platform is in the port's
platforms list.

This only handles the current platforms statement in the Portfile.

see https://trac.macports.org/ticket/15712

### DIFF
--- a/src/macports1.0/macports.tcl
+++ b/src/macports1.0/macports.tcl
@@ -2019,7 +2019,8 @@ proc _mportexec {target mport} {
     # xxx: set the work path?
     set workername [ditem_key $mport workername]
     $workername eval {validate_macportsuser}
-    if {![catch {$workername eval "check_variants $target"} result] && $result == 0 &&
+    if {![catch {$workername eval {check_supported_platforms}} result] && $result == 0 &&
+        ![catch {$workername eval "check_variants $target"} result] && $result == 0 &&
         ![catch {$workername eval {check_supported_archs}} result] && $result == 0 &&
         ![catch {$workername eval "eval_targets $target"} result] && $result == 0} {
         # If auto-clean mode, clean-up after dependency install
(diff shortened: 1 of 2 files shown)


kurthindenburg:
This is a WIP to see if this is how we want to do this.  This can be committed as-is or we can wait until I figure out all the conditionals in the trac ticket.  I'll also have to a lot of tests to verify it all works.

A wrong platform will error out before the arch check.

Note: the _mportexec check is for the dependencies install.

pmetzger:
Does this handle things like `platforms    {darwin >=8}` ?

jmroot:
No, it doesn't. This should not be merged in its current form because it does not handle `platforms macosx` correctly, and because we don't want all ports to error out on non-darwin platforms anyway. We should be looking at which platforms are known not to work, rather than which ones are not known to work.

kurthindenburg:
I don't think I understand exactly what needs done here so I'll close this for now.

pmetzger:
I think what was being asked for was the ability to specify conditionals describing which platforms and os versions to build or not to build on. We do desperately need a facility like this, though.

pmetzger:
The question isn't one of "did the maintainer test it". I have ports like OCaml that I know do not run on certain OSes but I get build errors for them anyway every time I touch them, which creates noise and confusion and wastes time on the build systems. The goal is to stop trying to build on things where we know building will not work and can never work. There are a lot of such things.

danielluke:
Ok. Making life easier for maintainer / macports-build infrastructure related and not a feature for end-users.

I think we just shouldn't bother building stuff for so many (old/non-supported/not being updated) OS releases.

pmetzger:
That's another argument to be had. For now, I'd like to mark that OCaml can't build on 10.4 and 10.5.

PR #69, opened 2018-04-01
upgrade: Show error if trying to upgrade current pseudo-port

raimue:
```
$ sudo port upgrade outdated
Nothing to upgrade.
```

```
$ ls Portfile
ls: Portfile: No such file or directory
$ sudo port upgrade
Error: port upgrade needs at least one port name to upgrade
```

```
$ ls Portfile
Portfile
$ sudo port upgrade
Error: port upgrade does not work with the 'current' pseudo-port
```

Closes: https://trac.macports.org/ticket/15186

### DIFF
--- a/src/port/port.tcl
+++ b/src/port/port.tcl
@@ -2767,15 +2767,34 @@ proc action_upgrade { action portlist opts } {
         return 1
     }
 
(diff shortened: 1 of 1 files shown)


raimue:
This is the same message we already use for other commands, such as `port variants --index`. I am not sure how to express that in a better way...

raimue:
Closing, we seem to need more discussion to decide what the behavior should be.
See https://trac.macports.org/ticket/15186

PR #72, opened 2018-04-04
macports.conf: Allow patterns for host matching

raimue:


src/macports1.0/macports.tcl
```diff
@@ -654,8 +654,7 @@ proc mportinit {{up_ui_options {}} {up_options {}} {up_variations {}}} {
         macports::macosx_deployment_target \
         macports::archivefetch_pubkeys \
         macports::ping_cache \
-        macports::host_blacklisted \
-        macports::host_preferred \
+        macports::host_cache \
```
neverpanic: `host_cache` does not seem to be used in mportinit, or did I miss it?

`host_backlist` and `preferred_hosts`, however, are – although I guess technically it was never necessary to add them to this global statement anyway?

src/macports1.0/macports.tcl
```diff
@@ -5219,12 +5213,25 @@ proc macports::revupgrade_buildgraph {port stackname adjlistname revadjlistname
...
-    global macports::ping_cache macports::host_blacklisted macports::host_preferred
-    if {[info exists host_blacklisted($host)]} {
-        return -1
-    } elseif {[info exists host_preferred($host)]} {
-        return 1
-    } elseif {[info exists ping_cache($host)]} {
+    global macports::ping_cache macports::host_cache \
+           macports::host_blacklist macports::preferred_hosts
+
+    if {[info exists host_cache($host)]} {
```
neverpanic: Shouldn't there be an `array set macports::host_cache {}` to initialize the array somewhere?

src/macports1.0/macports.tcl
```diff
@@ -5219,12 +5213,25 @@ proc macports::revupgrade_buildgraph {port stackname adjlistname revadjlistname
...
+        return $host_cache($host)
+    }
+    foreach pattern $macports::host_blacklist {
+        if {[string match $pattern $host]} {
```
neverpanic: We should probably make this `string match -nocase`.

src/macports1.0/macports.tcl
```diff
@@ -5219,12 +5213,25 @@ proc macports::revupgrade_buildgraph {port stackname adjlistname revadjlistname
...
+            set host_cache($host) -1
+            return -1
+        }
+    }
+    foreach pattern $macports::preferred_hosts {
+        if {[string match $pattern $host]} {
```
neverpanic: Same here.

src/macports1.0/macports.tcl
raimue: Technically not necessary here, but I thought I should list it for completeness.

src/macports1.0/macports.tcl
raimue: Technically also not necessary, but that would be something for `mportinit`.

src/macports1.0/macports.tcl
raimue: Sounds reasonable.

raimue:
Updated to address initializing `host_cache` as array and adding `-nocase`.

PR #73, opened 2018-04-04
port: -D also accepts port names

raimue:


src/port/port.tcl
```diff
@@ -5693,6 +5693,29 @@ if {[catch {mportinit ui_options global_options global_variations} result]} {
     fatal "Failed to initialize MacPorts, $result"
 }
 
+# Change to port directory if requested
+if {[info exists global_options(ports_dir)]} {
+    set dir $global_options(ports_dir)
+    if {[string first "/" $dir] == -1} {
```
neverpanic: Isn't this a change of behavior when specifying relative paths to port directories, e.g. when I'm in `/opt/myportstree/devel` and I specify `port -D foobar`, I may end up in `/opt/dports/textproc/foobar` if that's where portindex thinks the `foobar` port is.

This may not be a big issue in practice, but I'd like to make sure we're aware of it before merging this.

raimue: I don't think this will be an issue. You can always force to use the directory with -D ./foobar` or `-D foobar/`.

neverpanic: I agree, but it is a change of behavior. Anyway, `-D` isn't used very often, so I don't think that's an issue in this case.

jmroot:
Could you explain why you want to be able to do this?

src/port/port.tcl
```diff
@@ -5693,6 +5693,29 @@ if {[catch {mportinit ui_options global_options global_variations} result]} {
...
+        if {[catch {mportlookup $portname} result]} {
+            ui_debug $::errorInfo
+            fatal "lookup of portname $portname failed: $result"
+        }
+        if {[llength $result] < 2} {
+            ui_error "port -D failed to look up $portname: no such port"
+            exit 1
+        }
+        array set portinfo [lindex $result 1]
+        set dir $portinfo(portdir)
```
raimue: I just noticed this is flawed and it only worked by accident in my tests. The `portdir` here is a path relative to the root of the ports tree, but I wanted an absolute path.

raimue:
The previously mentioned problem has been fixed. Commits can be squashed before merging.

PR #77, opened 2018-04-14
pextlib: Remove dead code for base32

raimue:
This is not used anywhere in base.


pmetzger:
LGTM. If it isn't used, it isn't needed.

PR #79, opened 2018-04-14
Add dependency from tclPkgConfig.c to Makefile

emcrisostomo:
tclPkgConfig.c is not recompiled if the configuration changes (e.g.: the prefix is changed), resulting in libtcl containing an incorrect configuration.

### DIFF
--- a/vendor/tcl8.5.19/unix/Makefile.in
+++ b/vendor/tcl8.5.19/unix/Makefile.in
@@ -1150,7 +1150,7 @@ tclPkg.o: $(GENERIC_DIR)/tclPkg.c
 # every path can be configured separately we do not remember one general
 # prefix/exec_prefix but all the different paths individually.
 
-tclPkgConfig.o: $(GENERIC_DIR)/tclPkgConfig.c
+tclPkgConfig.o: Makefile $(GENERIC_DIR)/tclPkgConfig.c
 	$(CC) -c $(CC_SWITCHES)					\
 		-DCFG_INSTALL_LIBDIR="\"$(LIB_INSTALL_DIR)\"" \
 		-DCFG_INSTALL_BINDIR="\"$(BIN_INSTALL_DIR)\"" \


slewsys:
Autoconf typically handles this with a generated `config.h` that's pulled in by a top-level header.  Then every C source file that includes the top-level header is recompiled when config.h gets updated.

Following that strategy, the correct fix might be to add the generated `config.h` to `vendor/tcl8.5.19/generic/tclPort.h`.

slewsys:
Upon further inspection, it appears that MacPorts explicitly skips generating tclConfig.h in favor of tclConfig.sh, presumably as an optimization.  PR #80 reverts to using tclConfig.h, which I think is the correct solution, barring tclConfig.sh having additional benefits that I'm not aware of.

raimue:
First of all, this patch actually needs to be submitted to [Tcl](https://core.tcl.tk/tcl/) upstream.

The sources in `vendor/` are just extracted from the source tarball as provided by upstream. We cannot patch these files, as we would revert the changes when upgrading to a new version.

If we want to apply patches to `vendor/`, we need to figure out a way to do so during configure. For reference, the discussion which decided on the way the vendor source is added can be found in https://github.com/macports/macports-base/pull/6.

raimue:
See https://github.com/macports/macports-base/pull/80#issuecomment-381364837 for a proposed solution on how to keep track of patches.

pmetzger:
@raimue #80 is now closed so presumably we need another option.

raimue:
The solution discussed on #80 is still valid for this pull request. When modifying files in `vendor/`, please also keep a patch in the repository, so we can apply it again on the next update of the Tcl sources.

pmetzger:
@raimue So just to be clear, what is proposed is to add a `vendor/patches` directory and to explicitly add this patch in there for record keeping purposes?

raimue:
Yes, exactly. I would make it `vendor/patches/tcl` so it is clear to which codebase this needs to be applied without hardcoding that in the patch.

slewsys:
> On Apr 19, 2018, at 12:55 PM, Perry E. Metzger <notifications@github.com> wrote:
> 
> @raimue So just to be clear, what is proposed is to add a vendor/patches directory and to explicitly add this patch in there for record keeping purposes?
> 


Hi Perry,
One issue that I ran up against in #80 is that the tcl distribution doesn’t leverage automake's depcomp <https://www.gnu.org/software/automake/manual/html_node/Dependencies.html> tool, which automatically identifies dependencies. Instead, it declares dependencies statically.  So my thinking is that the proper solution is to propose to tcl upstream a patch that simplifies their build suite.  But they may have their own reasons for resisting such an update.  One reason:  it’s not clear to me if there’s a single “upstream”.

Encrio’s proposed patch, #79, which adds another static dependency is what piqued my curiosity, because it shouldn’t be necessary. 

The other issue is that MacPorts configure script itself creates a tcl Makefile and tclConfig.sh (see VENDOR_* variables in configure.ac), rather than running tcl’s own configure scripts.  It’s worked for many years, until Enrico stumbled on it.

So #79 is really a symptom of a lot of bad hacks.  The whole build suite should be fixed and made more robust.  But that’s a fairly big proposal, perhaps more from a political standpoint than actual code.

Having said that either #79 works, or running from MacPorts/base/coonfigure  `make -C vendor clean’ works.  Neither is ideal.  So take your pick.
-AM

slewsys:
> On Apr 19, 2018, at 12:55 PM, Perry E. Metzger <notifications@github.com <mailto:notifications@github.com>> wrote:
> 
> @raimue So just to be clear, what is proposed is to add a vendor/patches directory and to explicitly add this patch in there for record keeping purposes?
> 


Hi Perry,
One issue that I ran up against in #80 is that the tcl distribution doesn’t leverage automake's depcomp <https://www.gnu.org/software/automake/manual/html_node/Dependencies.html> tool, which automatically identifies dependencies. Instead, it declares dependencies statically.  So my thinking is that the proper solution is to propose to tcl upstream a patch that simplifies their build suite.  But they may have their own reasons for resisting such an update.  One reason:  it’s not clear to me if there’s a single “upstream”.

Encrio’s proposed patch, #79, which adds another static dependency is what piqued my curiosity, because it shouldn’t be necessary. 

The other issue is that MacPorts configure script itself creates a tcl Makefile and tclConfig.sh (see VENDOR_* variables in configure.ac), rather than running tcl’s own configure scripts.  It’s worked for many years, until Enrico stumbled on it.

So #79 is really a symptom of a lot of bad hacks.  The whole build suite should be fixed and made more robust.  But that’s a fairly big proposal, perhaps more from a political standpoint than actual code.

Having said that either #79 works, or running from MacPorts/base/coonfigure  `make -C vendor clean’ works.  Neither is ideal.  So take your pick.
-AM

jmroot:
> The other issue is that MacPorts configure script itself creates a tcl Makefile and tclConfig.sh (see VENDOR_* variables in configure.ac), rather than running tcl’s own configure scripts.
I think you're misinterpreting what our configure does. MP_CONFIG_SUBDIR runs the configure script in the given subdir.

slewsys:
On Apr 20, 2018, at 3:06 AM, Joshua Root <notifications@github.com> wrote:
> 
> The other issue is that MacPorts configure script itself creates a tcl Makefile and tclConfig.sh (see VENDOR_* variables in configure.ac), rather than running tcl’s own configure scripts.
> I think you're misinterpreting what our configure does. MP_CONFIG_SUBDIR runs the configure script in the given subdir.

I see. I was looking at AC_CONFIG_FILES which creates vendor/Makefile and vendor/tclsh and wrongly assumed that it created tclConfig.sh as well.

mojca:
This 9 character patch has been waiting in queue for 7 months. Is adding an explicit patch all that's needed? If so, I can amend the fix myself. If not, can someone please explain what else would be needed to make this PR queue shorter?

pmetzger:
I think adding an explicit patch is what is needed, yes.

neverpanic:
Ah, and here I was a while ago, wondering if had messed up my local test installation when I reconfigured for a different prefix and the old prefix was still used after installing…

I had a look at the current list of files, and unfortunately, it seems using the configure files in `tcl/macosx` is not the way to go. Citing the upstream source installation instructions at https://www.tcl-lang.org/doc/howto/compile.html (highlight mine):

> Compiling for Macintosh
> 
> [...] For Mac OS X, **use the unix instructions above** [...]

So I guess that explains why we never used the `tcl/macosx` subdirectory.

From my PoV, the issue really is an upstream bug – the documented method to build from source on macOS does not work correctly when re-running configure with different arguments. I'd work around the problem by running `make distclean` whenever we reconfigure. See #190 for that.

neverpanic:
I fixed this by running distclean before configure, see #190.

PR #80, opened 2018-04-14
Use tclConfig.h instead of tclConfig.sh.

slewsys:
MacPorts generates tclConfig.sh instead of tclConfig.h.  This prevents tcl from being properly rebuilt after running base/configure, e.g., with a new --prefix as reported in PR #79.

The attached diffs were created by adding `m4_define(SC_USE_CONFIG_HEADERS)` at the top of `vendor/tcl8.5.19/unix/configure.in` then running `autoreconf` in `vendor/tcl8.5.19/unix` to rebuild `vendor/tcl8.5.19/unix/configure` and `vendor/tcl8.5.19/unix/tclConfig.h.in`.  That's it.


raimue:
Same comment as on https://github.com/macports/macports-base/pull/79#issuecomment-381360656 applies here as well. We cannot patch code in `vendor/`.

jmroot:
Or rather, we can't *just* patch code in vendor/. We need to keep our changes distinct from the standard upstream distribution.

raimue:
Okay, I would be fine with that approach of just saving a copy of patches in `vendor/tcl-patches` that possibly need to be re-applied on each update if not fixed upstream yet.

slewsys:
Applying local patches to upstream sources is one thing.  In the case of modifying `unix/configure.in` though, it requires the additional step of running `autoreconf`, which is less than ideal.

Another approach (which could be submitted upstream) is to have `unix/configure` accept a MacPorts environment variable.  But that is probably further from ideal.

Perhaps simply documenting the problem is the least painful solution?  I.e., note somewhere that after `git pull` requires `make clean` before `configure`.

Rainer's suggestion of using `git worktree` is a nice alternative.

raimue:
So, back to your proposed patch. As you describe it, this creates a `tclConfig.h` in addition to `tclConfig.sh`, but only for the purpose to force a rebuild after running configure. I do not understand the implications of this. There must be a reason why it is normally disabled.

I think the patch would better be reviewed by upstream. The problem of re-running configure must exist outside of MacPorts base as well. Maybe upstream agrees it would make sense to always generate this header?

slewsys:
tclConfig.h and tclConfig.sh are mutually exclusive.  tclConfig.h is actually the "official" way to build tcl for macOS evidently.  And whoever implemented tclConfig.sh as an optimization probably didn't appreciate the implications when tcl is distributed as a third-party package.  I suppose this might be sufficient grounds to recommend that tclConfig.sh be removed as the default.

Unfortunately, tcl's macosx/Makefile is missing some targets needed by MacPorts.  So updating tcl's macosx build to support MacPorts might be another option that could be accepted upstream. Then in MacPorts base/configure.ac change MP_CONFIG_SUBDIR as follows:
```
MP_CONFIG_SUBDIR([vendor/tcl/macosx], [--prefix=${macports_libexec_dir} --enable-threads])
VENDOR_TCL_SUBDIR=tcl/macosx
```

slewsys:
On Apr 14, 2018, at 8:03 PM, Rainer Müller <notifications@github.com> wrote:
> 
> So, back to your proposed patch. As you describe it, this creates a tclConfig.h in addition to tclConfig.sh, but only for the purpose to force a rebuild after running configure. I do not understand the implications of this. There must be a reason why it is normally disabled.

The tclConfig.sh method invokes the compiler with an argument list of macros:  -DARG1=VAL1 -DARG2=VAL2 …

The tclConfig.h methods defines the macros in standard header file format (i.e., #define ARG1 VAL1)  and the compiler is invoked with -DHAVE_TCL_CONFIG_H, so that tclConfig.h gets included in tclPort.h via:

#ifdef HAVE_TCL_CONFIG_H
# include "tclConfig.h"
#endif

Let’s  assume for the moment that the command-line macros invoked by tclConfig.sh are in one-to-one correspondence with the macros defined in tclConfig.h (yet to be verified).  My claim is that  if the macros change, the tclConfig.sh method, which does not update tclConfig.h, has no way of telling make(1) which object files are out of date.

With the tclConfig.h method, In theory, updating tclConfig.h should trigger a rebuild of all sources that include it directly or indirectly.  In particular, tclPkgConfig.c includes tclInt.h which includes tclPort.h, so all is well.  And just running `touch vendor/tcl8.5.19/unix/tclConfig.h` should force a rebuild . Except that, unfortunately, it doesn’t.

Okay, that’s a problem that should probably be fixed.  But for the purposes of MacPorts, it brings us back to Enrico’s PR #79 which may be the best solution after all.  In particular, the macro CFG_INSTALL_LIBDIR appears to be the only one provided by MacPorts/base/configure, and it's  used exclusively for building tclPkgConfig.c, so adding a single dependency, albeit a Makefile, seems reasonable.

Adding to MacPorts/base/configure  the command `make -C vendor clean` works too, but forces most or all of tcl to be rebuilt.  On the other hand, when configure is run, that’s generally desirable.

As time permits, I’ll look closer at tcl's autoconf configuration, but until then, I have to retract PR #80 which doesn’t work as expected.
-AM

raimue:
Thank you for the detailed explanation. That would work if `tclConfig.h` was also listed as a dependency of every `*.c` in Makefiles. However, it seems like that requires a manual run of `make depend` to update.

I am closing #80 for now. We can always reopen it once we figured out what we want to change.

PR #81, opened 2018-04-20
Allow sandbox access to /dev/ptmx and /dev/ttys*

ryandesign:
This allows [Expect](https://core.tcl.tk/expect), which is sometimes needed to automate interaction with particularly pesky programs that require user input, to be used in a Portfile.

See https://lists.macports.org/pipermail/macports-dev/2018-April/038382.html

It would probably be good to have a test that verifies that expect works properly in a Portfile, but I'm not well-versed in creating tests.


fhgwright:
The gpsd port doesn't use expect, but it does include tests that use PTYs by default.  The tests pass because they fall back to using UDP when PTYs aren't available (which can also happen if PTYs run out).

If you run "port test gpsd" with the released MacPorts and look at the log, you'll see a bunch of messages of the form:

gpsfake: ptys not available, falling back to UDP.

If your change has the desired effect, it should make those go away.

ryandesign:
> If your change has the desired effect, it should make those go away.

Hmm, no, my change does not completely help the gpsd port's tests. I get `deny(1) file-write-mode /dev/ttys092`. My PR only allows file-write-data, not file-write-mode. I don't know whether it is safe to additionally allow file-write-mode for these. I don't even know if it's safe to allow file-write-data, I only know that it solved the problem I was experiencing with Expect.

neverpanic:
`file-write-data` to a pty should be fine, that's more or less like typing in a terminal. No idea why gpsd requires changing permissions on ptys, though.

pmetzger:
Just FYI, the pty man page indicates that ptys and the corresponding ttys may be found at:

`/dev/[pt]ty[p-sP-S][a-z0-9]`

ryandesign:
> Just FYI, the pty man page indicates that ptys and the corresponding ttys may be found at:
> 
> `/dev/[pt]ty[p-sP-S][a-z0-9]`

I didn't read the manpage, I just looked at what actually happened, which was that Expect wanted to use a path starting with /dev/ttys and ending with 3 numbers. So that's what I implemented. This seemed to me to be a different group of files from the one you mentioned. I didn't know whether the two groups of files are used for different purposes, and since Expect didn't seem to require access to the group of files you mentioned, I didn't allow access to them.

PR #63, opened 2018-03-13
portindex: support creating PortIndex for libc++

mojca:
I'm submitting @ryandesign's patch from https://trac.macports.org/ticket/55471 for easier review and testing. 

It would be great to see it deployed soon. For that to happen another minor patch is needed, see the ticket.

PR #83, opened 2018-04-21
portindex: support creating PortIndex for libc++ (portindex only)

raimue:
I extracted the changes to `portindex` only from #63. These changes are safe to be merged in this form and do not require additional changes to our infrastructure generating the PortIndex (in mprsyncup).

This ensures we are using the default cxx_stdlib value when generating a PortIndex for older systems.

This is meant to be merged now so it can be released with MacPorts 2.5.0. The rest of the corresponding ticket, involving coordinated changes to mprsyncup and base, can be addressed afterwards.

See: https://trac.macports.org/ticket/55471

### DIFF
--- a/src/port/portindex.tcl
+++ b/src/port/portindex.tcl
@@ -23,7 +23,7 @@ mportinit ui_options global_options global_variations
 # Standard procedures
 proc print_usage args {
     global argv0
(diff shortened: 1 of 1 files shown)


raimue:
As it seems strange to discuss what is not in this pull request, and also to avoid spreading the discussion to even more places, I added my comments to the [ticket](https://trac.macports.org/ticket/55471#comment:26).

PR #84, opened 2018-04-27
mportsync: Ignore top-level directory in snapshots

raimue:
The name of the top-level directory in a snapshot tarball had to be the same as the root name of the tarball, which means the file name without extension.

However, tarballs generated by GitHub do not follow this rule, as it uses a top-level directory name like "repository-branch/". Therefore, this will ignore the top-level directory in the tarball and always use the root name as directory name for extraction. This is fully compatible with existing snapshot tarballs as the tarball root name and top-level directory name matched anyway.

This change allows to to sync the ports tree directly from GitHub as a tarball updated after every commit by using this line in sources.conf:

    https://github.com/macports/macports-ports/archive/master/ports.tar.gz [default]


raimue:
When this is merged, we should update [PortTreeTarball](https://trac.macports.org/wiki/howto/PortTreeTarball) on Trac documenting this with a note that it works as of MacPorts 2.5.0.

ryandesign:
Ok, but I wouldn't remove our existing tarball method, since it includes portindexes.

PR #85, opened 2018-05-03
Switch to docbook-xsl-nons

ryandesign:
I assume these are the correct changes for base that correspond to https://github.com/macports/macports-guide/pull/17.


ryandesign:
Ok, changed.

PR #86, opened 2018-05-04
standard_configure.sh: Do what MacPorts port does

ryandesign:
Sets the compiler, `-arch` flags, `CFLAGS`, `LDFLAGS`, and `MACOSX_DEPLOYMENT_TARGET`.

See: https://trac.macports.org/ticket/56055

### DIFF
--- a/standard_configure.sh
+++ b/standard_configure.sh
@@ -1,9 +1,141 @@
 #!/bin/sh
 
-# This is how we run configure when building binary packages (more or less,
-# minus architecture selection). If you don't want a custom build, this
-# is probably how you should run it too.
-env PATH=/usr/bin:/bin:/usr/sbin:/sbin CFLAGS="-pipe -Os" ./configure --enable-readline "$@"
+# This is how we run configure when building the MacPorts installer packages.
+# If you don't want a custom build, this is probably how you should run it too.
+#
+# If you want to use a different prefix, or any other additional configure
+# arguments, you can supply them as arguments when invoking this script.
+#
+# Some environment variables are also supported for altering the defaults:
+#
+# ARCHFLAGS
+#   -arch flags with which to build. If not specified, it is contructed from
+#   ARCHS. See also UNIVERSAL.
+#
+# ARCHS
+#   The list of architectures for which to build. If you prefer, specify the
+#   -arch flags in ARCHFLAGS instead. See also UNIVERSAL. To disable the use of
+#   -arch flags, use "ARCHS=".
+#
+# CC
+#   The C compiler executable.
+#
+# CFLAGS
+#   Additional C compiler flags which are appended to the defaults.
+#
+# LDFLAGS
+#   Additional linker flags which are appended to the defaults.
+#
+# MACOSX_DEPLOYMENT_TARGET
+#   The minimum macOS version on which the compiled libraries and programs are
+#   intended to run. It has not been tested whether building for an earlier
+#   deployment target results in all aspects of MacPorts functioning correctly
+#   on earlier macOS versions, so changing the value of this variable is not
+#   recommended except to perform such testing. In addition, some values that
+#   vary based on macOS version are recorded in text files like macports.conf,
+#   macports_autoconf.tcl and port_autoconf.tcl, so they must be generated on
+#   the same major version of macOS as the one on which they will be used.
+#
+# OPTFLAGS
+#   Compiler optimization flags.
+#
+# SDKPATH
+#   The path to the macOS SDK with which to build.
+#
+# UNIVERSAL
+#   Whether the default ARCHS are universal (yes/no).
 
-# If you want to use a different prefix, add this to the above:
-# --prefix=/some/path --with-applications-dir=/some/path/Applications
+set -eu
+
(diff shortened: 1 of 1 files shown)


standard_configure.sh
```diff
@@ -1,9 +1,141 @@
...
+fi
+
+if [ -z "${ARCHFLAGS-}" ]; then
+    for A in ${ARCHS-}; do
+        ARCHFLAGS="${ARCHFLAGS-} -arch $A"
+    done
+    ARCHFLAGS="${ARCHFLAGS# }"
+fi
+
+: "${CC=cc}"
```
raimue: Shouldn't we let the configure script figure this out? It is currently checking for clang, cc, gcc, in this order and would therefore prefer `clang` over `cc`.

ryandesign: We could let the configure script figure it out, at the expense of not allowing a user to override it by environment variable when calling standard_configure.sh. Or we could pass along a user-provided `CC` if set, and otherwise not pass along any value.

My script also defaults to `clang` on 10.6 and up, you'll notice. So that wouldn't be a difference.

My script uses `gcc-4.2` on 10.5, like the MacPorts port does (because that's the MacPorts base default compiler on 10.5). If we let the configure script use `cc`, that would be `gcc-4.0`, because that was Apple's default compiler on 10.5. The reason why MacPorts base defaults to `clang` on 10.6 was presumably to use the newest and best compiler available, so I wanted to use the newest and best compiler available on 10.5 too. Maybe that change belongs in base rather than in this script.

On the other hand, I'm using `gcc-4.0` on 10.4, unlike the MacPorts port which uses `gcc-apple-4.2` because that's the MacPorts base default compiler on 10.4. I didn't want to require that the apple-gcc42 port be installed.

Presumably, as soon as the user runs `sudo port selfupdate`, base gets rebuilt using whatever compiler the configure script finds, so any advantage of having the binary built with a "better" compiler is lost at that point.

Should selfupdate run standard_configure.sh?

ryandesign: It looks like selfupdate sets `CC=/usr/bin/cc OBJC=/usr/bin/cc`. Should it not do that, and let the configure script decide?

ryandesign: > It looks like selfupdate sets `CC=/usr/bin/cc OBJC=/usr/bin/cc`. Should it not do that, and let the configure script decide?

We're doing it because [8 years ago I said we should](https://trac.macports.org/ticket/23095). But that presumes we allow the user's `PATH` to be used. Do we? I hope not. If we changed selfupdate to use standard_configure.sh, that would solve that.

raimue: `PATH` will be set up by mportinit before running the selfupdate code. I assume `/usr/bin/cc` never caused problems because it already pointed to clang on macOS versions where we wanted to prefer clang (although it might have pointed to llvm-gcc-4.2 at some point?).

We should put the logic to prefer `gcc-4.2` on 10.5 and `gcc-4.0` on 10.4 into the configure script if we always want that. But that also seems to be the case for the other settings here. Therefore, instead of requiring `standard_configure.sh` to be run, we could move these defaults directly into `configure`. However, I would see that as a long-term goal and it is not a showstopper for this pull request.

ryandesign: On 10.6, `/usr/bin/cc` is gcc-4.2, not clang. Using gcc-4.2 doesn't cause any problems that I'm aware of, but it contradicts the intention of ce44cdcc04b59d0b4f08038b7096d616280110e1 which made the configure script prefer clang (though there was no explanation for why that change was made; we could revert it).

On 10.4, `/usr/bin/cc` is gcc-4.0, so there's no need to do anything special there to use that compiler.

standard_configure.sh was created in 70d6bdb6062b88f5f13c3ee754f957b18f7d13b8 so that we had a place to specify that for a "standard" build, we want to ignore any custom `PATH` and that we want to use `--enable-readline`, even though there were reasons why we did not want the configure script to do those things by default. Perhaps similar reasoning applies to some of the settings I'm proposing in this PR. Or perhaps not. I haven't thought about it a great deal.

My goal with this PR was to move functionality that is currently implemented in the MacPorts port into MacPorts base. It did not occur to me to attempt to modify the configure script. I assumed there were good reasons why we weren't doing these things in the configure script already.

ryandesign: An additional issue is that although our configure script prefers clang, it does not pass that preference on to the other configure scripts it runs, like tcl's. This means that on 10.6, if we don't set `CC` when running our configure, MacPorts base is built with clang but tcl and the other bundled packages are built with gcc. It seems better to build everything we ship with the same compiler and same flags. Setting then in standard_configure.sh accomplishes that. It may be possible to accomplish the same through modifications of the configure script but I have not looked into it.

raimue: I removed the unused `OBJC` with 8475592e and 9f8eb1d8.

See #89 for a way to pass CC to configure in vendor sources. If it makes things easier, you can also fold the changes into this pull request and close #89, but I did not want to just push them to this branch right away as the changes do not actually depend on each other.

ryandesign: #89 doesn't seem related to #86; I would keep them separate.

mojca:
Ping?

PR #87, opened 2018-05-05
Fix plugin install names

ryandesign:
This fixes the `install_name` of the darwintrace1.0, machista1.0 and macports1.0 plugins. There was already code in the pextlib1.0 Makefile to set the `install_name` which I copied to the other Makefiles.

The registry2.0 plugin uses an `install_name` containing `@loader_path` and this PR doesn't change that. This is needed to prevent a build failure when MacPorts is not already installed, because the pextlib1.0 plugin links with the registry2.0 plugin as a library. See 5f7e3d27bcd8643c820cacd9c6883980ff2298eb.

The `install_name` of the other plugins doesn't technically matter, but it is tidier to set it correctly, especially on Tiger where otherwise a randomly-generated temporary filename is used as the `install_name` (see https://trac.macports.org/ticket/37867).

I removed the space after the comma in the `ifeq` for consistency with our other `ifeq`s.

There was already a variable `INSTALLDIR` in each Makefile containing the path where the library will be installed, except that it was preceded by `DESTDIR` which we don't want in the `install_name`. I added a new variable `INSTALLDIR_WITHOUT_DESTDIR` which is now used to set the `install_name` and to construct `INSTALLDIR`. I'm open to suggestions for a shorter name for the `INSTALLDIR_WITHOUT_DESTDIR` variable.

It looks like all `INSTALLDIR` variables, in all Makefiles, begin with `DESTDIR`. So maybe an even better idea is to factor that out of the variable's definition, and just write `${DESTDIR}${INSTALLDIR}` everywhere that we currently write `${INSTALLDIR}`. That's what we recommend to other projects, so why don't we do it in our own code?

### DIFF
--- a/doc/Makefile.in
+++ b/doc/Makefile.in
@@ -106,7 +106,7 @@ MANXML=		${MANTXT:.txt=.xml}
 MANHTML=	${MANTXT:.txt=.html} ${MANSOELIM:.soelim=.html}
 
 CONF=		archive_sites.conf macports.conf pubkeys.conf sources.conf variants.conf
-INSTALLDIR=	${DESTDIR}${prefix}
+INSTALLDIR=	${prefix}
 TOPSRCDIR=	..
 
 ifneq ($(MAKECMDGOALS),distclean)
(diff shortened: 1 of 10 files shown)


raimue:
I agree that `${INSTALLDIR}` should be changed not to contain `${DESTDIR}`. We should use `${DESTDIR}${INSTALLDIR}` in the `install` targets.

ryandesign:
Ok, I'll make those changes before merging this PR.

PR #89, opened 2018-05-08
configure: Pass the selected $CC to vendor sources

raimue:
This came up in discussion on [pull request #86](https://github.com/macports/macports-base/pull/86#discussion_r186274970).

Should we also do the same for the other variables listed in `./configure --help`?

```
Some influential environment variables:
  CC          C compiler command
  CFLAGS      C compiler flags
  LDFLAGS     linker flags, e.g. -L<lib dir> if you have libraries in a
              nonstandard directory <lib dir>
  LIBS        libraries to pass to the linker, e.g. -l<library>
  CPPFLAGS    (Objective) C/C++ preprocessor flags, e.g. -I<include dir> if
              you have headers in a nonstandard directory <include dir>
  CPP         C preprocessor
```

### DIFF
--- a/aclocal.m4
+++ b/aclocal.m4
@@ -270,6 +270,9 @@ AC_DEFUN([MP_CONFIG_SUBDIR], [
 			mp_arg="--srcdir=$ac_srcdir"
 			_MP_LIST_APPEND_QUOTED([mp_sub_configure_args], [mp_arg])
 
+			mp_arg="CC=$CC"
+			_MP_LIST_APPEND_QUOTED([mp_sub_configure_args], [mp_arg])
+
 			AC_MSG_NOTICE([running $SHELL $ac_srcdir/configure $mp_sub_configure_args in $ac_dir])
 			eval "\$SHELL \$ac_srcdir/configure $mp_sub_configure_args" || AC_MSG_ERROR([configure failed for $ac_dir])
 		else
--- a/configure
+++ b/configure
@@ -7936,6 +7936,16 @@ $as_echo "$as_me: === configuring in $ac_dir ($mp_popdir/$ac_dir)" >&6;}
 	as_fn_append mp_sub_configure_args " $mp_arg"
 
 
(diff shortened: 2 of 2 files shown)


ryandesign:
> Pass the selected `$CC` to vendor sources

Any `CC` the user specified on the command line when invoking `./configure` were already passed to sub-configures, so we're only talking about the default value of `CC` that the configure script determines on its own. Yes, I think this should be passed on to sub-configures.

> Should we also do the same for the other variables listed in `./configure --help`?

Any user-supplied values were already being passed on. What values does our configure script determine on its own? If we move the `CFLAGS`- and `LDFLAGS`-setting code from standard_configure.sh in #86 into configure, then yes those should be passed on to sub-configures.

raimue:
The `AC_PROG_CC` macro in the configure script adds `CFLAGS="-g -O2"` unless the environment specifies something else. Tcl does [not follow](https://github.com/macports/macports-base/blob/master/vendor/tcl8.5.19/unix/configure.in#L55-L61) this autoconf default, but defines custom flags depending on the target platform (looks like `-g -Os` for Darwin).

I was mainly asking for consistency, as we could forget to add these variables later when we define them in the configure script.

ryandesign:
> The `AC_PROG_CC` macro in the configure script adds `CFLAGS="-g -O2"` unless the environment specifies something else.

Where does that come from? I didn't find the string `-g -O2` in our source, except in the generated configure script. standard_configure.sh and the MacPorts port currently use `-pipe -Os`. `-g` generates debug information? I don't know if we need that.

> I was mainly asking for consistency, as we could forget to add these variables later when we define them in the configure script.

Ok, sounds good.

raimue:
`AC_PROG_CC` is defined in `/opt/local/share/autoconf/autoconf/c.m4`.

mojca:
Ping?

neverpanic:
Let's merge this before it has its first birthday :)

PR #91, opened 2018-05-18
curl.c: User-Agent should default to "curl/version"

slewsys:
When MacPorts attempts to fetch from DropBox, it gets HTML instead of the requested file.  This is because MacPorts sets curl's User-Agent header to a non-default value. The problem was first [reported by Mojca](https://trac.macports.org/ticket/53942) in regard to port audio/generaluser-soundfont.

This PR simply hardcodes a default value for userAgent in curl.c.  Doing so allows audio/generaluser-soundfont to fetch without a custom fetch block.

### DIFF
--- a/src/port1.0/portfetch.tcl
+++ b/src/port1.0/portfetch.tcl
@@ -49,7 +49,7 @@ namespace eval portfetch {
 options master_sites patch_sites extract.suffix distfiles patchfiles use_tar \
     use_bzip2 use_lzma use_xz use_zip use_7z use_lzip use_dmg dist_subdir \
     fetch.type fetch.user fetch.password fetch.use_epsv fetch.ignore_sslcert \
-    master_sites.mirror_subdir patch_sites.mirror_subdir \
+    fetch.user_agent master_sites.mirror_subdir patch_sites.mirror_subdir \
     bzr.url bzr.revision \
     cvs.module cvs.root cvs.password cvs.date cvs.tag cvs.method \
     svn.url svn.revision svn.method \
@@ -115,6 +115,7 @@ default fetch.use_epsv "yes"
 default fetch.ignore_sslcert "no"
 # Use remote timestamps
 default fetch.remote_time "no"
+default fetch.user_agent ""
 
 default global_mirror_site "macports_distfiles"
 default mirror_sites.listfile {"mirror_sites.tcl"}
@@ -510,7 +511,7 @@ proc portfetch::hgfetch {args} {
 proc portfetch::fetchfiles {args} {
     global distpath all_dist_files UI_PREFIX \
            fetch.user fetch.password fetch.use_epsv fetch.ignore_sslcert fetch.remote_time \
-           portverbose
+           fetch.user_agent portverbose
     variable fetch_urls
     variable urlmap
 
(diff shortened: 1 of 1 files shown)


jmroot:
You seem to have included several unrelated commits.

slewsys:
Fixed.  Sorry.

jmroot:
I agree, misidentifying ourselves to every server to work around one broken server doesn't seem like a good approach. The curl command in pextlib has a --user-agent option, let's use that to fake it only when needed.

slewsys:
> Please include all details in the commit message and do not describe the problem in the pull request only. The commit message is what will be archived in the Git repository and will not even have a reference to this pull request outside of GitHub.

Okay,  fixed.  Sorry.

> We are setting the user agent in MacPorts on purpose. The libcurl default would be not to send a User-Agent at all. Why would we want to lie and hardcode "curl/version"?

On the other hand, MacPorts is basically doing the equivalent of:
```tcl
fetch {
    system -W ${distpath} "curl -LO ${master_sites}/${distname}${extract.suffix}"
}
```
which would set User-Agent as per this PR.  So I would argue that the burden is on MacPorts:  Why change the default?  It's perfectly reasonable for DropBox to respond differently to different user agents.  If MacPorts wants to behave like a web browser, then let it set User-Agent accordingly.

ryandesign:
> What if another server does not like `curl` in the user agent?

This has happened a zillion times already. Sometimes we've been able to convince server administrators to fix their configuration, sometimes not.

ryandesign:
I concur with rejecting this PR in its current state.

slewsys:
> I proposed to introduce fetch.useragent to allow ports to override it when needed. 

Okay.  I just added an underscore (_) between user and agent (i.e., `fetch.user_agent`) for the sake of consistency.

As noted in the commit message, while DropBox accepts User-Agent header `curl/...`, it does not like `libcurl/...`, so simply swapping the order of values in curl.c userAgent variable is not an option.

Furthermore, it appears that neither `proc macports::version` nor `${macports::autoconf::macports_version}` is available to Portfiles without `package require macports`, so the value `fetch.user_agent` can't be very descriptive.  The commit message example is:
```
fetch.user_agent    curl/MacPorts/4.xx
```

jmroot:
> Furthermore, it appears that neither proc macports::version nor ${macports::autoconf::macports_version} is available to Portfiles without package require macports

The proc is called `macports_version` in the portfile interpreter.

> the default value of fetch.user_agent is set to the empty string in portfetch.tcl

It doesn't need a default; you can check if it has been set with `[info exists fetch.user_agent]`.

slewsys:
> The proc is called macports_version in the portfile interpreter.
```
fetch.user_agent    curl/MacPorts/[macports_version]
```
Okay, that works for me.

> It doesn't need a default; you can check if it has been set with [info exists fetch.user_agent].

Yeah, I appreciate that `default` is overkill.  But all the other`fetch.*`. vars have defaults, so at least it's consistent.

ryandesign:
> Overall, the problem you are describing seems to be at DropBox, not in MacPorts. Was their support contacted?

Has DropBox been contacted about this problem? If so, what's their response?

mojca:
Thanks a lot for this (I didn't notice the PR earlier in the sea of other traffic). What would be the best approach for the `Portfile` in the meantime? Or rather: how could I at least convince the mirror to mirror this file, so that we could upgrade the port before this commits ends up in release?

mojca:
This was solved in a different way for now. I need to check if this variable could be used for livecheck as well.

PR #92, opened 2018-06-02
Make port-tclsh a relative symlink, not absolute

ryandesign:
This is a step toward allowing the MacPorts installer package to install to any prefix.

Using Perl is the most straightforward and most compatible way I found to compute a relative path from two absolute paths.

I wrote the Perl command inline because this is the only place where it's needed right now. But it could be broken out into a custom Makefile function to make it reusable.

Is it ok to just use `perl`, since it is so widely available, or do we need to do autoconf stuff to find it and set a variable with its path? We already use `sed` in the same Makefile without checking for it.

### DIFF
--- a/src/port/Makefile.in
+++ b/src/port/Makefile.in
@@ -39,7 +39,7 @@ install: all mkdirs
 	$(INSTALL) -d -o "${DSTUSR}" -g "${DSTGRP}" -m "${DSTMODE}" "${DESTDIR}${INSTALLDIR}/var/macports"
 	$(INSTALL)    -o "${DSTUSR}" -g "${DSTGRP}" -m 555 port portindex portmirror "${DESTDIR}${INSTALLDIR}/bin/"
 	$(LN_S) -f port "${DESTDIR}${INSTALLDIR}/bin/portf"
-	$(LN_S) -f "${TCLSH}" "${DESTDIR}${INSTALLDIR}/bin/port-tclsh"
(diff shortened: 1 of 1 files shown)


ryandesign:
That should work and would be simpler.

neverpanic:
Are you going to update the PR with the shell-based solution?

ryandesign:
Yes I should do that.

ryandesign:
> how about using the `${foo#bar}` prefix pattern expansion syntax to get the relative path?

The problem with this is that we are perpetuating the [mistake](https://trac.macports.org/ticket/56756) of making assumptions about the "bin" and "libexec" directory names/paths and their locations relative to one another. If we fix that mistake, then we need to compute the relative path from the one to the other, e.g. using the Perl solution I proposed.

ryandesign:
If avoiding Perl is strongly desired, the following should work:

```
bindir="${INSTALLDIR}/bin" && shopt -s extglob && rel="$${bindir//+([^\/])/..}" && $(LN_S) -f "$${rel:1}${TCLSH}" "${DESTDIR}${INSTALLDIR}/bin/port-tclsh"
```

The relative path this constructs is not minimal/optimal, but it should be correct regardless of how different bindir and libexecdir are from one another.

(Obviously this still incorrectly hardcodes the bin directory, but at least it makes it obvious where `${INSTALLDIR}/bin` could be replaced by the right thing.)

PR #93, opened 2018-06-02
Extend argument --no-mirrors to port distfiles

l2dy:
Extend 0cf1faf216c59419fb22edcd304ce02571d10504 to port-distfiles(1).

Maybe not the best way, but it works.

### DIFF
--- a/src/port1.0/portdistfiles.tcl
+++ b/src/port1.0/portdistfiles.tcl
@@ -61,6 +61,8 @@ proc portdistfiles::distfiles_main {args} {
         return 0
     }
 
+    uplevel #0 upvar #0 ports_distfiles_no-mirrors ports_fetch_no-mirrors
+
     # from portfetch... process the sites, files and patches
     set fetch_urls {}
     portfetch::checkfiles fetch_urls
--- a/src/port/port.tcl
+++ b/src/port/port.tcl
@@ -4416,6 +4416,7 @@ array set cmd_opts_array {
     diagnose    {quiet}
     reclaim     {enable-reminders disable-reminders}
     fetch       {no-mirrors}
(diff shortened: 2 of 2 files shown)


ryandesign:
I found it annoying to have to do `sudo port fetch --no-mirrors && sudo port checksum`; I wanted to be able to do `sudo port checksum --no-mirrors`.

I wonder if there is a better way to make an option available to more than one command without making it a global option.

src/port1.0/portdistfiles.tcl
```diff
@@ -61,6 +61,8 @@ proc portdistfiles::distfiles_main {args} {
         return 0
     }
 
+    uplevel #0 upvar #0 ports_distfiles_no-mirrors ports_fetch_no-mirrors
```
raimue: Yes, this works at the moment, but relies on the call stack, which seems error prone. I think we should just map this in the options array in `proc action_target` in the port client.

l2dy: Honestly, I can't understand why it should be in `proc action_target`. Perhaps I should've filed a ticket instead... Feel free to open another PR and close this one.

raimue: It depends on how to look at it. On one hand, the client sets the `ports_fetch_no-mirrors` option to be respected in all fetch operations. On the other hand, you could argue that the `distfiles` phase using the `fetch` phase is just an implementation detail inside the library that should be kept hidden.

I mainly do not like the idea to introduce aliases like that in the global namespace. The alternative would be to add a parameter to `portfetch::checkfiles` whether to use mirrors or not. Then both `fetch` and `distfiles` would just pass the value of the respective flag as specified by the client.

PR #95, opened 2018-06-08
cregistry: add SQLITE3 flags

kencu:
allows cregistry Makefile.in  to respond to
--with-sqlite3prefix
similarly to registry2.0 Makefile.in


kencu:
This fixes the build of MacPorts when using non-default locations for `sqlite3` by allowing the use of `--with-sqlite3prefix=/foo/bar`

PR #98, opened 2018-06-14
port1.0: Include non-default SDK in trace sandbox.

Lord-Kamina:
Apple has a nasty habit of nuking folders when something is updated, which means it isn't really safe to store older SDKs in the Developer folder.

So, what does this do?

Considering macports supports pointing to a non-default SDK via ```configure.sdkroot``` (which is very necessary if one is building binaries meant to be bundled and shared), I made sure running in trace-mode includes that folder, for the cases when it is not under the Developer folder.

### DIFF
--- a/src/port1.0/porttrace.tcl
+++ b/src/port1.0/porttrace.tcl
@@ -131,7 +131,7 @@ namespace eval porttrace {
     # @param workpath The $workpath of the current installation
     proc trace_start {workpath} {
         global \
-            developer_dir distpath env macportsuser os.platform \
+            developer_dir distpath env macportsuser os.platform configure.sdkroot \
             portpath prefix
 
         variable fifo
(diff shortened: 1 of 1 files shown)


src/port1.0/porttrace.tcl
```diff
@@ -207,6 +207,9 @@ namespace eval porttrace {
             }
         }
 
+		# Allow access to SDK if it's not inside the Developer folder.
```
raimue: Indent looks off. Please use spaces instead of tabs to be consistent with the rest of the file.

Lord-Kamina: Sorry I prefer tabs and forget to convert to spaces half the time before submitting 😅

Didn't set configure.sdk in the portfile but rather in macports.conf (and sometimes also via the command-line). This is sometimes necessary when compiling binaries meant to be compatible with older systems because else you might get weak-linked symbols that will not exist on older versions of macOS. And the reason I didn't put the SDK under the developer directory is that I'm under the impression that folder is nuked whenever Xcode is updated (and technically speaking, Apple does not even ship the 10.12 SDK anymore, imagine ones older than that)

neverpanic:
Besides the whitespace, this lg2m.

Lord-Kamina:
Just fixed the whitespace.

PR #99, opened 2018-07-04
Append history in interactive sessions

ylluminarious:
As per the discussion [here.](https://trac.macports.org/ticket/56381) Please let me know if you have any improvements in mind. Pinging @raimue since we spoke about this together on the aforementioned Trac ticket.

### DIFF
--- a/src/pextlib1.0/readline.c
+++ b/src/pextlib1.0/readline.c
@@ -267,6 +267,7 @@ int ReadlineCmd(ClientData clientData UNUSED, Tcl_Interp *interp, int objc, Tcl_
 		add line
 		read filename
 		write filename
+		append filename
 		stifle max
 		unstifle
 */
@@ -277,6 +278,7 @@ int RLHistoryCmd(ClientData clientData UNUSED, Tcl_Interp *interp, int objc, Tcl
 	char* s = NULL;
 	int i = 0;
 	Tcl_Obj *tcl_result;
+	FILE *hist_file;
 #endif
 
 	if (objc < 2) {
@@ -309,6 +311,16 @@ int RLHistoryCmd(ClientData clientData UNUSED, Tcl_Interp *interp, int objc, Tcl
 		}
 		s = Tcl_GetString(objv[2]);
 		write_history(s);
+	}  else if (0 == strcmp("append", action)) {
+		if (objc != 3) {
+			Tcl_WrongNumArgs(interp, 1, objv, "append filename");
+			return TCL_ERROR;
+		}
+		s = Tcl_GetString(objv[2]);
+
+		hist_file = fopen(s, "a");
+		fprintf(hist_file, "%s\n", current_history()->line);
+		fclose(hist_file);
 	} else if (0 == strcmp("stifle", action)) {
 		if (objc != 3) {
 			Tcl_WrongNumArgs(interp, 1, objv, "stifle maxlines");
--- a/src/pextlib1.0/readline.c
+++ b/src/pextlib1.0/readline.c
@@ -11,7 +11,7 @@
 #endif
 
 /* required for strdup(3) on Linux and macOS */
-#define _XOPEN_SOURCE 600L
+#define _BSD_SOURCE
 
 #include <stdio.h>
 #include <stdlib.h>
@@ -23,6 +23,10 @@
 
 #ifdef HAVE_READLINE_HISTORY_H
 #include <readline/history.h>
+#endif
+
+#if HAVE_SYS_FILE_H
+#include <sys/file.h>
 #endif
 
 #include <tcl.h>
@@ -317,9 +321,12 @@ int RLHistoryCmd(ClientData clientData UNUSED, Tcl_Interp *interp, int objc, Tcl
 			return TCL_ERROR;
 		}
 		s = Tcl_GetString(objv[2]);
-
 		hist_file = fopen(s, "a");
+
+		flock(fileno(hist_file), LOCK_EX);
 		fprintf(hist_file, "%s\n", current_history()->line);
+		flock(fileno(hist_file), LOCK_UN);
+
 		fclose(hist_file);
 	} else if (0 == strcmp("stifle", action)) {
 		if (objc != 3) {
--- a/src/pextlib1.0/readline.c
+++ b/src/pextlib1.0/readline.c
@@ -16,6 +16,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <errno.h>
 
 #ifdef HAVE_READLINE_READLINE_H
 #include <readline/readline.h>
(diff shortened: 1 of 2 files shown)


src/pextlib1.0/readline.c
```diff
@@ -309,6 +311,16 @@ int RLHistoryCmd(ClientData clientData UNUSED, Tcl_Interp *interp, int objc, Tcl
...
+	}  else if (0 == strcmp("append", action)) {
+		if (objc != 3) {
+			Tcl_WrongNumArgs(interp, 1, objv, "append filename");
+			return TCL_ERROR;
+		}
+		s = Tcl_GetString(objv[2]);
+
+		hist_file = fopen(s, "a");
+		fprintf(hist_file, "%s\n", current_history()->line);
+		fclose(hist_file);
```
raimue: There is a chance for a race condition here when multiple port instances are running at the same time. Although it is very unlikely that anyone would be able to type in both port instances at the same time, so we could just ignore it.

ylluminarious: @raimue Hmm, okay, I will see if I can address this with lockfiles or the like. If things get too hairy, I may just ignore it as you say.

ylluminarious:
@raimue 

> I am sorry I did not review your patch on Trac. My time for MacPorts is currently a bit scarce, but the pull request makes it easier. Thank you for that.

No problem, I figured you were probably too busy to respond on Trac. I definitely understand not having time for things like Macports sometimes; real life can be a pain! But you're welcome, and thank you for the kindness.

> The changes look good to me, although I did not test them. I would appreciate if you could use the usual trailer with "Closes: ..." as reference to the ticket.

Okay, I will try to do that in the future. Where am I supposed to put the "Closes: ..." remark? In the title of the PR?

jmroot:
> Okay, I will try to do that in the future. Where am I supposed to put the "Closes: ..." remark? In the title of the PR?

In the commit message body, on a line by itself.
https://trac.macports.org/wiki/CommitMessages

ylluminarious:
@jmroot Oh, okay. I had read that article previously, but I missed the part which gives an example of that. Thanks for pointing me to it!

pmetzger:
@ylluminarious You will want to squash your commits into a single commit...

ylluminarious:
@raimue I have added some file-locking code as discussed above. Sorry for the delay; I've been quite busy but finally managed to make some time for this. Please let me know if anything else should be done. Also, should I squash my commits into a single one to have the "Closes: ..." trailer?

pmetzger:
You must indeed always squash your commits. (I have no opinion on the PR itself...)

ylluminarious:
@pmetzger Ok, thanks for the info.

ylluminarious:
@raimue Friendly ping... any thoughts on this? :)

src/pextlib1.0/readline.c
```diff
@@ -309,6 +311,16 @@ int RLHistoryCmd(ClientData clientData UNUSED, Tcl_Interp *interp, int objc, Tcl
...
 		s = Tcl_GetString(objv[2]);
 		write_history(s);
+	}  else if (0 == strcmp("append", action)) {
+		if (objc != 3) {
+			Tcl_WrongNumArgs(interp, 1, objv, "append filename");
+			return TCL_ERROR;
+		}
+		s = Tcl_GetString(objv[2]);
+
+		hist_file = fopen(s, "a");
```
neverpanic: Can you please add error handling to the `fopen(3)` call here? `mkdtemp.c` has an example on how to do this, e.g.

```c
if (NULL == (hist_file = fopen(s, "a"))) {
    Tcl_AppendResult(interp, "fopen(", s, "): ", strerror(errno), NULL);
    return TCL_ERROR;
}
```

src/pextlib1.0/readline.c
```diff
@@ -309,6 +315,19 @@ int RLHistoryCmd(ClientData clientData UNUSED, Tcl_Interp *interp, int objc, Tcl
 		}
 		s = Tcl_GetString(objv[2]);
 		write_history(s);
+	}  else if (0 == strcmp("append", action)) {
+		if (objc != 3) {
+			Tcl_WrongNumArgs(interp, 1, objv, "append filename");
+			return TCL_ERROR;
+		}
+		s = Tcl_GetString(objv[2]);
+		hist_file = fopen(s, "a");
```
neverpanic: Thank you GitHub review for hiding my review comment against the version in one of the lower commits. Please add error handling here. See `mkdtemp.c` for good examples. E.g., use:

```tcl
if (NULL == (hist_file = fopen(s, "a"))) {
    Tcl_AppendResult(interp, "fopen(", s, "): ", strerror(errno), NULL);
    return TCL_ERROR;
}
```

src/pextlib1.0/readline.c
```diff
@@ -317,9 +321,12 @@ int RLHistoryCmd(ClientData clientData UNUSED, Tcl_Interp *interp, int objc, Tcl
 			return TCL_ERROR;
 		}
 		s = Tcl_GetString(objv[2]);
-
 		hist_file = fopen(s, "a");
+
+		flock(fileno(hist_file), LOCK_EX);
 		fprintf(hist_file, "%s\n", current_history()->line);
```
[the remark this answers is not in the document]
ylluminarious: @neverpanic Ok, I went ahead and took your advice for `fflush`. Strangely, what you're saying makes perfect sense, but when I tested this out, the locking mechanism actually worked. It was as though it didn't need a flush. I tested it by opening two separate terminals and entering commands into them via [CloneKeys.](https://code.google.com/archive/p/clonekeys/) I'm curious now; do you know why that may have happened?

src/pextlib1.0/readline.c
ylluminarious: @neverpanic Thanks for the code, I went ahead and pasted it in verbatim :)

I also added some error-checking for the `fprintf` call; please let me know if it's okay. It's not as pretty as the code in `mktemp.c`, but it should do the trick.

ylluminarious:
@neverpanic I went ahead and implemented your changes as you requested. Please let me know if they're satisfactory. I also added some comments for visual and conceptual clarity since I stared at this code without comments and it felt like my eyes were getting poked with curly-braced needles :)

neverpanic:
Thanks!

ylluminarious:
@neverpanic You're welcome! Thanks for merging it and for the help!

Did you see [my comment](https://github.com/macports/macports-base/pull/99#discussion_r211449539) from last night wherein I had a question about the flushing seeming unnecessary? I'm curious to know what may have been going on... :)

ylluminarious:
@neverpanic Thanks so much for the explanation! This makes more sense to me now. I suspected that `fclose` was behind this. I guess the lock gave just enough of a delay that the flushing done by `fclose` happens right before the acquisition of the lock for the competing process, or at least before *that* competing process was able to do *its* flush. Thus, there was a *de facto* order of operations, although unreliable.

Yes, such a situation would indeed make it hard to expose the problem. That is a very hard-to-hit race condition, although possible. And as you also said, if the two writes were mixed into the same syscall then there would be no conflict anyway. Maybe I'll try that test you mention someday, just for fun :D

PR #100, opened 2018-08-20
Add option to pass custom Header to curl fetch

cjones051073:
See

https://trac.macports.org/ticket/56975

src/pextlib1.0/curl.c : Add --append-http-header option to allow custom HTTP headers to be appended to fetches.

src/port1.0/portlivecheck.tcl : Add new livecheck.curloptions options to allow additonal options to be passed to the curl fetch. Default is :-

```
{{ "--append-http-header" "Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8" }}
```

### DIFF
--- a/src/pextlib1.0/curl.c
+++ b/src/pextlib1.0/curl.c
@@ -178,6 +178,9 @@ CurlFetchCmd(Tcl_Interp* interp, int objc, Tcl_Obj* CONST objv[])
 		};
 		char* effectiveURL = NULL;
 		char* userAgent = PACKAGE_NAME "/" PACKAGE_VERSION " libcurl/" LIBCURL_VERSION;
+		const int MAXHTTPHEADERS = 100;
+		int numHTTPHeaders = 0;
+		const char* httpHeaders[MAXHTTPHEADERS];
 		int optioncrsr;
 		int lastoption;
 		const char* theURL;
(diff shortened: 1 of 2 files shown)


neverpanic:
LG2M, although I'd prefer to keep the first line of the commit message under 70 chars, e.g. using `pextlib: Support custom HTTP headers`.

Can you also comment on what changing the default on the `Accept` header achieves for livecheck? Have there been problems with livecheck not getting useful answers due to a missing `Accept` header?

cjones051073:
Ok, i will shorten the commit message tomorrow.

The linked trac ticket gives reasons for the change. Yes, github eventually reverted whatever change they made, but the argue is its better to pass something more akin to what browsers use.

jmroot:
What if you need to add more than one header?

cjones051073:
Then just add whatever you want to the second string in the above example.

jmroot:
So you'd need to separate headers with CRLF? That seems a little iffy given the warnings in the CURLOPT_HTTPHEADER man page.

cjones051073:
Ok, reading 

https://curl.haxx.se/libcurl/c/CURLOPT_HTTPHEADER.html

Again i think i see your point. Each header should be individually added using curl_slist_append, and h patch here thus only supports adding a single additional header.

cjones051073:
I've added support for using the `--append-http-header` multiple times (up to 100...).

cjones051073:
So, whats the deal with PRs to base. Who gets to eventually merge them ?

pmetzger:
There are no rules per se about who gets to merge them. I suppose the question is whether @neverpanic and @jmroot are happy. If they are, this can be merged.

PR #101, opened 2018-08-20
port lint: warn if checksum section is missing file size field

herbygillot:
As per discussion in macports/macports-guide#21

### DIFF
--- a/src/port1.0/portlint.tcl
+++ b/src/port1.0/portlint.tcl
@@ -556,6 +556,13 @@ proc portlint::lint_main {args} {
         }
     }
 
+    if {[info exists checksums]} {
+        if {![regexp {size\s+(\d+)} $checksums]} {
+            ui_warn "Checksum(s) should include the 'size' field"
+            incr warnings
+        }
+    }
+
     if {[info exists conflicts]} {
         foreach cport $conflicts {
             if {[regexp {[^[:alnum:]_.-]} $cport]} {


pmetzger:
LGTM, but I'd like to hear what people more skilled in tcl think of how you've done it.

pmetzger:
BTW, there's a few other things that would be good to get warnings about. If you feel like adding them:

1. Using old checksums (example, md5) instead of the more modern set.
2. Failing to include a github handle if the maintainer is not `nomaintainer`
3. Failing to include a valid `license` type.

pmetzger:
See also: https://trac.macports.org/ticket/37667

There may be other trac tickets open for various sorts of lint additions.

And if you don't have time, don't worry about it, we'll add just this one.

jmroot:
I would not warn about this at all given that it currently does nothing, but if you must add this check, please restrict it to nitpick mode.

herbygillot:
@pmetzger I may try to look at the issues you've highlighted in separate PRs, though I do have questions.

> 1. Using old checksums (example, md5) instead of the more modern set.
Are the only blessed checksum algorithms `sha256` and `rmd160`? Where can I see the list of deprecated/unwanted algorithms?

> 3. Failing to include a valid `license` type.
The only programmatic source of licenses seems to be in a totally different codebase, the `portmgr` codebase.  I'd love to know what programmatic source of truth I can use that lives inside inside of `macports-base`, so far I haven't been able to locate it.

As for this specific PR, I will wait until a final decision gets made on nitpick mode vs. standard warning (though I personally feel it should be a warning since all new port submissions are being told that the field should be included).

herbygillot:
I misspoke, the license source of truth is in a jobs directory in `base/portmgr/jobs` which only seems present in Trac, and not here in Github....?

jmroot:
It moved to the macports-infrastructure repo when we split things up during the move to GitHub. Also it doesn't technically list all possible licenses, but only those that are known to be binary-distributable.

pmetzger:
Do not restrict this to nitpick mode. I need people to see this when they request pulls. @jmroot has been around a long time, far longer than me, but he doesn't process the pull queue and I do. The most common issue I have is asking people to add "size". I very much appreciate adding this.

@jmroot If you want to argue about it, get consensus on the developers list. This is not the place for the discussion.

pmetzger:
> Are the only blessed checksum algorithms sha256 and rmd160?

Currently, that seems to be the case. I would probably rig things so that the list of blessed checksums is in a variable that can be updated if the recommended ones change in the future.

ryandesign:
I agree with not hiding this check behind the `--nitpick` flag.

Rather than having a list of good algorithms, I suggest having a list of bad algorithms, which would currently be md5 and sha1.

We might not even need a list of algorithms. I would say we should simply issue a warning if a port uses only one checksum type (not counting size). If a port uses two checksums for a file, even if they are deprecated algorithms, that's fine. Someone might be able to craft two files that have the same md5 or sha1 sum, but as far as I know it's still impossible to craft two files that have the same sum in more than one algorithm, since each algorithm works differently.

The proposed fix doesn't really parse the checksums correctly. For example, it doesn't consider that a port might have multiple distfiles (or non-local patchfiles), in which case, there will be a set of checksums for each file. The proposed fix will be satisfied if any of the files has a size listed, but we want to verify that each file has a size listed. We probably need to look at the implementation of how checksums are parsed in portchecksum.tcl and redo some of that here in portlint.tcl.

pmetzger:
I'm merging this for now, as it seems like a reasonable minimum. We probably should add more, as has been discussed. More PRs for additional checks actively solicited. Thanks for this, @herbygillot!

herbygillot:
Happy to help - will take a look at the other checks.

PR #103, opened 2018-08-31
port-info: output fields in the order given on the command line

kurthindenburg:
Currently due to the use of 'array names', the output of
'port info --multiple-fields' are displayed in a random order.
Now, the fields are kept in a list which preserves their order they
were given on the command line.
This has the effect of also allowing duplicate fields to be displayed.

Closes: https://trac.macports.org/ticket/57003

### DIFF
--- a/src/port/port.tcl
+++ b/src/port/port.tcl
@@ -1913,6 +1913,8 @@ proc action_log { action portlist opts } {
 
 proc action_info { action portlist opts } {
     global global_variations
+    global ui_options
+
     set status 0
     if {[require_portlist portlist]} {
         return 1
@@ -2094,7 +2096,7 @@ proc action_info { action portlist opts } {
 
         # Spin through action options, emitting information for any found
         set fields {}
-        set opts_todo [array names options ports_info_*]
+        set opts_todo $ui_options(options_info_order)
         set fields_tried {}
         if {![llength $opts_todo]} {
             set opts_todo {ports_info_heading
@@ -4468,6 +4470,8 @@ proc parse_options { action ui_options_name global_options_name } {
     upvar $ui_options_name ui_options
     upvar $global_options_name global_options
 
+    set options_info_order {}
+
     while {[moreargs]} {
         set arg [lookahead]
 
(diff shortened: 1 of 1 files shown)


src/port/port.tcl
```diff
@@ -4495,6 +4499,9 @@ proc parse_options { action ui_options_name global_options_name } {
                     set kargc [lindex $kopts 0 1]
                     if {$kargc == 0} {
                         set global_options(ports_${action}_${key}) yes
+                        if {${action} eq "info"} {
+                            lappend options_info_order ports_${action}_${key}
+                        }
```
raimue: I am not particularly fond of special cases for a certain action in the parser. I would suggest to always store `options_${action}_order` to keep this generic. Maybe the order will also be useful for other actions later.

src/port/port.tcl
```diff
@@ -4591,6 +4598,7 @@ proc parse_options { action ui_options_name global_options_name } {
 
         advance
     }
+    set ui_options(options_info_order) $options_info_order
```
raimue: Not really sure this fits into `ui_options`, as everything else related to an action is in `global_options`. Or am I missing something?

kurthindenburg:
I somehow missed the email w/ your comments.  Let me research this again.

pmetzger:
@kurthindenburg I tend to just look at Github notices for myself on the web site. It's very reliable.

kurthindenburg:
The original code also had a number of bugs including mishandling --pretty, --line, --index.  I will update shortly.

kurthindenburg:
ping? before this gets stale...

pmetzger:
@kurthindenburg What would you like done at this point?

kurthindenburg:
A review on the code and if that's the best way to handle this.

pmetzger:
@raimue are you in a position to do another code review?

src/port/port.tcl
```diff
@@ -2094,7 +2095,20 @@ proc action_info { action portlist opts } {
...
         # Spin through action options, emitting information for any found
         set fields {}
-        set opts_todo [array names options ports_info_*]
+
+        # This contains the display fields in random order
+        set opts_info [array names options ports_info_*]
+        # This contains all parameters in order given on command line
+        set opts_action $global_options(options_${action}_order)
+        # Get the display fields in order provided on command line
+        #  ::struct::set intersect does not keep order of items
```
raimue: Maybe I am missing something in this comment, but are we actually using `struct::set` anywhere...?

kurthindenburg: It is not used, I added the comment in case someone looks at this and thinks ::set intersect could be used (which is what I tried first).

pmetzger:
@kurthindenburg, I think you can merge this.

PR #104, opened 2018-08-31
port: add -W to touch command

kurthindenburg:
Instead of having to use system(), use touch -W dir file

See https://trac.macports.org/ticket/53730


raimue:
Not completely opposed to this, but what exactly do we gain over writing `touch dir/file`?

ryandesign:
The ability to touch multiple files in a folder without having to repeat the folder name for each one.

raimue:
Okay, I did not think of that. In the case of multiple files, that is indeed useful.

PR #105, opened 2018-09-06
port lint: lint checksums much more thoroughly

herbygillot:
This PR builds upon discussion originating in macports/macports-guide#21, and work started in #101.

The changes here make `port lint` perform a much more thorough vetting of checksum stanzas in Portfiles.

Any checksum type that isn't part of the canonical list specified in `portchecksum` is counted as an error, as well as any unrecognized field or token.  (This will also catch situations where a filename is hanging without any checksums specified for it, erroring out.) . Errors also happen when a particular checksum does not resemble its specified type (ie., md5 1234).

In the case of deprecated checksum types, warnings are issued.  Warnings are also issued in the case that a set of checksums does not include the current set of recommended checksum types (sha256, rmd160 & size, also as per `portchecksum`).

As part of this work, it was realized that `portlint` had no tests.  This adds a test suite for `portlint`, but is currently only testing the checksum linting logic.  At the moment, not very much else in `portlint` seems to be written in a testable way, but I am not a TCL expert, so perhaps I'm wrong.

As just mentioned, I am not an experienced TCL developer, so this was very likely not done in the most succinct/cleanest way.  Thanks in advance.

### DIFF
--- a/src/port1.0/portchecksum.tcl
+++ b/src/port1.0/portchecksum.tcl
@@ -40,6 +40,13 @@ target_requires ${org.macports.checksum} main fetch
 target_prerun ${org.macports.checksum} portchecksum::checksum_start
 
 namespace eval portchecksum {
+
+    # The list of the types of checksums we know.
+    variable checksum_types [list md5 sha1 rmd160 sha256 size]
+
+    # types to recommend if none are specified in the portfile
+    variable default_checksum_types [list rmd160 sha256 size]
+
 }
 
 # Options
@@ -51,14 +58,43 @@ default checksum.skip false
 
 set_ui_prefix
 
-# The list of the types of checksums we know.
-set checksum_types [list md5 sha1 rmd160 sha256 size]
+# verify_checksum_format
+#
+# Given a checksum type as string and the actual checksum:
+#
+# - return 1  if the value has the expected format
+# - return 0  if the value does not look as expected
+# - return -1 if the checksum type is unrecognized
+proc portchecksum::verify_checksum_format {type value} {
+    set result 0
+
+    switch [string tolower $type] {
+        sha256 {
+          set result [regexp {^\w{64}$} $value]
+        }
+        rmd160 {
+          set result [regexp {^\w{40}$} $value]
+        }
+        sha1 {
+          set result [regexp {^\w{40}$} $value]
+        }
+        md5 {
+          set result [regexp {^\w{32}$} $value]
+        }
+        size {
+          set result [regexp {^\d+$} $value]
+        }
+        default {
+          # unrecognized checksum type
+          set result -1
+        }
+    }
+
+    return $result
+}
 
 # The number of types we know.
-set checksum_types_count [llength $checksum_types]
-
-# types to recommend if none are specified in the portfile
-set default_checksum_types [list rmd160 sha256 size]
+set checksum_types_count [llength $portchecksum::checksum_types]
 
 # Using global all_dist_files, parse the checksums and store them into the
 # global array checksums_array.
@@ -70,12 +106,12 @@ set default_checksum_types [list rmd160 sha256 size]
 # Portfile is in format #1 if:
 # (1) There is only one distfile.
 # (2) There are an even number of words in checksums (i.e. "md5 cksum sha1 cksum" = 4 words).
-# (3) There are no more than $checksum_types_count checksums specified.
+# (3) There are no more than $portchecksum::checksum_types_count checksums specified.
 # (4) first word is one of the checksums types.
 #
 # return yes if the syntax was correct, no if there was a problem.
 proc portchecksum::parse_checksums {checksums_str} {
-    global checksums_array all_dist_files checksum_types checksum_types_count
+    global checksums_array all_dist_files checksum_types_count
 
     # Parse the string of checksums.
     set nb_checksum [llength $checksums_str]
@@ -83,7 +119,7 @@ proc portchecksum::parse_checksums {checksums_str} {
     if {[llength $all_dist_files] == 1
         && [expr {$nb_checksum % 2}] == 0
         && [expr {$nb_checksum / 2}] <= $checksum_types_count
-        && [lindex $checksums_str 0] in $checksum_types} {
+        && [lindex $checksums_str 0] in $portchecksum::checksum_types} {
         # Convert to format #2
         set checksums_str [linsert $checksums_str 0 [lindex $all_dist_files 0]]
         # We increased the size.
@@ -112,7 +148,7 @@ proc portchecksum::parse_checksums {checksums_str} {
             incr ix_checksum
             while {1} {
                 set checksum_type [lindex $checksums_str $ix_checksum]
-                if {$checksum_type in $checksum_types} {
+                if {$checksum_type in $portchecksum::checksum_types} {
                     # append the type and the value.
                     incr ix_checksum
                     set checksum_value [lindex $checksums_str $ix_checksum]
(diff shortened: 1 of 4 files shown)


src/port1.0/portlint.tcl
```diff
@@ -115,6 +116,132 @@ proc portlint::seems_utf8 {str} {
...
+
+    for {set ctr $ctr_start} \
+        {[expr $ctr < [llength $checksum_tokens]] && !$is_error} \
+        {} {
+
+        set current [lindex $checksum_tokens $ctr]
+
+        if {$current in [portchecksum::list_checksum_types]} {
+            set c_type  $current
+            set c_value [lindex $checksum_tokens [expr $ctr + 1]]
```
jmroot: Input to expr should be passed as a single argument, wrapped in braces. http://wiki.tcl.tk/10225

Also, lindex can do simple arithmetic directly, so this expr call isn't needed.

src/port1.0/portlint.tcl
```diff
@@ -115,6 +116,132 @@ proc portlint::seems_utf8 {str} {
...
+    set filename ""
+    set pfx ""
+    set has_filenames false
+
+    set types [list]
+
+    # List of all tokens in the checksum string
+    set checksum_tokens [regexp -all -inline {\S+} $checksum_string]
+
+    if {[string equal [lindex $checksum_tokens 0] "checksum"]} {
```
jmroot: You can use the `eq` operator instead of calling `string equal`.

src/port1.0/portlint.tcl
```diff
@@ -115,6 +116,132 @@ proc portlint::seems_utf8 {str} {
...
+        set current [lindex $checksum_tokens $ctr]
+
+        if {$current in [portchecksum::list_checksum_types]} {
+            set c_type  $current
+            set c_value [lindex $checksum_tokens [expr $ctr + 1]]
+
+            switch [portchecksum::verify_checksum_format $c_type $c_value] {
+                1 {
+                    # checksum type recognized, and checksum looks good
+                    set ctr [expr $ctr + 2]
```
jmroot: You can use `incr` here.

src/port1.0/portlint.tcl
```diff
@@ -115,6 +116,132 @@ proc portlint::seems_utf8 {str} {
...
+                    lappend types $c_type
+                }
+
+                0 {
+                    # checksum type recognized, but checksum looks bad
+                    lappend errors [concat "${pfx}checksum type" \
```
jmroot: Not sure why you're using concat here.

src/port1.0/portlint.tcl
```diff
@@ -115,6 +116,132 @@ proc portlint::seems_utf8 {str} {
...
+                -1 {
+                    # checksum type not recognized
+                    lappend errors [concat "${pfx}invalid checksum" \
+                                           "type: $c_type $c_value"]
+                    set is_error true
+                    continue
+                }
+            }
+
+        } elseif {[expr $ctr > $ctr_start] && !$has_filenames} {
```
jmroot: No need to use expr here or for inequalities in `if` statements in general.

src/port1.0/portlint.tcl
```diff
@@ -557,9 +684,18 @@ proc portlint::lint_main {args} {
...
-            ui_warn "Checksum(s) should include the 'size' field"
-            incr warnings
+        set checksum_lint [portlint::lint_checksum $checksums]
+
+        foreach err [lindex $checksum_lint 0] {
+            ui_error $err
+            incr errors
+        }
+
+        if {[llength [lindex $checksum_lint 0]] == 0} {
```
jmroot: Warnings should be shown regardless of whether there were errors.

pmetzger:
Thanks so much for working on this! @jmroot has pointed out tcl issues above.

src/port1.0/portchecksum.tcl
```diff
@@ -51,14 +51,63 @@ default checksum.skip false
...
+proc portchecksum::list_checksum_types {} {
+    return [list md5 sha1 rmd160 sha256 size]
+}
+
+# list_preferred_checksum_types
+#
+# Return a list of the preferred checksum types recommended for ports.
+proc portchecksum::list_preferred_checksum_types {} {
+    return [list rmd160 sha256 size]
+}
```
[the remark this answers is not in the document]
herbygillot: I figured procs were more robust for exporting, and more versatile in case anyone was going to change just _how_ the list of these algorithms were being determined in `portchecksum` in the future.  Will fix.

src/port1.0/portlint.tcl
```diff
@@ -115,6 +116,132 @@ proc portlint::seems_utf8 {str} {
...
+# Given a list of checksum types, return a list of strings which are warnings
+# about deprecated checksum types, or missing recommended types.
+#
+# Returns an empty list if no issues are found.
+proc portlint::lint_checksum_type_list {types} {
+    set issues [list]
+
+    foreach type $types {
+        if {$type ni [portchecksum::list_preferred_checksum_types]} {
+            lappend issues "checksum type is deprecated: $type"
```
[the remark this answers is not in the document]
pmetzger: I think it is. SHA1 is no longer secure.

src/port1.0/portlint.tcl
herbygillot: Was trying to keep things below a certain column length.  Will fix.

src/port1.0/portlint.tcl
herbygillot: One of the reasons is because we stop processing immediately in an error situation, especially in the case of invalid tokens that don't make any sense for checksum.  In that situation, sometimes the warnings are inaccurate.

src/port1.0/portlint.tcl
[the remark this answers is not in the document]
herbygillot: Agreed.  I had the impression we wanted to warn when any deprecated insecure algorithm was used in any case.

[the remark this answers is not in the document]
jmroot: No, we want to warn when secure algorithms are _not_ being used. Adding something else to the mix on top of the recommended ones _increases_ certainty that the file is correct. It's literally the opposite of a problem.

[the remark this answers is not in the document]
herbygillot: Makes sense.  Will fix.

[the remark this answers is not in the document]
herbygillot: OK, now deprecation warnings only occur if no recommended checksum types have been seen.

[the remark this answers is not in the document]
pmetzger: > No, we want to warn when secure algorithms are not being used. Adding something else to the mix on top of the recommended ones increases certainty that the file is correct. It's literally the opposite of a problem

I think it's best not to encourage use of obsolete algorithms for human factors reasons because (as I've noted) SHA1 is not secure. Yes, in the context of using other secure hashes it can't cause much damage, but we don't want to encourage people to continue using the algorithm at all.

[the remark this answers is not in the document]
pmetzger: It's not a valid use because you can't trust the upstream md5 or sha1 anyway so who cares if they advertise one. An md5 is especially easy to forge at this point, and a sha1 is getting there.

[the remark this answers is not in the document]
pmetzger: (And if someone is still advertising the md5 of their package, that probably means they haven't been paying attention for years and years and the thing probably isn't maintained, which is another issue.)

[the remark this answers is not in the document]
danielluke: I'd say it's still useful to compare with upstream checksums. We don't have a lot of control over what upstream chooses to advertise (and believe it or not, some projects are actively worked on  without being run by people with understanding of which hashes are still good or not).

src/port1.0/portlint.tcl
jmroot: OK, so don't show the ones that might be spurious then. Are there others that are still valid in that case though?

src/port1.0/portlint.tcl
jmroot: You can use a backslash to continue the line without concat. A backslash followed by a newline and any amount of whitespace is parsed as a single space.

src/port1.0/portlint.tcl
herbygillot: OK, this has been resolved.

pmetzger:
Merged! Thanks, @herbygillot!

herbygillot:
Thank you @pmetzger and @jmroot.

PR #106, opened 2018-09-10
Portfile.7 fixes

lemzwerg:


mojca:
This is not the question for PR submitter, but I'm confused. Aren't our man pages now in asciidoc format? I don't seem to find the asciidoc source for `portfile.7` though.

lemzwerg:
Yeah, my work is probably for nothing :-)  But fixing the man page doesn't hurt, and it's my way of learning stuff...

doc/portfile.7
```diff
@@ -31,39 +31,49 @@
 .Dd February 13, 2007
 .Dt PORTFILE 7 "MacPorts"
 .Os Darwin
+.
+.
```
pmetzger: What is the purpose of these periods?

doc/portfile.7
```diff
@@ -31,39 +31,49 @@
...
 .Sh NAME
 .Nm Portfile
 .Nd MacPorts description file reference
+.
+.
```
pmetzger: What is the purpose of these periods?

doc/portfile.7
```diff
@@ -31,39 +31,49 @@
...
 .Sh DESCRIPTION
 A complete reference of all available
 .Nm
 variables and example syntax.
 .Nm Portfiles
-consist of valid TCL, allowing for simple key/value pair syntax as well
-as utilization of TCL's extensive functionality.
+consist of valid TCL, allowing for simple key/value pair syntax as well as
+utilization of TCL's extensive functionality.
```
pmetzger: What is the purpose of moving this from one line to another?

doc/portfile.7
```diff
@@ -31,39 +31,49 @@
...
 .Pp
 The
-\.Nm MacPorts System
-uses a target dependency system based on a
-depends/provides model, allowing for targets to be registered and
-executed in the correct order based on their individual requirements.
+.Nm MacPorts System
+uses a target dependency system based on a depends/provides model, allowing
+for targets to be registered and executed in the correct order based on their
+individual requirements.
```
pmetzger: Removing the `\` is probably correct, but what is the purpose of moving the rest from one line to another?

doc/portfile.7
```diff
@@ -31,39 +31,49 @@
...
 .Pp
 A
 .Nm
-author needs to be aware of the various standard targets, the options
-that they require and the variables that both the targets and the port
-system provide.
+author needs to be aware of the various standard targets, the options that
+they require and the variables that both the targets and the port system
+provide.
```
pmetzger: What is the purpose of moving text here from one line to another?

doc/portfile.7
```diff
@@ -31,39 +31,49 @@
...
+.
+.
```
pmetzger: What is the purpose of these periods?

doc/portfile.7
```diff
@@ -31,39 +31,49 @@
...
 .Sh PORTSYSTEM
 Portfiles must begin with a PortSystem line that defines which version of the
 Portfile interpreter should be used.
 .br
 .Sy Synopsis:
 .Dl PortSystem 1.0
+.
+.
```
pmetzger: What is the purpose of these periods?

doc/portfile.7
```diff
@@ -31,39 +31,49 @@
...
 .Sh MAIN VARIABLES
 All ports are required to set certain variables.
+.
```
pmetzger: What is the purpose of this period?

doc/portfile.7
raimue: A single dot is an empty request for the roff processor and is ignored. As the commit message says this is used throughout the file to visualize the structure in the roff source.

doc/portfile.7
```diff
@@ -2696,7 +2975,7 @@ Display a
 to the user, at various different levels.
 .br
 .Sy Example:
-.Dl ui_msg \*qAdd each user to the system using the clamav command\*q
+.Dl ui_msg \*qThe MySQL client has been installed.\*q
```
raimue: This is not really what `ui_*` should be used for, but the previous example was not that good either (because we have `notes` for that now). In fact, I cannot even think of a valid use case for `ui_msg` except for debugging purposes...

doc/portfile.7
pmetzger: It's very unconventional. I've been writing `mdoc` files for a very long time and it's almost never done.

lemzwerg: Is it?  I was more than 15 years the maintainer of groff, and I've used such structuring all the time.  Actually, it was me together with Ruslan Ermilov who have completely rewritten the mdoc macros to be consistent and readable, fixing many bugs...

doc/portfile.7
lemzwerg: Simple formatting to have a right margin of 78 characters, as is common for text documents.  There were many overlong lines that I've also formatted in due course.

doc/portfile.7
lemzwerg: As I've mentioned in the commit message, I didn't change the contents of the document – to do that, my knowledge is simply too limited.

doc/portfile.7
lemzwerg: BTW, I suggest to have a look into the source code of the [`groff_mdoc.7` man page](http://git.savannah.gnu.org/cgit/groff.git/tree/tmac/groff_mdoc.7.man) to see how such files should look like IMHO.

lemzwerg: Ah, and please show me the incorrect changes of the *formatting* (not contents).  Maybe I've missed something, or there are typos.

doc/portfile.7
lemzwerg: I don't have time to do such a conversion, sorry.  I suggest to use mandoc's ability to convert [mdoc to markdown](https://undeadly.org/cgi?action=article&sid=20170304230520) and then [pandoc](http://pandoc.org) to convert markdown to AsciiDoc – the more precise the mdoc sources are tagged, the better the conversion, BTW.  So maybe my work wasn't completely useless...

Note that one of the authors of mandoc, Ingo Schwarze, is very helpful; he can certainly assist in case problems arise during the conversion.  Maybe you can even persuade him to make mandoc directly emit AsciiDoc, assuming that you give him a helping hand.

pmetzger:
I'm much less sanguine about these changes. Moving words between lines in the document isn't very useful.

doc/portfile.7
lemzwerg: It's *very* irritating that you obviously didn't have a closer look at the three commits.  The first one only introduces structure; the second one does the mdoc syntax fixes and improvements, and the third one does the 'useless' formatting, i.e., re-formatting to get decent line lengths, together with using two spaces after a full dot.

Guess why I did this split-up into logical units – it wouldn't cross your mind that, say, the third commit could be omitted, right?

For me, adding the empty dots is almost *everything* – suddenly I can quickly find the next `.Sh` without using a forward search...  IMHO it's a weakness of the NetBSD manual linter that it doesn't enforce at least *a bit* structure to otherwise very hard to read man pages.  Similarly, it's a weakness that it doesn't enforce two spaces after a full dot.

Whatever.  Do what you want with my changes.  I no longer care.

pmetzger: > It's very irritating that you obviously didn't have a closer look at the three commits

I went to "files changed", as I always do, and started reading. After a page or two of things that did nothing I stopped. If you want the changes considered separately, then you should submit distinct pull requests.

> For me, adding the empty dots is almost everything – suddenly I can quickly find the next .Sh without using a forward search

I've never had such a problem.

doc/portfile.7
raimue: Sorry to reopen this, but I pointed at this hunk particularly because the content did in fact change. The question was whether this happened intentionally, as I did not expect it in this commit.

Your changes definitely are useful and I am convinced we should merge this. I just wanted to hint at the need for a  conversion to AsciiDoc to match the other man pages. However, I fully understand if this is out of your scope. For the other man pages (such as `port.1` to `port.1.txt`) I used a manual approach and rewrote them in AsciiDoc from the rendered man output, as no automatic tool produced satisfactory results (but note this was like 10 years ago).

pmetzger: Since mandoc is semantic markup, scripting the conversion might not be too bad.

lemzwerg: @raimue I wanted to shorten the text to avoid overlong lines as much as possible if the manpage gets formatted with `groff -ww -Tpdf -mdoc -rS12 portfile.7 > portfile.7.pdf` – having too small a body font makes the long lines of the text hard to read.  I did a grep for `ui_msg` in the 'macports_ports' git repository and arbitrarily took a short string.

ryandesign: `notes`, which prints a message after installation, has indeed largely made `ui_msg` unnecessary. There are a few ports that use `ui_msg` to print a message during installation, such as to notify the user that a fetch or a build may take longer than usual (bazel, InsightToolkit, inkscape-devel); I'd change the example to something like that.

```
.Dl ui_msg \*qThis build may take a very long time\*q
```

lemzwerg: Done.

raimue:
Even if it does not change much in the content, I think we should merge this.

lemzwerg:
OK, I've now done a forced push with a single commit, omitting the 'useless' source-code structuring, also avoiding (almost all) unnecessary reformatting.

doc/portfile.7
```diff
@@ -102,42 +104,57 @@ decreased.
...
-One line description of the software and what it does.  To appear in the
-description, brackets and semi-colons need to be escaped with a backslash (i.e.
-they must be written as "\\[", "\\]" and "\\;").  The escape sequences listed in
+One-line description of the software and what it does.
+To appear in the description, brackets and semi-colons need to be escaped with
+a backslash (i.e., they must be written as
+.Ql \e[ ,
+.Ql \e]
+and
+.Ql \e; ) .
```
mojca: This renders as
```
(i.e., they must be written as `\[', `\]' and `\;)'
```
Shouldn't the closing `)` come after the quotation mark?

doc/portfile.7
```diff
@@ -102,42 +104,57 @@ decreased.
...
-A verbose description of the software and what it does.  To appear in the
-description, brackets and semi-colons need to be escaped with a backslash (i.e.
-they must be written as "\\[", "\\]" and "\\;").  The escape sequences listed in
+A verbose description of the software and what it does.
+To appear in the description, brackets and semi-colons need to be escaped with
+a backslash (i.e., they must be written as
+.Ql \e[ ,
+.Ql \e]
+and
+.Ql \e;) .
```
mojca: This renders as
```
(i.e., they must be written as `\[', `\]' and `\;)'
```
Shouldn't the closing `)` come after the quotation mark?

doc/portfile.7
```diff
@@ -102,42 +104,57 @@ decreased.
...
-
-    (GIMP) is a powerful tool for the preparation and \e\
-
-    manipulation of digital images. The GIMP provides \e\
-
-    the user with a wide variety of image manipulation, \e\
-
-    painting, processing, and rendering tools.
+long_description The GNU Image Manipulation \e
+    Program (GIMP) is a powerful tool for the \e
```
mojca: This has nothing to do with your changes, but we actually use a slightly different spacing (20 spaces) :)

doc/portfile.7
```diff
@@ -293,37 +328,40 @@ can be omitted if there is only one distfile.
...
+checksums dictd-1.7.1.tar.gz md5 81317b86ea0a5df0163900ad2e6bb12c \e
+          magic-words-1.7.1.tar.gz md5 897a005182928613eadd30c267ce9c5b
 .Ed
 .br
 .Sy Example (ledit 1.11):
 .Bd -literal -offset indent -compact
-checksums md5 a2d38ba641682509c1e964ad699a9dd2 \e\ 
-	sha1 1fb6443b5fdf3c83787953f06282d256477c1288
+checksums md5 a2d38ba641682509c1e964ad699a9dd2 \e
+          sha1 1fb6443b5fdf3c83787953f06282d256477c1288
```
mojca: Not in the scope of this PR, but this definitely needs to be corrected. We now want `rmd160`, `sha256`, `size`. By providing `sha1` and `md5` we give users exceptionally bad examples.

doc/portfile.7
```diff
@@ -590,59 +638,67 @@ Location to store downloaded distfiles.
 .Sy Default:
 .Em ${portdbpath}/distfiles/${dist_subdir}/
 .It Ic os.arch
-Identifies hardware type (e.g. "powerpc").
+Identifies hardware type (e.g.,
+.Dq powerpc ) .
```
mojca: This has a different quoting from some instances of `powerpc` mentioned above (where it was underlined) and different from some other quoted variables (with a single quote) written above. Is there a special reason for that?

doc/portfile.7
```diff
@@ -655,26 +711,33 @@ Port dependencies should refer to other MacPorts ports
 whenever possible, therefore each dependency should be
 expressed in the format:
 .Bl -tag -width ls
-.It port: Ns Ao port Ac
+.It Li port: Ns Aq port
```
mojca: Does it make sense to add some indentation here?

doc/portfile.7
lemzwerg: Yep.  A typo, now fixed.

doc/portfile.7
lemzwerg: Done.

doc/portfile.7
lemzwerg: I don't see another quoted instance of `powerpc`.  Which line do you mean?  And where do you see single quotes?

PR #108, opened 2018-10-20
add archflags to FFLAGS environment variable

MarcusCalhoun-Lopez:


pmetzger:
Could you edit the description so we'll know why you want to do this? Makes reading your change easier.

jmroot:
This is not the correct way to fix this. The bug appears to have been introduced in aa375e5, where it was incorrectly assumed that the environment variable corresponding to `$tool` is always `[string toupper $tool]FLAGS`. This does not hold for f77 and FFLAGS.

A special case needs to be added on line 904 in master.

PR #109, opened 2018-11-16
portconfigure.tcl add -Wc,-isysroot to LDFLAGS

kencu:
clang does not prepend -syslibroot path to link search path
it appends it to the search path set in -isysroot

unless an -isysroot is sent to clang during the link, it uses
the default -isysroot instead

sending the -isysroot to clang during link uses the requested
SDK for link instead of the default

closes: https://trac.macports.org/ticket/57612

### DIFF
--- a/src/port1.0/portconfigure.tcl
+++ b/src/port1.0/portconfigure.tcl
@@ -892,6 +892,7 @@ proc portconfigure::configure_main {args} {
             foreach env_var {CPPFLAGS CFLAGS CXXFLAGS OBJCFLAGS OBJCXXFLAGS} {
                 append_to_environment_value configure $env_var -isysroot${configure.sdkroot}
             }
+            append_to_environment_value configure "LDFLAGS" -Wc,-isysroot,${configure.sdkroot}
             append_to_environment_value configure "LDFLAGS" -Wl,-syslibroot,${configure.sdkroot}
         }
 


jeremyhu:
Why not just add LDFLAGS to the line above such that it gets -isysroot?  Why the -Wc,?  That will trip up xcrun whereas the former won't.

kencu:
That would indeed be much simpler, once I can get the fixed-up `libtool` you mentioned (that doesn't strip the `-isysroot` out of the link line) to be the one we use all the time. Seems like we pretty much want that one to be the default libtool.

The `-Wc` business was only to force all the `non-fixed-up libtools` (which no doubt were all the ones that generated errors for me) to pass `-isysroot` to clang during the link.

kencu:
There was one other reason to wrap `-isysroot `in the `-Wc,` guard rather than just put it in `LDFLAGS`; I found if somehow -isysroot made it's way to `ld` it just generates an error during the link.

And then there's `scons` and the myriad of `use_configure no` ports and `gcc` bootstrapping, and `haskell` and  ... that I couldn't fully predict. So wrapping it in `-Wc` made sure it would only go to `clang` or `gcc` I thought.

jeremyhu:
Yes, there is historic confusion over the consumer of `$LDFLAGS`.  It is expected to be passed to the compiler front end, but some build systems treat `$LDFLAGS` as flags to be passed directly to ld.

However, ld should similarly bawk if it sees -Wc,-isysroot passed to it.

That myriad of "other" ports that don't use configure should be honoring `configure.sdkroot` directly then.  Eg cmake does:

    if {${configure.sdkroot} != ""} {
        configure.env-append CMAKE_OSX_SYSROOT="${configure.sdkroot}"
    }

jeremyhu:
What about additionally setting the SDKROOT environment variable then?

kencu:
> However, ld should similarly bawk if it sees -Wc,-isysroot passed to it.

you are totally right about that. It just errors out, of course. So that is no real solution either, you're right.
```
ld: unknown option: -Wc,-isysroot,/
```
> What about additionally setting the SDKROOT environment variable then?
that seems like the best idea at the moment. I will play with that and see how it works out.

PR #111, opened 2018-12-04
Fix archive_site_local, master_site_local, patch_site_local when multiple URLs are specified

ryandesign:
Correctly add multiple URLs specified in `archive_site_local`, `master_site_local`, and `patch_site_local` when a tag is used.

Closes: https://trac.macports.org/ticket/57718


jmroot:
I'm not sure these were ever intended to support more than one URL; after all the name contains "site" not "sites". But if it's this easy to support it, why not I suppose.

Some test cases would be very useful to encapsulate what the correct behaviour is in each of the many combinations of master_sites, master_site_local, single/multiple URLs in each, single/multiple/no tags used, etc.

jmroot:
This does add the additional constraint that *_SITE_LOCAL must be a valid Tcl list, but it should be pretty uncommon for that to be a problem.

PR #112, opened 2018-12-13
darwintrace: build 32-bit library on macOS Mojave

danchr:
Certain ports, such as qemu, break when using trace mode under Mojave due to invoking 'Rez', a 32-bit binary. This change fixes them by forcing `darwintrace` to build against the libraries on the system root.

### DIFF
--- a/aclocal.m4
+++ b/aclocal.m4
@@ -752,10 +752,10 @@ AC_DEFUN([MP_UNIVERSAL_OPTIONS],[
 
 	if test "x$UNIVERSAL_ARCHS" = "x"; then
 		case "$MACOSX_VERSION" in
-			10.1[[0-3]]*)
+			10.1[[0-4]]*)
 				UNIVERSAL_ARCHS="x86_64 i386"
 			;;
-			10.1[[4-9]]*)
+			10.1[[5-9]]*)
 				UNIVERSAL_ARCHS="x86_64"
 			;;
 			10.[[0-5]]*)
@@ -771,6 +771,12 @@ AC_DEFUN([MP_UNIVERSAL_OPTIONS],[
 	for arch in $UNIVERSAL_ARCHS; do
 		UNIVERSAL_ARCHFLAGS="$UNIVERSAL_ARCHFLAGS -arch $arch"
 	done
+
+	case "$MACOSX_VERSION" in
+			10.14*)
+				UNIVERSAL_ARCHFLAGS="$UNIVERSAL_ARCHFLAGS -Wl,-syslibroot,/"
+			;;
+	esac
 
 	AC_MSG_CHECKING([for Universal CPU architectures])
 	AC_MSG_RESULT([$UNIVERSAL_ARCHS])
--- a/configure
+++ b/configure
@@ -6405,10 +6405,10 @@ fi
 
 	if test "x$UNIVERSAL_ARCHS" = "x"; then
 		case "$MACOSX_VERSION" in
-			10.1[0-3]*)
+			10.1[0-4]*)
 				UNIVERSAL_ARCHS="x86_64 i386"
 			;;
-			10.1[4-9]*)
+			10.1[5-9]*)
 				UNIVERSAL_ARCHS="x86_64"
 			;;
 			10.[0-5]*)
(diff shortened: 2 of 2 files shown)


neverpanic:
I'm not sure whether regenerating `configure` must happen after modifying `aclocal.m4` to prevent the build system from automatically re-generating it because it considers is out of date. Other than that, this lg2m.

ryandesign:
If this is all that is needed to build universal on Mojave, why aren't we doing that in portconfigure.tcl to be able to build ports universal too?

kencu:
This is unlikely to work. As we know, syslibroot does nothing useful on MacOS, and all 32 bit support has been stripped out of the 10.14 headers.

jeremyhu:
Also, Rez is 64bit, so your original comment seems off too:

$ file /usr/bin/Rez
/usr/bin/Rez: Mach-O 64-bit x86_64 executable, flags:<NOUNDEFS|DYLDLINK|TWOLEVEL|PIE>

danchr:
> Also, Rez is 64bit, so your original comment seems off too:
> 
> $ file /usr/bin/Rez
> /usr/bin/Rez: Mach-O 64-bit x86_64 executable, flags:<NOUNDEFS|DYLDLINK|TWOLEVEL|PIE>

That's not the actual `Rez` binary, but a wrapper:

```
$ otool -L /usr/bin/Rez
/usr/bin/Rez:
	/usr/lib/libxcselect.dylib (compatibility version 1.0.0, current version 1.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1252.200.5)
$ locate '*/Rez' | grep -v /opt/local | xargs file
/Applications/Xcode.app/Contents/Developer/Tools/Rez:   Mach-O executable i386
/Applications/Xcode.app/Contents/Developer/usr/bin/Rez: Mach-O executable i386
/Library/Developer/CommandLineTools/usr/bin/Rez:        Mach-O executable i386
/usr/bin/Rez:                                           Mach-O 64-bit executable x86_64
```

danchr:
> This is not the right approach. The 10.14 SDK does not support i386. If you want to build i386, put the 10.13 SDK in /Library/Developer/CommandLineTools/SDKs and configure MacPorts to use it instead of the 10.14 SDK .

1) `Rez` is a 32-bit binary.
2) Some ports currently use `Rez`.
3) Invoking Rez works outside trace mode.
4) Invoking Rez within trace mode does not work, as `darwintrace` is injected and 64-bit only.

screenshot 2019-02-18 at 11 18 09

I'm not arguing that we should build 32-bit binaries in general, but that _specifically_ `darwintrace.dylib` should be built as a universal binary since it links against nothing but `libSystem.dylib` — which remains universal.

FWIW these legacy binaries remain 32-bit only: BuildStrings, CpMac, DeRez, GetFileInfo, MergePef, MvMac, ResMerger, Rez, RezDet, RezWack, SetFile, SplitForks & UnRezWack.

kencu:
sometimes using -Wl,-syslibroot can work to force a 32 bit build on Mojave, if the build system is invoking LD directly and not calling clang to do the link (as per the ticket I referenced). If you call clang, then syslibroot will not change the syslibroot because -isysroot is put in first and overrides it. 

It is also possible to get lucky and the headers you are calling in for the particular 32bit build happen to not be 32bit sensitive, and all the 32bit header stripping that occurred in the Mojave headers just don't affect that build. But you will find that many/most 32bit builds do fail against the Mojave SDK even passing syslibroot.

neverpanic:
What's the executive summary of the state of this PR? Should we merge this or just accept that Rez is broken in trace mode and hope that Apple will eventually update it to 64 bit?

kencu:
FYI, something might have changed:
```
$ file /Applications/Xcode.app/Contents/Developer/usr/bin/Rez
/Applications/Xcode.app/Contents/Developer/usr/bin/Rez: Mach-O universal binary with 2 architectures: [x86_64:Mach-O 64-bit executable x86_64] [i386:Mach-O executable i386]
/Applications/Xcode.app/Contents/Developer/usr/bin/Rez (for architecture x86_64):	Mach-O 64-bit executable x86_64
/Applications/Xcode.app/Contents/Developer/usr/bin/Rez (for architecture i386):	Mach-O executable i386
```

neverpanic:
So this may in fact no longer be necessary and we can let 32-bit support die?

kencu:
> So this may in fact no longer be necessary and we can let 32-bit support die?

It certainly appears Apple has 64_bit upgraded Rez at least since this PR was first put forth. Whether there are other parts that remain 32bit only, and whether dylib injection works fully now, I couldn't say. Perhaps @danchr might comment.

danchr:
> > So this may in fact no longer be necessary and we can let 32-bit support die?
> 
> It certainly appears Apple has 64_bit upgraded Rez at least since this PR was first put forth. Whether there are other parts that remain 32bit only, and whether dylib injection works fully now, I couldn't say. Perhaps @danchr might comment.

From a quick inspection, the only 32-bit-only binaries remaining are explicitly named `*32` — I can't say whether anything actually uses them, but I'd expect not. To say that dylib injection “works fully” might be a bit of an overstatement; I think it still has issues with Apple-signed binaries.

kencu:
@danchr -- noted. Leave you to close this, then?

PR #113, opened 2019-02-14
Fix a crash due to an out-of-bounds array access

saagarjha:
entry_search had a "type confusion" bug where it would perform array accesses that bypassed bounds checks if given arguments that looked like strategies. I've rewritten most of the argument parsing loop in a way that should eliminate this issue, while being shorter and more straightforward (with less duplicated code), only requiring one pass through the arguments. The downside is that we allocate a bit more memory and have a couple of gotos to handle errors.

You can reproduce this crash with `port install [some port name] --`, if you'd like to test it. If there's something stylistically wrong with my pull request, please feel free to let me know–I'm not completely sure how you'd like it formatted.

### DIFF
--- a/src/registry2.0/entry.c
+++ b/src/registry2.0/entry.c
@@ -248,116 +248,92 @@ static strategy_type strategies[] = {
  * Can be given an option of -exact, -glob, -regexp or -null to specify the
  * matching strategy; defaults to exact.
  */
-static int entry_search(Tcl_Interp* interp, int objc, Tcl_Obj* CONST objv[]) {
-    int i, j;
-    reg_registry* reg = registry_for(interp, reg_attached);
-    if (reg == NULL) {
+static int entry_search(Tcl_Interp* interp, int objc, Tcl_Obj* CONST *objv) {
+    reg_registry *reg = registry_for(interp, reg_attached);
+    if (!reg) {
         return TCL_ERROR;
+    }
+    const char **keys = malloc(objc * sizeof(char *));
+    const char **vals = malloc(objc * sizeof(char *));
+    int *strats = malloc(objc * sizeof(int));
+    if (!keys || !vals || !strats) {
+        goto cleanup_error;
+    }
+    int key_count = 0;
+
+    /* Skip the "registry::entry" "search" bit */
+    objv += 2;
+    objc -= 2;
+
+    for (int i = 0; i < objc; ++key_count /* i in incremented in the loop */) {
+        int key_index, strategy_index;
+
+        /* Grab the key */
+        if (Tcl_GetIndexFromObj(interp, objv[i], entry_props, "search key",
+            0, &key_index) != TCL_OK) {
+            goto cleanup_error;
+        }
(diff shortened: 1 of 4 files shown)


saagarjha:
I'm not sure why the tests are failing; did my change affect these parts of MacPorts as well?

mojca:
It seems that the tests already failed in the same way when @MarcusCalhoun-Lopez merged #88.

jmroot:
Yes, that commit broke the tests: https://trac.macports.org/ticket/57947

saagarjha:
How have the other pull requests managed to pass the tests? Is there some way I should configure the CI runner to not get tripped up on this?

jmroot:
I'll try to fix the test fallout from #88 tonight. There are other issues with it that I've been trying to find time to address, but this one part shouldn't be too hard.

mojca:
> How have the other pull requests managed to pass the tests?

By being submitted before the problematic commit sneaked in?

> Is there some way I should configure the CI runner to not get tripped up on this?

If you really really really wanted to do this, you could add a bogus commit reverting the changes from the problematic commit(s), but ideally we need to fix this in the master branch. Once that is done, you would only rebase your changes.

From the lesson learned: it would be awesome to provide some tests for the functionality you just added (so that the new tests would fail in master and pass after your fixes).

jmroot:
Tests are fixed. Not sure if you can trigger a new CI run without pushing a commit though.

mojca:
Reruning won't help, even if it was trivial to do it. Code needs to be rebased, so that your changes will be seen by the CI.

saagarjha:
I've added a new test for this change, but I'm not really great at Tcl so I'm not sure if I did it correctly.

src/registry2.0/tests/entry.tcl
```diff
@@ -74,6 +74,9 @@ proc main {pextlibname} {
         test_set {[registry::entry search name -glob vi*]} {$vim1 $vim2 $vim3}
         test_set {[registry::entry search name -regexp {zlib|pcre}]} \
             {$zlib $pcre}
+
+        # test that passing in confusing arguments doesn't crash
+        test {[catch {[registry::entry search name vim1 --]}] == 1}
```
jmroot: The innermost pair of square brackets are a syntax error. This means catch will return 1 even if the args are something valid.

saagarjha: Thanks, I think I've fixed this now.

src/registry2.0/entry.c
```diff
@@ -248,116 +248,89 @@ static strategy_type strategies[] = {
...
-                }
-
-                i++;
-            }
-
-            key_count++;
+    }
+    const char **keys = malloc(objc * sizeof(char *));
+    const char **vals = malloc(objc * sizeof(char *));
+    int *strats = malloc(objc * sizeof(int));
```
jmroot: Need to check for failure here.

saagarjha: Check added a line down.

jmroot:
Thanks.

PR #114, opened 2019-02-25
If with-no-root-privileges, set without-startupitems=yes

kurthindenburg:
configure --with-no-root-privileges implies --without-startupitems=yes
Otherwise, startupitems will attempt to be installed under /Library

closes: https://trac.macports.org/ticket/56743

### DIFF
--- a/configure
+++ b/configure
@@ -6167,6 +6167,8 @@ $as_echo "$mpconfigdir" >&6; }
 # Check whether --with-no-root-privileges was given.
 if test "${with_no_root_privileges+set}" = set; then :
   withval=$with_no_root_privileges; ROOTPRIVS=$withval
+  # with-no-root-privileges implies without-startupitems
+  with_startupitems=no
 fi
 
 


mojca:
I agree with the idea here, but the configure script is a generated file. You need to change either `aclocal.m4` or `configure.ac`. I don't know the code, but I need to admit that the variable name is strange if `ROOTPRIVS` is only set when no root privileges are there. It sounds reversed.

kurthindenburg:
Argh, I committed the wrong file - I'll look at it again

neverpanic:
@kurthindenburg gentle reminder :)

kurthindenburg:
Yea thanks.  I looked at this again and the fix doesn't appear straight forward.   Base seems to use ./configure but doesn't appear to regenerate it.  I'll look at it more.

raimue:
You need to run the `./autogen.sh` command to regenerate the `configure` script from `configure.ac`. The code you want to change would be in [aclocal.m4](https://github.com/macports/macports-base/blob/0ef3362e553d8a7ee6467ca188ca29e1a524728f/aclocal.m4#L546-L576).

Hope that helps!

kurthindenburg:
This is taking too long - I'll redo when I have more time.

PR #115, opened 2019-03-03
Change puts to ui_msg in reclaim.tcl

arghyatiger:
```reclaim.tcl``` states change from ```puts``` to ```ui_msg``` as a finished task in line 45. 
But I noticed that one single puts instance wasn't changed, and as far as I understand there isn't any need for it to be ```puts```.

### DIFF
--- a/src/macports1.0/reclaim.tcl
+++ b/src/macports1.0/reclaim.tcl
@@ -72,7 +72,7 @@ namespace eval reclaim {
             write_last_run_file disabled
             return
         }
-
+        ui_msg "$macports::ui_prefix Initializing Reclaim"
         uninstall_unrequested
         uninstall_inactive
         remove_distfiles
@@ -339,7 +339,7 @@ namespace eval reclaim {
         set fd -1
         try -pass_signal {
             set fd [open $path w]
-            puts $fd $contents
+            ui_msg " $fd $contents"
         } catch {*} {
             # Ignore error silently
         } finally {
--- a/src/macports1.0/reclaim.tcl
+++ b/src/macports1.0/reclaim.tcl
@@ -72,7 +72,7 @@ namespace eval reclaim {
             write_last_run_file disabled
             return
         }
-        ui_msg "$macports::ui_prefix Initializing Reclaim"
+        
         uninstall_unrequested
         uninstall_inactive
         remove_distfiles

### CODE REVIEW

File: src/macports1.0/reclaim.tcl [hunk 1]
```diff
@@ -72,7 +72,7 @@ namespace eval reclaim {
             write_last_run_file disabled
             return
         }
-
+
```
mojca: Bogus whitespace was probably not intended here?

arghyatiger: yes, I'll make a lint commit

arghyatiger: @mojca  I'll make a lint commit for that :D

arghyatiger: @mojca  Changing whitespace doesn't show as a valid change on git. How do I fix this?

mojca: That sounds unusual. What does `git status` say? It could be that you either forgot to do `git add`, or that you added the file, but then forgot the `--cached` flag in `git diff --cached` when you asked for changes. Or maybe you forgot to save the file, or the editor is playing with you (I have the editor configured in such a way that it always displays invisible characters, so that I can spot the difference between tabs vs. spaces, trailing whitespace etc.).

src/macports1.0/reclaim.tcl
```diff
@@ -339,7 +339,7 @@ namespace eval reclaim {
         set fd -1
         try -pass_signal {
             set fd [open $path w]
-            puts $fd $contents
+            ui_msg " $fd $contents"
```
mojca: How did you test the change?

jmroot:
This change is incorrect: `puts $fd $contents` writes $contents to the file that was just opened, whereas `ui_msg " $fd $contents"` displays the fd identifier followed by $contents to the user.

mojca:
Eh ... I wanted the student to figure that out on his own, but the spoiler came too quickly :)

src/macports1.0/reclaim.tcl
arghyatiger: @mojca  just realized my mistake. Sorry about the bogus PR

MarcusCalhoun-Lopez: @arghyatiger, thank you for the pull request.
Even if this one didn't quite work out, we do appreciate your efforts.
We are hoping you will submit many more pull requests that will get merged.

PR #116, opened 2019-03-19
Treat command.env as a list

jmroot:
The previous behaviour involved a complicated regex, made it very
difficult to set an environment variable value including quote
characters, and was responsible for many cases of variables being in
foo.env but not actually appearing in the environment.

Portfiles will need to be reviewed for cases of writing `foo.env-append VAR="foo bar"` since that adds two elements to foo.env, `VAR="foo` and `bar"` (double quotes only have a special meaning to Tcl if they appear at the start of a word).

### DIFF
--- a/src/port1.0/portutil.tcl
+++ b/src/port1.0/portutil.tcl
@@ -820,12 +820,14 @@ proc parse_environment {command} {
     global ${command}.env ${command}.env_array
 
     if {[info exists ${command}.env]} {
-        # Flatten the environment string.
-        set the_environment [join [set ${command}.env]]
-
-        while {[regexp "^(?: *)(\[^= \]+)=(\"|'|)(\[^\"'\]*?)\\2(?: +|$)(.*)$" ${the_environment} matchVar key delimiter value remaining]} {
-            set the_environment ${remaining}
-            set ${command}.env_array(${key}) ${value}
+        foreach assignment [set ${command}.env] {
+            set equals_pos [string first = $assignment]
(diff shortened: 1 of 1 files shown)


Ionic:
> made it very difficult to set an environment variable value including quote characters

If I understood the old regexp correctly, this is an understatement. It was impossible to have ANY quote character be part of the value, unless immediately followed by the first detected quoting character.

Something like:
`FOO="bar'"`
would have worked, while
`FOO="bar'baz"`
would have not matched the regexp in the first place, ignoring the entry and others after it. No way to get around that.


We should also make very clear that this will break backwards compatibility.

It's not just Portfile that used `...env FOO="bar baz"`, but also `...env FOO=\"bar baz\"` and probably other variations.

jmroot:
If you mean in terms of number of affected portfiles, there are currently 991 that contain `.env` and will need to be reviewed. Likely only a fraction of those will need changes, but it's reasonably common among ports that don't use a configure script and pass CFLAGS etc to a Makefile via the environment.

jmroot:
I've started reviewing the portfiles and adding changes conditional on the base version where necessary (and removing some overquoting that was never needed). Once I'm done I'll push the portfile changes and merge this at the same time.

neverpanic:
Right, thanks for looking into it. Should we merge this PR then?

jmroot:
Things will break for anyone running master if this is merged and not the portfile changes. So I would want to merge both at close to the same time.

PR #118, opened 2019-03-30
auto*: bump config.guess and config.sub to 2019-03-04

grimreaper:
These are primarily non-functional changes for us but reduce the diff in
the future should we ever need to update again.


grimreaper:
This is likely safe to merge but sending as a PR in order to ensure a second level of review and testing.

PR #119, opened 2019-03-31
tracelib.c: Removed unnecessary condition in if

MihirLuthra:
if ((kev.flags & EV_ERROR) == 0 || ((kev.flags & EV_ERROR) > 0 && kev.data != 0))
If compilers find first operand true while evaluating || , the second operand isn’t checked.
So the second operand is only checked if (kev.flags & EV_ERROR) != 0.
So (kev.flags & EV_ERROR) > 0  it doesn’t need to be checked in conjunction with kev.data != 0.
It doesn’t return negative.(tested)

### DIFF
--- a/src/pextlib1.0/tracelib.c
+++ b/src/pextlib1.0/tracelib.c
@@ -771,7 +771,7 @@ static int TracelibRunCmd(Tcl_Interp *in) {
         /* kevent(2) on EV_RECEIPT: When passed as input, it forces EV_ERROR to
          * always be returned. When a filter is successfully added, the data field
          * will be zero. */
-        if ((kev.flags & EV_ERROR) == 0 || ((kev.flags & EV_ERROR) > 0 && kev.data != 0)) {
+        if ((kev.flags & EV_ERROR) == 0 || (kev.data != 0)) {
             error2tcl("kevent (listen socket receipt): ", kev.data, in);
             goto error_locked;
         }
--- a/src/port1.0/porttrace.tcl
+++ b/src/port1.0/porttrace.tcl
@@ -155,13 +155,13 @@ namespace eval porttrace {
         create_slave $workpath $fifo
 
         # Launch darwintrace.dylib.
-        set tracelib [file join ${portutil::autoconf::tcl_package_path} darwintrace1.0 darwintrace.dylib]
+        set darwintracepath [file join ${portutil::autoconf::tcl_package_path} darwintrace1.0 darwintrace.dylib]
 
         # Add darwintrace.dylib as last entry in DYLD_INSERT_LIBRARIES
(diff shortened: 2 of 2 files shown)


jmroot:
Please rebase your branch. https://trac.macports.org/wiki/WorkingWithGit#upstream-fetch

jmroot:
> It doesn’t return negative.(tested)

To determine this by testing, you would have to generate every possible error condition and see what is returned. I really doubt you did that. :)

Fortunately, you don't actually have to test because flags is a uint16_t and thus can't be negative by definition.

jmroot:
One more thing: this pattern is repeated in a few other places in the file. Why not change all of them?

MihirLuthra:
On Mon, 1 Apr 2019, 07:35 Joshua Root, <notifications@github.com> wrote:

> It doesn’t return negative.(tested)
>
> To determine this by testing, you would have to generate every possible
> error condition and see what is returned. I really doubt you did that. :)
>
> Fortunately, you don't actually have to test because flags is a uint16_t
> and thus can't be negative by definition.
>
Glad u told me that uint16_t. ^_^

i tried to research there types but cudnt really find.
By tested, i didn't mean like that.

There is already no handling for negative case even if it arises.
Also there are only three flags being formed in there EV_ADD, EV_RECEIPT &
EV_ERROR. I just researched on them. and pretty much combination of these 3
just give arise to 8 conditions.

That is a what i meant to say.

Sorry if i was wrong, i will take more care from next time.

Regards,
Mihir

MihirLuthra:
On Mon, 1 Apr 2019, 07:38 Joshua Root, <notifications@github.com> wrote:

> One more thing: this pattern is repeated in a few other places in the
> file. Why not change all of them?
>
> To be honest, i noticed that now.
I will rebase my branch and change it if it makes sense. :)

Regards
Mihir

jmroot:
> Glad u told me that uint16_t. ^_^
> i tried to research there types but cudnt really find.

If you look at the kevent man page, it refers to a few headers, one of which is sys/event.h. That contains the definition of struct kevent.

MihirLuthra:
Aah, srry for that. My mistake! That didn't come to my attention.
Will be more careful next time.

jmroot:
No need to apologise. This is meant to be a learning experience after all. :)

MihirLuthra:
> No need to apologise. This is meant to be a learning experience after all. :)

Thanks ^_^,
I will rebase my branch and submit a new PR :)

jmroot:
Oh, there should be no need for a new PR. If you update your fork with the commits from macports/macports-base, the PR will update automatically.

One note: the usual GitHub procedure would be to create a new branch in your fork for each PR, and then delete the branch after the PR is merged. That way you can keep master in your fork the same as master in our repo.

MihirLuthra:
Thanks for the helpful info, I will do exactly that. :)

Mihir

>

PR #120, opened 2019-04-02
port1.0: add portbump.tcl

satraul:
This is still a WIP, part of my GSOC 2019 proposal. Any feedback is greatly appreciated

Backlog:
* [X] Calculate checksums
* [X] Raise privileges to write to Portfile
* [X] Implement checksum replacement using regex
* [X] Implement revision number replacement using regex
* [X] Implement create patch use case (`--patch`)
* [ ] Create test suites
* [ ] Create documentation

For context: the `port bump` action would bump Portfile version and checksums as requested here: https://trac.macports.org/ticket/53851

My current implementation is a "bump" stage which requires the "main" and "fetch" stages to be
sucessfully run first. portbump.tcl largely reuses code from portchecksums.tcl. The desired output is something like this:
```
$ port bump
--->  Fetching distfiles for chromedriver
--->  Verifying checksums for chromedriver
Old sha256: 00b6ac50b0de0bd63d260cc7472c6e6af44b1240
New sha256: 2aa256d17e4b2cc21f888b0e1b9ed211b925bf40
Old size: 6809243
New size: 7057233
Do you approve of these changes [Y/n]? Y
Do you want to reset revision number (currently 2) to 0 [Y/n]? Y
Version successfully updated. Suggested commit message:
    chromedriver: update to 6.6.3
```

I did a "draft pull request" which is a new github feature. Hope it's okay.

Thank you.

### DIFF
--- a/src/port1.0/portbump.tcl
+++ b/src/port1.0/portbump.tcl
@@ -1,0 +1,263 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+# portbump.tcl
+#
+# Copyright (c) 2019 The MacPorts Project
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+# 1. Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+# 2. Redistributions in binary form must reproduce the above copyright
+#    notice, this list of conditions and the following disclaimer in the
+#    documentation and/or other materials provided with the distribution.
+# 3. Neither the name of Apple Inc. nor the names of its contributors
+#    may be used to endorse or promote products derived from this software
+#    without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+# AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+# ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+# LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+# CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+# SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+package provide portbump 1.0
+package require portutil 1.0
+package require portchecksum 1.0
+
+set org.macports.bump [target_new org.macports.bump portbump::bump_main]
+target_provides ${org.macports.bump} bump
+target_runtype ${org.macports.bump} always
+target_requires ${org.macports.bump} main fetch
+target_prerun ${org.macports.bump} portbump::bump_start
+
+namespace eval portbump {
+}
+
+# bump_start
+#
+# Target prerun procedure; simply prints a message about what we're doing.
+#
+proc portbump::bump_start {args} {
+    global UI_PREFIX
+
+    ui_notice "$UI_PREFIX [format [msgcat::mc "Bumping checksums for %s"] [option subport]]"
+}
+
+# bump_main
+#
+# Target main procedure. Bumps the checksums for distfiles.
+#
+proc portbump::bump_main {args} {
+    global UI_PREFIX all_dist_files checksums_array portpath ports_bump_patch
+
+    set portfile "${portpath}/Portfile"
+
+    # If no files have been downloaded, there is nothing to bump.
+    if {![info exists all_dist_files]} {
+        return 0
+    }
+
+    # So far, no mismatches yet.
+    set mismatch no
+
+    # Set the list of checksums as the option checksums.
+    set checksums_str [option checksums]
+
+    # Store the calculated checksums to avoid repeated calculations
+    array set calculated_checksums_array {}
+
+    # If everything is fine with the syntax, keep on and check the checksum of
+    # the distfiles.
+    if {[portchecksum::parse_checksums $checksums_str] eq "yes"} {
+        set distpath [option distpath]
+
+        foreach distfile $all_dist_files {
+            ui_info "$UI_PREFIX [format [msgcat::mc "Checksumming %s"] $distfile]"
+
+            # Get the full path of the distfile.
+            set fullpath [file join $distpath $distfile]
+            if {![file isfile $fullpath]} {
+                return -code error "$distfile does not exist in $distpath"
+            }
+
+            # Check that there is at least one checksum for the distfile.
+            if {![info exists checksums_array($distfile)] || [llength $checksums_array($distfile)] < 1} {
+                ui_error "[format [msgcat::mc "No checksum set for %s"] $distfile]"
+                set mismatch yes
+            } else {
+                # Retrieve the list of types/values from the array.
+                set portfile_checksums $checksums_array($distfile)
+                set calculated_checksums {}
+                set both_checksums {}
+
+                # Iterate on this list to check the actual values.
+                foreach {type sum} $portfile_checksums {
+                    set calculated_sum [portchecksum::calc_$type $fullpath]
+                    lappend calculated_checksums $type
+                    lappend calculated_checksums $calculated_sum
+                    lappend both_checksums $type $sum $calculated_sum
+
+                    if {$sum eq $calculated_sum} {
+                        ui_debug "[format [msgcat::mc "Correct (%s) bump for %s"] $type $distfile]"
+                    } else {
+                        ui_info "[format [msgcat::mc "Portfile bump: %s %s %s"] $distfile $type $sum]"
+                        ui_info "[format [msgcat::mc "Distfile bump: %s %s %s"] $distfile $type $calculated_sum]"
+
+                        # Raise the failure flag
+                        set mismatch yes
+                    }
+                }
+
+                # Save our calculated checksums in case we need them later
+                set calculated_checksums_array($distfile) $calculated_checksums
+
+                if {![regexp {\.html?$} ${distfile}] &&
+                    ![catch {strsed [exec [findBinary file $portutil::autoconf::file_path] $fullpath --brief --mime] {s/;.*$//}} mimetype]
+                    && "text/html" eq $mimetype} {
+                    # file --mime-type would be preferable to file --mime and strsed, but is only available as of Snow Leopard
+                    set wrong_mimetype yes
+                    set htmlfile_path ${fullpath}.html
+                    file rename -force $fullpath $htmlfile_path
+                }
+            }
+        }
+    } else {
+        # Something went wrong with the syntax.
+        return -code error "[msgcat::mc "Unable to verify file checksums"]"
+    }
+
+    if {![tbool mismatch]} {
+        ui_msg "No changes needed."
+        return 0
+    }
+
+    if {[tbool wrong_mimetype]} {
+        # We got an HTML file, though the distfile name does not suggest that one was
+        # expected. Probably a helpful DNS server sent us to its search results page
+        # instead of admitting that the server we asked for doesn't exist, or a mirror that
+        # no longer has the file served its error page with a 200 response.
+        ui_notice "***"
+        ui_notice "The non-matching file appears to be HTML. See this page for possible reasons"
+        ui_notice "for the bump mismatch:"
+        ui_notice "<https://trac.macports.org/wiki/MisbehavingServers>"
+        ui_notice "***"
+        ui_notice "The file has been moved to: $htmlfile_path"
+        
+        return -code error "[msgcat::mc "Unable to verify file checksums"]"
+    } else {
+        # Show the desired checksum line for easy cut-paste
+        # based on the previously calculated values, plus our default types
+        set sums {}
+
+        global version revision
+
+        ui_msg "We will bump these:"
+        foreach {type sum calculated_sum} $both_checksums {
+            ui_msg [format "Old %-8s %s" ${type}: $sum]
+            ui_msg [format "New %-8s %s" ${type}: $calculated_sum]
+        }
+
+        set patterns {}
+
+        set whitespace {[[:space:]\\]+}
+        # Create substitution pattern(s) for checksum
+        foreach {type sum calculated_sum} $both_checksums {
+            lappend patterns "s/(${type}${whitespace})${sum}/\\1${calculated_sum}/g"
+        }
+        # Create substitution pattern for revision (reset to 0)
+        lappend patterns {s/(revision[[:space:]\\]+)[0-9]+/\10/g}
+
+        # Construct sed command
+        set cmdline {}
+        lappend cmdline $portutil::autoconf::sed_command -E
+        foreach pattern $patterns {
+            lappend cmdline -e $pattern
+        }
+
+        # Get the uid of Portfile owner
+        set owneruid [name_to_uid [file attributes ${portfile} -owner]]
+
+        # root -> owner id
+        exec_as_uid $owneruid {
+            # Create temporary Portfile.bump.XXXXXX
+            if {[catch {set tmpfile [mkstemp "${portpath}/Portfile.bump.XXXXXX"]} error]} {
+                ui_debug $::errorInfo
+                ui_error "mkstemp: $error"
+                return -code error "mkstemp failed"
+            }
+
+            # Extract the Tcl Channel number
+            set tmpfd [lindex $tmpfile 0]
+
+            # Set tmpfile to only the file name
+            set tmpfile [join [lrange $tmpfile 1 end]]
+
+            # Get Portfile attributes
+            set attributes [file attributes $portfile]
+
+            # Direct sed command output to tempfile
+            lappend cmdline "<${portfile}" ">@$tmpfd"
+
+            # Run sed command and write to Portfile.bump
+            ui_info "$UI_PREFIX [format [msgcat::mc "Patching %s: %s"] $portfile $patterns]"
+            if {[catch {exec -ignorestderr -- {*}$cmdline} error]} {
+                ui_debug $::errorInfo
+                ui_error "sed: $error"
+                file delete "$tmpfile"
+                close $tmpfd
+                return -code error "sed sed(1) failed"
+            }
+
+            if {[tbool ports_bump_patch]} {
+                # Patch mode
+                # Set Potfile.patch path
+                set patchfile "${portpath}/Portfile.patch"
+                set patchfd [open $patchfile w]
+
+                # Construct diff command
+                set diffcmd {}
+                lappend diffcmd $portutil::autoconf::diff_path -u --label old/Portfile --label new/Portfile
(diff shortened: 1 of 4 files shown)


satraul:
I did a "draft pull request" which is a new github feature. Hope it's okay.

For context: the `port bump` action would bump Portfile version and checksums as requested here: https://trac.macports.org/ticket/53851
Something like this:
```
$ port bump
--->  Fetching distfiles for chromedriver
--->  Verifying checksums for chromedriver
Old sha256: 00b6ac50b0de0bd63d260cc7472c6e6af44b1240
New sha256: 2aa256d17e4b2cc21f888b0e1b9ed211b925bf40
Old size: -
New size: 7057233
Do you approve of these changes [Y/n]? Y
Do you want to reset revision number (currently 2) to 0 [Y/n]? Y
Version successfully updated. Suggested commit message:
    chromedriver: update to 6.6.3
```

mojca:
Thanks for making us (or at least me) aware of the new "Draft" feature in MacPorts.

I didn't test this yet, but `sudo port` runs under `macports` user with no permissions to write to your home directory (for a good reason). I assume that `port bump` running as "you" might have a better chance to work in this scenario.

src/port1.0/portbump.tcl
```diff
@@ -0,0 +1,234 @@
...
+        # Show the desired checksum line for easy cut-paste
+        # based on the previously calculated values, plus our default types
+        set sums {}
+
+        foreach distfile $all_dist_files {
+            if {[llength $all_dist_files] > 1} {
+                lappend sums $distfile
+            }
+
+            set missing_types $portchecksum::default_checksum_types
```
mojca: I'm not too skilled in tcl, but this line failed for me. What worked was to keep just `$default_checksum_types` (remove `portchecksum`), but I had to add that to `global` variables on top. I don't know what the proper solution is though.

satraul:
> Just a super short comment as I'm not able to be of much help, and I'm stuck with the same permission error, even if I use the `Portfile` owned by macports. Let's hope for some skilled developers to respond quickly.

About permissions: Realised I haven't pushed my changes so I did just now. I created a solution by adapting some code from `portinstall.tcl` to raise privileges like this:

```tcl
# start gsoc08-privileges
if {![file writable $portpath] || ([getuid] == 0 && [geteuid] != 0)} {
    # if Portfile is not writable, need root privileges
    elevateToRoot "bump"
}
# end gsoc08-privileges
```

So reinplace already works on my machine (yay). There could be a better solution tho, like running as "me" as you said.

>The idea with checksum reinplace sounds really nice. I'm not sure how reliable this would be with size that can accidentally match some other number in the Portfile, or might initially be missing completely. 

That's right, although I think the chance is really low. A solution: we could use an exact regex matcher, something like `(md5\[ \t\n\\]*)\d`.

src/port1.0/portbump.tcl
satraul: I think it's the branch yeah, as it runs fine on my machine. I'll check it out later when the implementation is done. Can I resolve this for now?

raimue:
Hey, this looks like a nice start for a first contribution! I see you already made yourself familiar with base.

I am not sure if the `port1.0` is the right place for this functionality. The "targets" implemented in `port1.0` are originally meant to resemble the steps needed to produce a binary port archive. I am aware this was already weakened by other functionality that was added as a port target that should rather have been somewhere else...

In this case, implementing inside `port1.0` seems to make things more complicated. The code for bumping also does not interact with the statefile in any way, so why could it not just be inside the `macports1.0` API layer? Especially, this does not yet address updating the version as described in the original ticket, but I doubt this will work as a target inside `port1.0`. I mean, later you will likely want to run the livecheck target on the existing Portfile, then update the version, and after that parse the Portfile again to run the fetch target and so on. I don't see how this would be possible within a target in `port1.0`.

Elevating to root is the wrong choice to solve the permission issue. I think you should elevate to the user owning the ports tree (or the Portfile?) you are working on, similar to what `port sync` would do. Especially, you only need to elevate to write to the Portfile or other actions that require privileges, but in the general case from the start.

Even better would be to write to a temporary file first and only replace the Portfile atomically once you are done with a `file move`. That ensures you will either get a successfully bumped Portfile, or no change at all in case anything fails in the process. Maybe there is also a use case to produce a patch with some flag instead of overwriting the Portfile?

satraul:
Hi raimue, thanks for the feedback.

> I am not sure if the `port1.0` is the right place for this functionality. The "targets" implemented in `port1.0` are originally meant to resemble the steps needed to produce a binary port archive. I am aware this was already weakened by other functionality that was added as a port target that should rather have been somewhere else...

Ah okay, I didn't know about this. I had a misunderstanding that all actions are supposed to be in `ports1.0`. Based on what you said, that means that code in the `macports1.0` API can run targets such as `fetch` and `livecheck`, correct? Are there any examples of this? I will migrate the code ASAP after I finish the implementation.

> Elevating to root is the wrong choice to solve the permission issue. I think you should elevate to the user owning the ports tree (or the Portfile?) you are working on, similar to what `port sync` would do. Especially, you only need to elevate to write to the Portfile or other actions that require privileges, but in the general case from the start.

I have read the `port sync` code. From what I gather, it changes the `uid` from `root` -> `user` (file owner), modifies a file, then back again `user` -> `root`. Since the current code runs as the `macports` user, is it ok to elevate the privileges like this: `macports` -> `root` -> `user`, modify file, and back again? Or is it possible to run as `user` from start? For now, I've removed the elevation to root.

> Even better would be to write to a temporary file first and only replace the Portfile atomically once you are done with a `file move`. That ensures you will either get a successfully bumped Portfile, or no change at all in case anything fails in the process.

I commited changes based on your advice. Now all replacements are written to a temporary file first. What's left now is the `file move`, but I want to clarify on the privileges first.

>Maybe there is also a use case to produce a patch with some flag instead of overwriting the Portfile?

Sure, added to the backlog. It's doable manually now:
```
$ sudo diff -U 0 Portfile work/.tmp/Portfile.bump
--- Portfile	2019-04-04 16:28:46.000000000 +0700
+++ work/.tmp/Portfile.bump	2019-04-04 16:40:48.000000000 +0700
@@ -8 +8 @@
-revision            2
+revision            0
@@ -26,2 +26,2 @@
-checksums           md5 e509449e52142757c2c75af124847941 \
-                    rmd160 02e628f018d075cc72ff5c2bf8fb0989e2dc63cc
+checksums           md5 c06483349e496501248f5a4dc9193184 \
+                    rmd160 f4c103e24b313744633a20e55365fdd126ac980d
```

satraul:
**Progress update**. `port bump` and `port bump --patch` are now functional.

For the elevation problem: `exec_as_uid` proc is used so all file operations are done as the owner of the Portfile. `[file attributes $portfile $attributes]` are also used to restore the original attributes.

I've also implemented a `--patch` use case thus a `Portfile.patch` is created using `$portutil::autoconf::diff_path` instead of overwriting the original Portfile. Here's a 1 minute demonstration: https://asciinema.org/a/1X9VFUxRXncOE1m7PhB9DVKcX

The next to do list:
- Interactive prompts. Since `macports::ui_options(questions_yesno)` requires `macports1.0` it's a bit hard to implement. It's one of the drawbacks of implementing this in `ports1.0` as raimue said. The alternative is using good old `ui_warn` and `gets`.
- Version argument. An idea of mine is to support `@version` in port fetch so we can do `port fetch @version`, which can fetch an old version (exists in MacPorts repo) or new version (not yet in MacPorts). This in turn will support `port bump @version` since it runs `fetch`. There's [demand for this feature](https://stackoverflow.com/questions/5235215/specify-which-version-to-install-with-macports) as the [current way to do it is bit convoluted](https://trac.macports.org/wiki/howto/InstallingOlderPort). I will add this to my GSOC proposal.

src/port1.0/portbump.tcl
```diff
@@ -280,32 +282,71 @@ proc portbump::bump_main {args} {
...
-                ui_error "bump: $error"
-                file delete "$tmpfile"
-                return -code error "bump overwrite failed"
-            }
+                # Set Potfile.patch path
+                set patchfile "${tempdir}/Portfile.patch"
+
+                # Construct diff command
+                set diffcmd {}
+                lappend diffcmd $portutil::autoconf::diff_path -U 0
```
raimue: Why `diff -U 0`? That is not really common... I would use just `diff -u`.

Furthermore, you could use something like `--label a/Portfile --label b/Portfile` to hide any temporary file names.

src/port1.0/portbump.tcl
```diff
@@ -280,32 +282,71 @@ proc portbump::bump_main {args} {
...
+                    file attributes $portfile {*}$attributes
+
+                    ui_msg "Portfile.patch successfully created at $patchfile"    
+                } else {
+                    ui_error "diff: diff returned 0 exit code (no differences found)"
+                    file delete "$tmpfile"
+                    file delete "$patchfile"
+                    close $tmpfd
+                    close $patchfd
+                    return -code error "diff failed"
```
raimue: Is that really supposed to be an error? I mean, `port bump` would also just leave the Portfile "unchanged", wouldn't it?

src/port1.0/portbump.tcl
```diff
@@ -43,11 +43,12 @@ namespace eval portbump {
 }
 
 # Options
-options version use_livecheck
+options patch use_livecheck
```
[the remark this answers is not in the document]
satraul: Right, I thought it was related to target options. Will remove

src/port1.0/portbump.tcl
satraul: I was thinking to shorten the lines. Yeah it might confuse people, I'll change it. Will add labels

src/port1.0/portbump.tcl
satraul: Right, now returns 0

src/port1.0/portbump.tcl
```diff
@@ -256,27 +237,65 @@ proc portbump::bump_main {args} {
...
+        # root -> owner id
+        exec_as_uid $owneruid {
+            # Create temporary file Portfile.bump at Portfile directory
+            if {[file isdirectory ${portpath}]} {
+                set tempdir ${portpath}
+            } else {
+                set tempdir /tmp
+            }
+            
+            if {[catch {set tmpfile [mkstemp "${tempdir}/Portfile.bump"]} error]} {
```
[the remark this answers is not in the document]
satraul: Right, I didn't know how mkstemp worked before. Changed to `Portfile.bump.XXXXXX`

src/port1.0/portbump.tcl
```diff
@@ -256,27 +237,65 @@ proc portbump::bump_main {args} {
...
+            # Check Portfile overwrite permissios
+            if {[catch {file attributes $portfile -permissions u+w} error]} {
+                ui_debug $::errorInfo
+                ui_error "bump: $error"
+                file delete "$tmpfile"
+                return -code error "bump overwrite permissions failed"
+            }
+
+            # Replace Portfile with Portfile.bump
+            if {[catch {file copy -force $tmpfile $portfile} error]} {
```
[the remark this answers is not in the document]
satraul: Done

src/port1.0/portbump.tcl
```diff
@@ -256,27 +237,65 @@ proc portbump::bump_main {args} {
...
 
-        # TODO: Replace Portfile with Portfile.bump
+        # Get the uid of Portfile owner
+        set owneruid [name_to_uid [file attributes ${portfile} -owner]]
 
-        file delete "$tmpfile"
+        # root -> owner id
+        exec_as_uid $owneruid {
+            # Create temporary file Portfile.bump at Portfile directory
+            if {[file isdirectory ${portpath}]} {
```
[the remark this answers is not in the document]
satraul: Lol hahahah, thanks for the catch. Changed

src/port1.0/portbump.tcl
```diff
@@ -256,27 +237,65 @@ proc portbump::bump_main {args} {
...
+            if {[catch {exec -ignorestderr -- {*}$cmdline} error]} {
+                ui_debug $::errorInfo
+                ui_error "sed: $error"
+                file delete "$tmpfile"
+                close $tmpfd
+                return -code error "sed sed(1) failed"
+            }
+
+            # Check Portfile overwrite permissios
+            if {[catch {file attributes $portfile -permissions u+w} error]} {
```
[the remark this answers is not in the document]
satraul: Done. First time encountering TOCTTOU

satraul:
@raimue I'm a confused about migrating to `macports1.0`. As you said, someday we would like to run the `livecheck` target, update the version, then run the `fetch` target. I don't see examples of "running targets from non-targets" in the code right now, besides actions like `reclaim` running `rev-upgrade`. Both are non-targets.

A solution would be to do something like `target_requires ${org.macports.bump} main pre_bump fetch`, where `pre_bump` is our in-betweener which updates the version and can run livecheck. Sounds suboptimal but it will work.. is there a better way to do this?

Another reason to keep it in `port1.0` is because we're using many helpful utility procs like `exec_as_uid`, `tbool`, and `move` from `port1.0/portutil.tcl`. Also checksum calculation procs from `port1.0/portchecksum.tcl`.

satraul:
And thank you for the code review, I've implemented your suggestions

src/port1.0/portbump.tcl
```diff
@@ -0,0 +1,208 @@
+# et:ts=4
```
neverpanic: The modeline seems incomplete to me. Did you partially copy this line from a different file?

src/port1.0/portbump.tcl
```diff
@@ -0,0 +1,316 @@
...
+
+namespace eval portbump {
+}
+
+# bump_start
+#
+# Target prerun procedure; simply prints a message about what we're doing.
+#
+proc portbump::bump_start {args} {
+    global UI_PREFIX portpath
```
neverpanic: `portpath` seems unused in this function, so it should be removed from the `global` statement.

src/port1.0/portbump.tcl
```diff
@@ -0,0 +1,316 @@
...
+
+    ui_notice "$UI_PREFIX [format [msgcat::mc "Bumping checksums for %s (Portfile --> Distfile)"] [option subport]]"
```
neverpanic: What is the `(Portfile --> Distfile)` output supposed to mean?

src/port1.0/portbump.tcl
```diff
@@ -0,0 +1,316 @@
...
+}
+
+# bump_main
+#
+# Target main procedure. Bumps the checksums for distfiles.
+#
+proc portbump::bump_main {args} {
+    global UI_PREFIX all_dist_files checksums_array portverbose portpath workpath ports_bump_patch
```
neverpanic: `portverbose` and `workpath` seems to be unused.

src/port1.0/portbump.tcl
```diff
@@ -0,0 +1,316 @@
...
+        foreach {type sum calculated_sum} $both_checksums {
+            lappend patterns "s/(${type}${whitespace})${sum}/\\1${calculated_sum}/g"
+        }
+        # Create substitution pattern for revision (reset to 0)
+        lappend patterns {s/(revision[[:space:]\\]+)[0-9]+/\10/g}
+
+        # Construct sed command
+        set cmdline {}
+        lappend cmdline $portutil::autoconf::sed_command -E
+        foreach {pattern} $patterns {
```
neverpanic: Why the quotes around `pattern` here? Those don't seem to be required to me?

src/port1.0/portbump.tcl
```diff
@@ -0,0 +1,316 @@
...
+
+        # root -> owner id
+        exec_as_uid $owneruid {
+            # Create temporary Portfile.bump.XXXXXX
+            if {[catch {set tmpfile [mkstemp "${portpath}/Portfile.bump.XXXXXX"]} error]} {
+                ui_debug $::errorInfo
+                ui_error "mkstemp: $error"
+                return -code error "mkstemp failed"
+            } else {
+                # Extract the Tcl Channel number
```
neverpanic: Since you're erroring out in the if block, there's no need to put this code in an else block.

src/port1.0/portbump.tcl
satraul: I just realised it's confusing, I was trying to explain that we're updating the portfile based on the new distfile. I'll remove it

src/port1.0/portbump.tcl
satraul: Removed

src/port1.0/portbump.tcl
satraul: Removed

src/port1.0/portbump.tcl
```diff
@@ -0,0 +1,208 @@
+# et:ts=4
+# portbump.tcl
+#
+# Copyright (c) 2002 - 2004 Apple Inc.
+# Copyright (c) 2004 - 2005 Paul Guyot <pguyot@kallisys.net>
+# Copyright (c) 2006 - 2012, 2014 - 2016, 2018 The MacPorts Project
```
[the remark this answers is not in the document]
satraul: Right, changed.

src/port1.0/portbump.tcl
satraul: My mistake, I copied the foreach from earlier lines. Removed

src/port1.0/portbump.tcl
satraul: Yes, but it's an exact copy from `portchecksum.tcl`. It's also present in some other files like `portinstall.tcl`. Some files are more whole tho like this:
```suggestion
# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
```
Should we use that instead? These are used in traditional editors like vim right?

src/port1.0/portbump.tcl
satraul: Removed the else

satraul:
@neverpanic Thank you for the code review, I've implemented the suggested changes.

> As for moving this to macports1.0; that could be done, but you'd have to get a handle to the Portfile interpreter and use it to obtain the variables from it, where now you have direct access to those variables. See for example, `proc action_target` in `src/port/port.tcl`. With `mportopen` you get a reference to the Portfile interpreter, and instead of calling `mportexec` then, you can use `_mportkey` and `[ditem_key $mport workername] eval {code_to_be_run_in_portfile_context}` to run code "inside" of a Portfile.

Alright, I'll check this out

src/port1.0/portbump.tcl
jmroot: Yes, the modeline described in [HACKING](https://github.com/macports/macports-base/blob/master/HACKING) should be used.

src/port1.0/portbump.tcl
```diff
@@ -0,0 +1,315 @@
...
+                    set found [lsearch -exact ${missing_types} ${type}];
+                    if { ${found} != -1} {
+                        set missing_types [lreplace ${missing_types} ${found} ${found}]
+                    }
+                }
+            }
+
+            # Append the string for any of our default types that were
+            # not previously calculated
+            if {[llength $missing_types]} {
```
neverpanic: This block currently does nothing. I'd propose you move it to a separate commit on top and open a further PR for this and we can merge this as-is now.

src/port1.0/portbump.tcl
```diff
@@ -0,0 +1,315 @@
...
+            ui_msg [format "Old %-8s %s" ${type}: $sum]
+            ui_msg [format "New %-8s %s" ${type}: $calculated_sum]
+        }
+
+        if {[info exists macports::ui_options(questions_yesno)]} {
+            set retval [$macports::ui_options(questions_yesno) $msg "BumpPrompt" "" {y} 0 "Do you approve of these changes?"]
+            if {$retval == 0} {
+                # User said yes
+                # TODO: Ask to reset revision number only if > 0 and pass revision number to message.
+                set retval [$macports::ui_options(questions_yesno) $msg "BumpPrompt" "" {y} 0 [format "Do you want to reset revision number (currently %s) to 0?" $revision]]
```
neverpanic: Same here, this part contains unimplemented TODOs. If we can move those to a commit on top, I'd merge the changes below.

src/port1.0/portbump.tcl
```diff
@@ -0,0 +1,315 @@
...
+        lappend patterns {s/(revision[[:space:]\\]+)[0-9]+/\10/g}
+
+        # Construct sed command
+        set cmdline {}
+        lappend cmdline $portutil::autoconf::sed_command -E
+        foreach pattern $patterns {
+            lappend cmdline -e $pattern
+        }
+
+        # TODO: Implement patch use case
```
neverpanic: Isn't this already implemented, i.e. shouldn't the TODO be removed?

satraul:
> LG2M now. Can we move the remaining TODOs to a commit on top, move that commit to a separate future PR and squash the other changes into a single commit? Once you've done that, I'd merge this as-is.

Will do, I'll be back with the squashed PR in a few.

src/port1.0/portbump.tcl
satraul: Done, moved the remaining TODOs to separate branches.

PR #121, opened 2019-04-06
portindex.tcl: Add option to generate index of changed ports also

arjunsalyan:
I am trying to devise an efficient way to keep the "ports information" up-to-date in a web-app that would display a dynamic page for each port [[demo](https://frozen-falls-98471.herokuapp.com/ports/SoapyAirspyHF/)]. I am storing data from the port index in a database- which should contain latest information about each port.

For this purpose, I have attempted to add an option to the PortIndex command which would generate a file with only the ports that have changed since last indexing.
The command is working fine till now. But I believe that the code can be improved.

Any suggestions and feedbacks on the code would be really helpful.

### DIFF
--- a/src/port/portindex.tcl
+++ b/src/port/portindex.tcl
@@ -9,6 +9,7 @@ package require Pextlib
 # Globals
 set full_reindex 0
 set permit_error 0
+set index_changes 0
 set stats(total) 0
 set stats(failed) 0
 set stats(skipped) 0
@@ -29,6 +30,7 @@ proc print_usage args {
     puts "-e:\tExit code indicates if ports failed to parse"
     puts "-o:\tOutput all files to specified directory"
     puts "-p:\tPretend to be on another platform"
+    puts "-c:\tAlso generate an index of ports changed since last indexing"
 }
 
 proc _read_index {idx} {
@@ -45,15 +47,19 @@ proc _read_index {idx} {
     return [list $name $len $line]
 }
 
-proc _write_index {name len line} {
-    global fd
+proc _write_index {name len line write_to_changedports_file} {
+    global fd index_changes changesfd
+    if {$index_changes == 1 && $write_to_changedports_file == 1 } {
+        puts $changesfd [list $name $len]
+        puts $changesfd $line
+    }
 
     puts $fd [list $name $len]
     puts $fd $line
 }
 
 proc _write_index_from_portinfo {portinfoname {is_subport no}} {
-    global keepkeys
+    global keepkeys index_changes
 
     upvar $portinfoname portinfo
 
@@ -75,7 +81,11 @@ proc _write_index_from_portinfo {portinfoname {is_subport no}} {
 
     set output [array get keep_portinfo]
     set len [expr {[string length $output] + 1}]
-    _write_index $portinfo(name) $len $output
+    if {$index_changes == 1} {
+        _write_index $portinfo(name) $len $output 1
+    } else {
+        _write_index $portinfo(name) $len $output 0
+    }
 }
 
 proc _open_port {portinfo_name portdir absportdir port_options_name {subport {}}} {
@@ -124,7 +134,7 @@ proc pindex {portdir} {
 
                 # reuse entry if it was made from the same portdir
                 if {[info exists portinfo(portdir)] && $portinfo(portdir) eq $portdir} {
-                    _write_index $name $len $line
+                    _write_index $name $len $line 0
                     incr stats(skipped)
 
                     if {[info exists ui_options(ports_debug)]} {
@@ -136,7 +146,8 @@ proc pindex {portdir} {
                         return
                     }
                     foreach sub $portinfo(subports) {
-                        _write_index {*}[_read_index [string tolower $sub]]
+                        lassign [_read_index [string tolower $sub]] sname slen sline
+                        _write_index $sname $slen $sline 0
                         incr stats(skipped)
                     }
 
@@ -238,6 +249,8 @@ for {set i 0} {$i < $argc} {incr i} {
                 set full_reindex 1
             } elseif {$arg eq "-e"} { # Non-zero exit code on errors
                 set permit_error 1
+            } elseif {$arg eq "-c"} { # Generate one more file along with PortIndex that will contain only changes.
+                set index_changes 1
             } else {
                 puts stderr "Unknown option: $arg"
                 print_usage
@@ -280,6 +293,9 @@ if {[info exists outdir]} {
 
 puts "Creating port index in $outdir"
 set outpath [file join $outdir PortIndex]
+if {$index_changes == 1} {
+    set changesoutpath [file join $outdir ChangedPorts]
+}
 # open old index for comparison
 if {[file isfile $outpath] && [file isfile ${outpath}.quick]} {
     set oldmtime [file mtime $outpath]
@@ -298,6 +314,10 @@ if {[file isfile $outpath] && [file isfile ${outpath}.quick]} {
 
 set tempportindex [mktemp "/tmp/mports.portindex.XXXXXXXX"]
 set fd [open $tempportindex w]
+if {$index_changes == 1} {
+    set tempchangedports [mktemp "/tmp/mports.changedports.XXXXXXX"]
+    set changesfd [open $tempchangedports w]
+}
 set save_prefix ${macports::prefix}
 foreach key {categories depends_fetch depends_extract depends_patch \
              depends_build \
@@ -321,12 +341,18 @@ try {
         close $oldfd
     }
     close $fd
+    if {$index_changes == 1} {
+        close $changesfd
+    }
 }
 if {$exit_fail} {
     exit 1
 }
 
 file rename -force $tempportindex $outpath
+if {$index_changes == 1} {
+    file rename -force $tempchangedports $changesoutpath
+}
 file mtime $outpath $newest
 mports_generate_quickindex $outpath
 puts "\nTotal number of ports parsed:\t$stats(total)\


umeshksingla:
I have gone through the PR, it looks fine at the first look but I do have some comments, mostly about nitpicking. What I do want to know is what is the previous port index file we are taking the diff with (it has nothing to do with your PR but I see it in the original code as well, I will also be learning about the codebase along with you :) ). Do we have a cron job running somewhere? ( @mojca )

Also, I don't know if it's okay to expose the diff option to users as well?

If possible, it would be great if you can link the email in the PR's description above to know why was the above feature required in the first place (or maybe a Trac ticket?).

mojca:
@umeshksingla: after something gets changed in git, we need to update the database in the app. One option is to send the full portindex to the app (probably more than 20 MB file), but it would be much more efficient if the app would only receive the differences since the last import: which ports were added, deleted, or modified.

We have https://build.macports.org/builders/jobs-portindex which generates portindex. We would probably add generation of json on the same buildslave (requested by repology author). I would either run differential portindex on the same server where the app would be running, or make sure that buildbot always submits the differences to the app (after each commit). But it would probably be more reliable and efficient to run this next to the app than on some remote server. It's a pretty trivial process (but needs macports running, of course).

I assume that the previous PortIndex is simply the PortIndex in current directory (I don't know if we have that configurable), but I didn't check.

I don't know what harm a diff option could do to users, but I prefer having one "useless to users" options than implementing super strange workarounds elsewhere and repeat 90% of the code written here to get the same effect. Users should not be using `-diff` for generating the regular portindex, that's for sure. And the diff option should not overwrite the default PortIndex.

If you could give it some testing etc., that would be great. If something is not yet clear, just ask.

arjunsalyan:
> What I do want to know is what is the previous port index file we are taking the diff with

It checks if a previous PortIndex exists, if yes, then it takes the `last modified` time of that PortIndex and all the ports which have been updated after that time are added to the `ChangedPorts` file.

> If possible, it would be great if you can link the email in the PR's description above to know why was the above feature required in the first place (or maybe a Trac ticket?).

Thank You, will do.

umeshksingla:
Well, I built the PortIndex for the first time and I realize the pain and time it takes. I don't think the size of the port index file is an issue (in context to storage) but yes, bigger file means more time to parse.

mojca:
@umeshksingla: yes, it takes "forever" to build it for the first time. It's not about its size alone, it's about the fact that one might take several hundred thousand queries to the database to update from a full portindex. Differential output probably won't help you with the first run though ...

umeshksingla:
I have tested the PR by adding and modifying ports, works correctly. Need to handle the deletion cases as well now.

Some side notes:
1. It'd be better to have the diff filename with a timestamp instead of just 'ChangedPorts'. "Changed" should be reported with respect to some reference.
2. Is it possible to have 'U', 'A', 'D' (for updated, added, deleted) in the output file? I know this way we are able to reuse most of the existing code but having U/A/D would help while updating in application's db. 
3. Be sure not to delete the ports in your db when they get deleted in portindex, we'd like to keep the old statistics. Simply mark it as deleted or obsolete in a separate column.

src/port/portindex.tcl
```diff
@@ -29,6 +30,7 @@ proc print_usage args {
     puts "-e:\tExit code indicates if ports failed to parse"
     puts "-o:\tOutput all files to specified directory"
     puts "-p:\tPretend to be on another platform"
+    puts "-c:\tAlso generate an index of ports changed since last indexing"
```
umeshksingla: remove "Also".

arjunsalyan:
> I have tested the PR by adding and modifying ports, works correctly. Need to handle the deletion cases as well now.

Yes, I am working on deletions now. However, the approach I could think of would double the time of execution.
Currently, PortIndex command traverses the entire 'macports-ports' directory and for portfile it checks if there is a previous entry for that port in the old PortIndex.

To handle deletions, we would have to do exactly the opposite. We would have to open the old PortIndex and for each entry we will search the directory if the ports exist. This is doable. To me it seems like the only way of detecting deletions.

> 1. It'd be better to have the diff filename with a timestamp instead of just 'ChangedPorts'. "Changed" should be reported with respect to some reference.

That would actually be helpful and would remove the possibility of skipping any update. Thank You

> 2. Is it possible to have 'U', 'A', 'D' (for updated, added, deleted) in the output file? I know this way we are able to reuse most of the existing code but having U/A/D would help while updating in application's db.

Yes, it is possible. Will add it with the next update. To make it consistent with `portindex2json.tcl`, I will add it as a key to the `portinfo` array. Something like `status: "U"`

> 3. Be sure not to delete the ports in your db when they get deleted in portindex, we'd like to keep the old statistics. Simply mark it as deleted or obsolete in a separate column.

Okay! I shall update the PR if the 'detecting deletions' method is fine.

src/port/portindex.tcl
```diff
@@ -45,15 +47,19 @@ proc _read_index {idx} {
     return [list $name $len $line]
 }
 
-proc _write_index {name len line} {
-    global fd
+proc _write_index {name len line write_to_changedports_file} {
+    global fd index_changes changesfd
+    if {$index_changes == 1 && $write_to_changedports_file == 1 } {
```
umeshksingla: I initially spent some time figuring out if variable `$write_to_changedports_file` is redundant, turns out it's not. :)

src/port/portindex.tcl
```diff
@@ -238,6 +249,8 @@ for {set i 0} {$i < $argc} {incr i} {
                 set full_reindex 1
             } elseif {$arg eq "-e"} { # Non-zero exit code on errors
                 set permit_error 1
+            } elseif {$arg eq "-c"} { # Generate one more file along with PortIndex that will contain only changes.
```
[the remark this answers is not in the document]
arjunsalyan: I have updated the PR with diff. It definitely sounds better.

src/port/portindex.tcl
arjunsalyan: Yes, it tells the `proc _write_index` to write to the `differences` file.

I have renamed it to `write_diff`.

arjunsalyan:
> @umeshksingla: yes, it takes "forever" to build it for the first time.

That is because it has to print "Adding port XXXX" for each port. If we stop printing this- the time is only a few seconds.

arjunsalyan:
I have added the feature to detect deletions also. I have tested it with many possibilities.

jmroot:
On 2019-4-20 07:43 , Arjun Salyan wrote:
>     @umeshksingla <https://github.com/umeshksingla>: yes, it takes
>     "forever" to build it for the first time.
> 
> That is because it has to print "Adding port XXXX" for each port. If we
> stop printing this- the time is only a few seconds.

This seemed unlikely to me, so I did the experiment. I commented out the
lines 'puts "Adding port $portdir"' and 'puts "Adding subport $sub"' and
timed generating a new PortIndex. This was the result:

portindex  1072.62s user 503.53s system 93% cpu 28:12.59 total

arjunsalyan:
> This seemed unlikely to me, so I did the experiment. I commented out the lines 'puts "Adding port $portdir"' and 'puts "Adding subport $sub"' and timed generating a new PortIndex. This was the result: portindex 1072.62s user 503.53s system 93% cpu 28:12.59 total

Yes, I just thought that since generating the PortIndex in second (or further) run takes quite less time, even though `$fd` is rewritten, so the printing of thousand of entries might be taking up the time. But then I tried and seeing the results I had taken back my comment. It seems that the time consuming step here is picking data from portfiles, but in future runs it is not needed- we just reuse data from old PortIndex.
Sorry, I should have tried first.

umeshksingla:
@mojca just asking to avoid going through the code, is there any relation between the name of a port and the name of the directory containing that port's portfile in macports-ports?

mojca:
@umeshksingla: port name and directory name should match by our policy (I'm not saying we have one officially written, but I hope this is the case for all of our ports), but of course you have subports (when this is definitely not the case).

I don't know if this is also required by the code (you may always expect user mistakes), but I faintly remember experiencing somewhat strange behaviour with portindex when the portname didn't match. Probably the portindex keeping parsing the same file.

jmroot:
The portindex script does use the portdir to check whether Portfiles have been indexed already and skip them if so, yes.

src/port/portindex.tcl
```diff
@@ -321,12 +382,18 @@ try {
...
+        close $diff_fd
+    }
 }
 if {$exit_fail} {
     exit 1
 }
 
 file rename -force $tempportindex $outpath
+if {$index_diff == 1} {
+    file rename -force $tempchangedports $diffoutpath
```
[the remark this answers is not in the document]
arjunsalyan: Currently what I could achieve is use the 'last modified' time of the previous portindex in the name of the file (telling us that this contains the changes since the last indexing). It is in UNIX time. Is it fine?

But, it would create multiple files with names `ChangedPorts-{timestamp}`. And since we are using timestamp of previous portindex in the name, I could not find a way to pick these files and clear them up (except for deleting all files that start with 'ChangedPorts')

src/port/portindex.tcl
```diff
@@ -107,6 +129,24 @@ proc _open_port {portinfo_name portdir absportdir port_options_name {subport {}}
     set portinfo(portdir) $portdir
 }
 
+proc write_deleted_ports {} {
+    global qindex list_of_parsed_ports
+    foreach port [array names qindex] {
+        if {[lsearch -exact $list_of_parsed_ports $port] == -1 && [info exists qindex($port)] && [string trim $port] != ""} {
```
umeshksingla: I doubt if `[info exists qindex($port)]` condition is necessary in `if` above. You are looping over qindex already.

src/port/portindex.tcl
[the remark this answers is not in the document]
umeshksingla: Not sure I understand. Why do we need to delete any file? We are certainly not reading any old diff file while creating the diff file. Even if a folder has multiple diff files, it's okay.

umeshksingla:
Since you worked on capturing diff, just wanted to tell you that it's actually a standard problem in the industry dealing with a large amount of data. They call it "change data capture". I tried looking for a good read but only got [this](https://www.flydata.com/blog/what-change-data-capture-cdc-is-and-why-its-important/) link. You can read it if you are curious, can give you a slight hint on log-based transactions and triggers.

src/port/portindex.tcl
jmroot: Using a linear search here means a quadratic number of operations for the loop. That shouldn't be hard to avoid.

umeshksingla: Right.

arjunsalyan: Yes, it is not needed. Thank you.

arjunsalyan:
> Since you worked on capturing diff, I just wanted to say that it's actually a standard problem in the industry dealing with a large amount of data. They call it "change data capture". I tried looking for a good read but only got [this](https://www.flydata.com/blog/what-change-data-capture-cdc-is-and-why-its-important/) link. You can read it if you are curious, can give you a slight hint on log-based transactions and triggers.

Thank You Umesh. This work on `portindex` is one of the most interesting problems of the project so far. I shall read more about it.

src/port/portindex.tcl
arjunsalyan: Okay, thank you.
Any views on using `lintersect3`?

> Call the first argument to intersect3 A and the second B. Then the first of the three results is A - B, the second result is A intersect B, and the third reuslt is B - A.

umeshksingla: No, you can simply use a map instead of maintaining a list of parsed ports where lookup time is O(1).

arjunsalyan: Please see if I got your point correctly. Thanks for bearing with me.

PR #122, opened 2019-04-09
tracelib.c: Removed unnecessary condition in if

MihirLuthra:
if ((kev.flags & EV_ERROR) == 0 || ((kev.flags & EV_ERROR) > 0 && kev.data != 0))
If compilers find first operand true while evaluating || , the second operand isn’t checked.
So the second operand is only checked if (kev.flags & EV_ERROR) != 0.
So (kev.flags & EV_ERROR) > 0 it doesn’t need to be checked in conjunction with kev.data != 0.
It doesn’t return negative as its of type uint16_t.


jmroot:
Thanks.

MihirLuthra:
> Thanks.

Glad to contribute! ^_^
Also sorry for the delay, before I got engaged in preparing for the proposal and the previous PR I rebased before which didn’t lemme continue with that PR.

PR #124, opened 2019-05-01
Logically reorder output of port selfupdate

saagarjha:
This splits up a conditional to make version-related output come before updating the ports tree. Notably, this means that output that used to look like this:

```console
# port selfupdate
---> Updating MacPorts base sources using rsync
MacPorts base version 2.5.4 installed,
MacPorts base version 2.5.4 downloaded.
---> Updating the ports tree
---> MacPorts base is already the latest version
```

now should look like this:

```console
# port selfupdate
---> Updating MacPorts base sources using rsync
MacPorts base version 2.5.4 installed,
MacPorts base version 2.5.4 downloaded.
---> MacPorts base is already the latest version
---> Updating the ports tree
```

Closes: https://trac.macports.org/ticket/57950

### DIFF
--- a/src/port/port.tcl
+++ b/src/port/port.tcl
@@ -2438,6 +2438,9 @@ proc action_selfupdate { action portlist opts } {
     }
 
     if {$base_updated} {
+        ui_msg "Rerunning selfupdate with updated version of MacPorts\n"
+        execl $::argv0 $::argv
+        
         # exit immediately if in batch/shell mode
         return -999
     } else {
--- a/src/macports1.0/selfupdate.tcl
+++ b/src/macports1.0/selfupdate.tcl
@@ -153,19 +153,11 @@ proc selfupdate::main {{options {}} {updatestatusvar {}}} {
     # check if we we need to rebuild base
     set comp [vercmp $macports_version_new $macports_version]
 
-    # syncing ports tree.
-    if {![dict exists $options ports_selfupdate_no-sync] || ![dict get $options ports_selfupdate_no-sync]} {
-        set syncoptions $options
-        if {$comp > 0} {
-            # updated portfiles potentially need new base to parse - tell sync to try to
-            # use prefabricated PortIndex files and signal if it couldn't
-            dict set syncoptions no_reindex 1
-            dict set syncoptions needed_portindex_var needed_portindex
-        }
-        try {
-            mportsync $syncoptions
-        } on error {eMessage} {
-            error "Couldn't sync the ports tree: $eMessage"
+    if {!$use_the_force_luke} {
+        if {$comp < 0} {
+            ui_msg "$ui_prefix MacPorts base is probably master or a release candidate"
+        } elseif {$comp == 0} {
+            ui_msg "$ui_prefix MacPorts base is already the latest version"
         }
     }
 
@@ -240,7 +232,7 @@ proc selfupdate::main {{options {}} {updatestatusvar {}}} {
             }
 
             # do the actual configure, build and installation of new base
-            ui_msg "Installing new MacPorts release in $prefix as ${owner}:${group}; permissions ${perms}\n"
+            ui_msg "Installing new MacPorts release in $prefix as ${owner}:${group}; permissions ${perms}"
             try {
                 system -W $mp_source_path "${cc_arg}${sdk_arg}./configure $configure_args && ${sdk_arg}make -j${jobs} SELFUPDATING=1 && make install SELFUPDATING=1"
             } on error {eMessage} {
@@ -250,10 +242,6 @@ proc selfupdate::main {{options {}} {updatestatusvar {}}} {
                 set updatestatus yes
             }
         }
-    } elseif {$comp < 0} {
-        ui_msg "$ui_prefix MacPorts base is probably master or a release candidate"
-    } else {
-        ui_msg "$ui_prefix MacPorts base is already the latest version"
     }
 
     # set the MacPorts sources to the right owner
@@ -265,13 +253,22 @@ proc selfupdate::main {{options {}} {updatestatusvar {}}} {
         error "Couldn't change permissions of the MacPorts sources at $mp_source_path to ${sources_owner}: $eMessage"
     }
 
+    ui_msg "\nThe ports tree has been updated. To upgrade your installed ports, you should run"
+    ui_msg "  port upgrade outdated"
+
+    # syncing ports tree.
     if {![dict exists $options ports_selfupdate_no-sync] || ![dict get $options ports_selfupdate_no-sync]} {
-        if {[info exists needed_portindex]} {
-            ui_msg "Not all sources could be fully synced using the old version of MacPorts."
-            ui_msg "Please run selfupdate again now that MacPorts base has been updated."
-        } else {
-            ui_msg "\nThe ports tree has been updated. To upgrade your installed ports, you should run"
-            ui_msg "  port upgrade outdated"
+        set syncoptions $options
+        if {$comp > 0} {
+            # updated portfiles potentially need new base to parse - tell sync to try to
+            # use prefabricated PortIndex files and signal if it couldn't
+            dict set syncoptions no_reindex 1
+            dict set syncoptions needed_portindex_var needed_portindex
+        }
+        try {
+            mportsync $syncoptions
+        } on error {eMessage} {
+            error "Couldn't sync the ports tree: $eMessage"
         }
     }
 


jmroot:
But if base is outdated, the message indicating that is still printed after updating the ports tree? Isn't that less consistent?

saagarjha:
That's a good point. Would moving the entire block doing the actual updating up to be above the sync of the ports tree work, or do those actions have to be performed in the order they are currently (since it seems like there's a special case to handle Portfiles that need an update, which is somewhat strange)?

jmroot:
It's not safe to do anything but exit after updating base. So yeah, the operations happen in the order they do for good reasons.

saagarjha:
Could we run the sync of the ports tree before the MacPorts updating logic, so that the output would look like this:

```console
# port selfupdate
---> Updating the ports tree
---> Updating MacPorts base sources using rsync
MacPorts base version 2.5.4 installed,
MacPorts base version 2.5.4 downloaded.
---> MacPorts base is outdated, installing new version 2.5.4
```

I'm essentially trying to keep the base stuff together.

jmroot:
Not really, as we don't know whether it's safe to run portindex until we know if base is about to be updated.

saagarjha:
Hmm, sounds like a real Catch-22. Why can't we update the ports tree first? If the user runs `port sync` wouldn't that be conceptually equivalent to this, since we wouldn't know about base being updated?

jmroot:
We don't sync first because of this: https://trac.macports.org/ticket/30739

Yes the same would happen if the user just runs `port sync` when base is outdated, but selfupdate can avoid that, which is why we always recommend using selfupdate.

saagarjha:
Thanks for the explanation. Could we keep the order but improve the messages to be a bit more clear as to what's going on? Here's an idea (phrasing is bikesheddable):

```console
# port selfupdate
---> Updating MacPorts base sources using rsync
MacPorts base version 2.5.4 installed,
MacPorts base version 2.5.4 downloaded.
---> MacPorts base is outdated, running prerequisites
---> Updating the ports tree
---> Installing MacPorts base version 2.5.4
```

jmroot:
Sure, if we can adjust the messages to reduce confusion that would be good. The phrasing could use some workshopping as you say.

Another solution that would require more work would be to delay the sync until after base is updated, which would require spawning a new process to run the new base version. That would also allow us to avoid having to tell the user to sync again after a base update when there is no pre-generated PortIndex available for a source.

neverpanic:
Note that code to achieve this has been written already and is currently living on the gsoc17-migrate branch. See

  https://github.com/macports/macports-base/blob/gsoc17-migrate/src/macports1.0/migrate.tcl#L72-L74
  https://github.com/macports/macports-base/blob/gsoc17-migrate/src/port/port.tcl#L2790-L2794

on how to achieve this.

saagarjha:
So we could just check if the base is outdated, run an update, then just `execl` `port sync` immediately afterwards to complete the job? Or is there anything else we need to do?

neverpanic:
Due to the way MacPorts base is structured you should implement this in a very similar way to what I've linked – theoretically, `macports1.0` is the "macports base functionality library" and everything in the `port` folder is a "port client". The API is designed so that there could be multiple "port clients", e.g. the general purpose command line client in `port/port.tcl`, a GUI client or a special-purpose command line client.

To support this correctly, `proc macports::selfupdate` in `macports1.0` should return a marker value that indicates that the port client in use should restart itself and said port client should then implement the correct procedure to restart itself.

Hence, I'd modify `proc macports::selfupdate` in `macports1.0` to return `999` if base was updated and adjust `proc  action_selfupdate` to run `execl $::argv0 $::argv` if `macports::selfupdate` returned `999`. You should keep in mind that users might have specified additional options that should be preserved when re-executing, for example users that did run `port -d selfupdate` would also expect debug output after the re-exec. You could try walking through `$::argv` to replace `selfupdate` with `sync`, but it's probably simplest to just re-run selfupdate as-is.

saagarjha:
I see a `base_updated` variable that seems to do a similar thing: should I still return 999 as well? As far as I can tell (I'm no Tcl expert, sadly) any client of `macports::selfupdate` should be able to get the result of whether base was updated through this variable, so when I was writing the code to check for this value it seemed like we had two "results" that returned the same thing.

To ensure the right output order (which would require deleting the code where we do some sort of sync with prefabricated Portfiles, with the assumption that the client will handle this now) I was thinking of not dealing with `port sync` at all inside of `selfupdate::main` if base is outdated, bubbling this information up to `action_selfupdate` through `macports::selfupdate` which would run `execl` instead of exiting with -999. I guess this technically alters the "API" of `macports::selfupdate` for the reasons I mentioned above but I'm not sure if there is any way around this, so is this something I need to care about?

neverpanic:
> I see a `base_updated` variable that seems to do a similar thing: should I still return 999 as well? As far as I can tell (I'm no Tcl expert, sadly) any client of `macports::selfupdate` should be able to get the result of whether base was updated through this variable, so when I was writing the code to check for this value it seemed like we had two "results" that returned the same thing.

That's correct – I didn't check the existing code, so if there already is a way to check for this condition, we should probably just use it.

> To ensure the right output order (which would require deleting the code where we do some sort of sync with prefabricated Portfiles, with the assumption that the client will handle this now) I was thinking of not dealing with `port sync` at all inside of `selfupdate::main` if base is outdated, bubbling this information up to `action_selfupdate` through `macports::selfupdate` which would run `execl` instead of exiting with -999. I guess this technically alters the "API" of `macports::selfupdate` for the reasons I mentioned above but I'm not sure if there is any way around this, so is this something I need to care about?

No. Altering the API is fine as long as the port client is adjusted appropriately. There are currently only very few other port clients (living mostly in https://github.com/macports/macports-contrib/) and most of them probably don't call the `macports::selfupdate` API.

saagarjha:
I've rewritten the pull request to take into account the suggestions that were mentioned here. Of note, I removed code dealing with syncing after base was updated from `selfupdate::main`, and `action_selfupdate` now runs the `port sync` step if it detects that base was updated. One hitch I came across is that `execl` seizes control of the process, which means we cannot have any useful code after it–notably, this means that our standard

```
--->  Updating the ports tree

The ports tree has been updated. To upgrade your installed ports, you should run
  port upgrade outdated
```

message can't come after this, so I tweaked the wording slightly and moved it above the call:

```
The ports tree will be updated. To upgrade your installed ports, you should run
  port upgrade outdated

--->  Updating the ports tree
```

Again, I don't really know Tcl or the semantics of selfupdate all that well, so I would highly suggest that someone make sure that I don't accidentally break something or open up vulnerabilities ;)

src/port/port.tcl
```diff
@@ -2734,6 +2734,17 @@ proc action_selfupdate { action portlist opts } {
     }
 
     if {$base_updated} {
+        for {set i 0} {$i < $::argc} {incr i} {
+            if {[lindex $::argv $i] == "selfupdate"} {
```
neverpanic: This will turn a invocation of `sudo port selfupdate --nosync` into `sudo port sync --nosync`, which will fail because it does not recognize the flag. I see multiple options to fix this:

1. Drop all parameters after `selfupdate` and run `sync`.
2. Just re-run the same command again. This will run `selfupdate` twice, but avoid having to put the "you should run port upgrade outdated" message above the sync step.
3. Find out whether `--nosync` was passed and exit immediately instead of running `sync` if it was.

Also note that MacPorts does have an interactive mode which can be started using `sudo port`. If the interactive mode was started, there will be no `selfupdate` in `$::argv` and your replacement will fail. You'd still re-exec but the ports tree wouldn't have been updated. Since we can't tell MacPorts to 'run sync, then start interactive mode' and we've previously exited interactive mode after a selfupdate, I'd say it's fine to just run `port sync` (or, as discussed above, selfupdate again) and then exit.

saagarjha: Hmm…given your concerns, I think the second option is probably the safest and least likely to break down the line if the set of options that `selfupdate` and `sync` take diverge more. I think I'll tweak the message in this case to explain that we're running `selfupdate` again: this at least gives the user a bit of context as to what we're doing, and should only happen infrequently anyways so I don't think it's a huge issue.

saagarjha: @neverpanic It's been a while since you looked at this last but would you mind doing another pass?

jmroot:
I believe a variation of this was included in the http selfupdate changes that are now in master. Are any further changes needed?

saagarjha:
Nope, that seems like exactly what I was trying to do here. Closing this.

(FWIW on my personal branch I also have a change to print more details about what's outdated, but mine lists all the ports instead of the number. Also, since I built MacPorts from source, it also checks for changes upstream and rebuilds the project so it's always up-to-date with master. Not sure if you would be interested in either of those changes.)

PR #125, opened 2019-05-09
port-install(1): remove stray backslash

chrstphrchvz:


neverpanic:
Unfortunately due to how our documentation build system works, we are not rebuilding the generated manpages on users' machines (because they lack the required dependencies) and the regeneration thus needs to be in a separate and newer commit.

I've done that now and will merge as soon as the CI build is finished.

PR #130, opened 2019-06-15
port1.0: Implement use_xcode option

satraul:
To my understanding, MacPorts runs fine using only the Command Line Tools and this has been a problem for quite some time. In this PR, as part of the GSOC program, I have implemented several changes as the first step in providing a smooth experience of using MacPorts without Xcode:

1. In addition to the `port 1.0` PortGroup, maintainers can declare Xcode dependency using a new Portfile option `use_xcode`
2. Unless ports declare Xcode dependency:
    - Default ‘DEVELOPER_DIR’ to point Command Line Tools. Thus, shim Xcode binaries such as /usr/bin/clang and /usr/bin/gcc will point to Command Line Tools instead of Xcode. This will ensure all Xcode-independent ports have reproducible builds
    - Installing/building with trace mode will hide the Xcode installation, so MacPorts can fail builds for ports that actually need Xcode but are not declared as such. This will help maintainers identify Xcode dependency.
3. Error out in the `fetch` phase when users try to install Xcode-dependent ports without installing Xcode (if `$xcodeversion eq "none"`)

Relevant:
- https://trac.macports.org/ticket/35045
- https://trac.macports.org/ticket/58016
- Prior discussion in https://github.com/satraul/macports-base/pull/11

Thank you.

### DIFF
--- a/src/port1.0/portbuild.tcl
+++ b/src/port1.0/portbuild.tcl
@@ -48,6 +48,7 @@ options build.asroot \
         use_parallel_build
 commands build
 # defaults
+default build.env {[get_default_env]}
 default build.asroot no
 default build.dir {${worksrcpath}}
 default build.cmd {[portbuild::build_getmaketype]}
--- a/src/port1.0/portbuild.tcl
+++ b/src/port1.0/portbuild.tcl
@@ -48,7 +48,6 @@ options build.asroot \
         use_parallel_build
 commands build
 # defaults
-default build.env {[get_default_env]}
 default build.asroot no
 default build.dir {${worksrcpath}}
 default build.cmd {[portbuild::build_getmaketype]}
--- a/src/port1.0/portconfigure.tcl
+++ b/src/port1.0/portconfigure.tcl
@@ -138,7 +138,7 @@ default compiler.fortran_fallback    {[portconfigure::get_fortran_fallback]}
 # define options
 commands configure autoreconf automake autoconf xmkmf
 # defaults
-default configure.env       ""
+default configure.env       {[get_default_env]}
 default configure.pre_args  {--prefix=${prefix}}
 default configure.cmd       ./configure
 default configure.nice      {${buildnicevalue}}
--- a/src/port1.0/portconfigure.tcl
+++ b/src/port1.0/portconfigure.tcl
@@ -515,17 +515,16 @@ proc portconfigure::configure_get_sdkroot {sdk_version} {
 
 # internal function to determine DEVELOPER_DIR according to Xcode dependency
 proc portconfigure::configure_get_developer_dir {} {
-    global use_xcode macosx_version
-    if {[vercmp $macosx_version 10.9] >= 0} {
-        set cltpath "/Library/Developer/CommandLineTools"
+    global use_xcode developer_dir
+    set cltpath "/Library/Developer/CommandLineTools"
+    # Assume that the existence of libxcselect indiciates the earliest version of
+    # macOS that places CLT in /Library/Developer/CommandLineTools
+    # If port is Xcode-dependent or CommandLineTools directory is invalid, set to developer_dir
+    if {[tbool use_xcode] || ![file exists /usr/lib/libxcselect.dylib] || ![file executable [file join $cltpath usr bin make]]} {
+        return ${developer_dir}
     } else {
-        set cltpath "/"
+        return ${cltpath}
     }
-    # if use_xcode or CLT not installed, do not set DEVELOPER_DIR
-    if {[tbool use_xcode] || (![file isdirectory [file join $cltpath usr include]] || ![file executable [file join $cltpath usr bin make]])} {
-        return ""
-    }
-    return ${cltpath}
 }
 
 # internal function to determine the "-arch xy" flags for the compiler
--- a/src/port1.0/portconfigure.tcl
+++ b/src/port1.0/portconfigure.tcl
@@ -513,13 +513,19 @@ proc portconfigure::configure_get_sdkroot {sdk_version} {
     return -code error "Unable to determine location of a macOS SDK."
 }
 
-# internal function to determine developer_dir according to Xcode dependency
+# internal function to determine DEVELOPER_DIR according to Xcode dependency
 proc portconfigure::configure_get_developer_dir {} {
-    global use_xcode developer_dir
-    if {[tbool use_xcode]} {
+    global use_xcode macosx_version
+    if {[vercmp $macosx_version 10.9] >= 0} {
+        set cltpath "/Library/Developer/CommandLineTools"
+    } else {
+        set cltpath "/"
+    }
+    # if use_xcode or CLT not installed, do not set DEVELOPER_DIR
+    if {[tbool use_xcode] || (![file isdirectory [file join $cltpath usr include]] || ![file executable [file join $cltpath usr bin make]])} {
         return ""
     }
-    return "/Library/Developer/CommandLineTools"
+    return ${cltpath}
 }
 
 # internal function to determine the "-arch xy" flags for the compiler
--- a/src/port1.0/portconfigure.tcl
+++ b/src/port1.0/portconfigure.tcl
@@ -138,7 +138,7 @@ default compiler.fortran_fallback    {[portconfigure::get_fortran_fallback]}
 # define options
 commands configure autoreconf automake autoconf xmkmf
 # defaults
-default configure.env       {[get_default_env]}
+default configure.env       ""
 default configure.pre_args  {--prefix=${prefix}}
 default configure.cmd       ./configure
 default configure.nice      {${buildnicevalue}}
@@ -256,11 +256,13 @@ default configure.pkg_config        {}
 default configure.pkg_config_path   {}
 
 options configure.build_arch configure.ld_archflags \
-        configure.sdk_version configure.sdkroot
+        configure.sdk_version configure.sdkroot \
+        configure.developer_dir
 default configure.build_arch    {[portconfigure::choose_supported_archs ${build_arch}]}
 default configure.ld_archflags  {[portconfigure::configure_get_ld_archflags]}
 default configure.sdk_version   {$macosx_sdk_version}
 default configure.sdkroot       {[portconfigure::configure_get_sdkroot ${configure.sdk_version}]}
+default configure.developer_dir {[portconfigure::configure_get_developer_dir]}
 foreach tool {cc objc f77 f90 fc} {
     options configure.${tool}_archflags
     default configure.${tool}_archflags  "\[portconfigure::configure_get_archflags $tool\]"
@@ -509,6 +511,15 @@ proc portconfigure::configure_get_sdkroot {sdk_version} {
 
     ui_error "Unable to determine location of a macOS SDK."
     return -code error "Unable to determine location of a macOS SDK."
+}
+
+# internal function to determine developer_dir according to Xcode dependency
+proc portconfigure::configure_get_developer_dir {} {
+    global use_xcode developer_dir
+    if {[tbool use_xcode]} {
+        return ""
+    }
+    return "/Library/Developer/CommandLineTools"
 }
 
 # internal function to determine the "-arch xy" flags for the compiler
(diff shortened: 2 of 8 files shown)


src/port1.0/portbuild.tcl
```diff
@@ -48,6 +48,7 @@ options build.asroot \
         use_parallel_build
 commands build
 # defaults
+default build.env {[get_default_env]}
```
jmroot: This is not the right way to add DEVELOPER_DIR to the environment; build.env is to be managed by the Portfile and we don't want thing breaking because the Portfile overwrites the default value.

src/port1.0/portfetch.tcl
```diff
@@ -607,11 +607,20 @@ proc portfetch::fetch_addfilestomap {filemapname} {
     }
 }
 
+# Utility function to error out if Xcode is needed
+proc portfetch::checkxcode {} {
```
jmroot: Why not do this in the existing procedure _check_xcode_version?

src/port1.0/portmain.tcl
```diff
@@ -154,6 +154,14 @@ set egid [getegid]
 default worksymlink {[file normalize [file join $portpath work]]}
 default distpath {[file normalize [file join $portdbpath distfiles ${dist_subdir}]]}
 
+default use_xcode {[portmain::get_default_use_xcode]}
```
jmroot: This could simply be `default use_xcode {[expr {[option build.type] eq "xcode"}]}`

src/port1.0/porttrace.tcl
```diff
@@ -213,18 +213,29 @@ namespace eval porttrace {
...
         # Allow access to developer_dir; however, if it ends with /Contents/Developer, strip
         # that. If it doesn't leave that in place to avoid allowing access to "/"!
         set ddsplit [file split [file normalize [file join ${developer_dir} ".." ".."]]]
         if {[llength $ddsplit] > 2 && [lindex $ddsplit end-1] eq "Contents" && [lindex $ddsplit end] eq "Developer"} {
             set ddsplit [lrange $ddsplit 0 end-2]
         }
-        allow trace_sandbox [file join {*}$ddsplit]
+        lappend xcode_paths [file join {*}$ddsplit]
+
+        if {$use_xcode eq "no"} {
```
jmroot: If use_xcode is a boolean then doing a string comparison is not really correct.

src/port1.0/portutil.tcl
```diff
@@ -508,6 +508,17 @@ proc default_check {optionName index op} {
...
 
+# get_default_env
+# Unless port is Xcode dependent, returns env which sets developer_dir to
+# Command Line Tools
+proc get_default_env {} {
+    global use_xcode
+    if {$use_xcode eq "yes"} {
+        return ""
+    }
+    return "DEVELOPER_DIR=/Library/Developer/CommandLineTools"
```
jmroot: I doubt this works with all the Xcode and OS versions that base runs on.

src/port1.0/porttrace.tcl
satraul: Right, I assume because `tbool` should be instead? Changed

src/port1.0/portmain.tcl
satraul: Changed

src/port1.0/portfetch.tcl
satraul: Right, moved to `_check_xcode_version`

src/port1.0/portutil.tcl
satraul: After reading up, I see that the other possible place for CommandLineTools is `"/"` for `$macosx_version < 10.9`
https://github.com/macports/macports-base/blob/1d6588d178b251c86abc37cbea5ab580e2b85fd1/src/port1.0/portutil.tcl#L3316
I tested it, and it looks like it works to use `DEVELOPER_DIR="/"` for both cases.
```
$ DEVELOPER_DIR="/" clang -v
Apple LLVM version 10.0.0 (clang-1000.10.44.4)
Target: x86_64-apple-darwin17.7.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```
To ensure it's not because xcode-select is selecting Command Line Tools:
```
$ xcode-select -p
/Applications/Xcode.app/Contents/Developer
```
Just to be safe, I'll investigate further.

src/port1.0/portbuild.tcl
satraul: I see. So then it should be to added to the environment in portutil's `command_exec`, or using a separate option like `build.developer_dir` then? I'll read up further.

jmroot: Have a look at how other options that add an environment variable are handled. Examples include `configure.python` and `compiler.cpath`.

src/port1.0/portutil.tcl
jmroot: If you go back far enough, the Command Line Tools are an add-on to Xcode, not an alternative to it. For those Xcode versions we require both to be installed. For new enough Xcode versions, it's viable to have Xcode installed but not the CLT. We also want to support that case.

src/port1.0/portutil.tcl
```diff
@@ -3294,6 +3305,9 @@ proc _check_xcode_version {} {
             }
         }
         if {$xcodeversion eq "none"} {
+            if {[tbool use_xcode]} {
+                return -code error "This port requires Xcode, which was not found on your system."
+            }
             ui_warn "Xcode does not appear to be installed; most ports will likely fail to build."
             if {[file exists "/Applications/Install Xcode.app"]} {
                 ui_warn "You downloaded Xcode from the Mac App Store but didn't install it. Run \"Install Xcode\" in the /Applications folder."
```
jmroot: This message should still be shown in the case where use_xcode is true.

src/port1.0/portutil.tcl
neverpanic: @jmroot Is it actually viable to have only Xcode without the CLTs? It was my understanding that we always require the CLTs because some ports expect `/usr/include` to contain headers. Is that no longer the case? Do we support only having Xcode installed but not the CLTs?

src/port1.0/portutil.tcl
satraul: Moved

src/port1.0/portbuild.tcl
satraul: Thanks for the examples. Changed in ee7e416d

src/port1.0/portutil.tcl
satraul: Is something like this better? Assume CLT is in "/" for < 10.9 and not setting DEVELOPER_DIR if CLT not installed. 
```tcl
if {[vercmp $macosx_version 10.9] >= 0} {
    set cltpath "/Library/Developer/CommandLineTools"
} else {
    set cltpath "/"
}
# If CLT not installed, do not set developer_dir.
if {![file isdirectory [file join $cltpath usr include]] || ![file executable [file join $cltpath usr bin make]]} {
    return ""
}
return ${cltpath}
```
Referenced how to check if CLT not installed from _check_xcode_version proc

jmroot: > It was my understanding that we always require the CLTs because some ports expect /usr/include to contain headers. Is that no longer the case? Do we support only having Xcode installed but not the CLTs?

Requiring /usr/include already isn't viable on 10.14. According to jeremyhu the Xcode-only configuration has been working for quite some time already: https://lists.macports.org/pipermail/macports-dev/2017-July/036045.html

jmroot: > Is something like this better? Assume CLT is in "/" for < 10.9 and not setting DEVELOPER_DIR if CLT not installed.

That does look better. However CLT does not install /usr/include on 10.14+, and I'm not sure what effect DEVELOPER_DIR=/ would have on old Xcode versions where the CLT is reliant on having a full Xcode installation as well. Someone needs to test on old OS versions at some point (@kencu?)

Also, we have an existing developer_dir variable with a default value determined by `macports::set_developer_dir`. It seems like this should be integrated with that somehow, or using the same name would be confusing.

satraul: >Also, we have an existing developer_dir variable with a default value determined by macports::set_developer_dir. It seems like this should be integrated with that somehow, or using the same name would be confusing.

Yeah I guess it's confusing since then we have DEVELOPER_DIR preferring CLT but developer_dir preferring what's selected by `xcode-select`. I don't think it causes any conflicts besides in the [final fallback](https://github.com/macports/macports-base/blob/1d6588d178b251c86abc37cbea5ab580e2b85fd1/src/port1.0/portconfigure.tcl#L1081) of `portconfigure::find_developer_tool` where developer_dir is used.

If we should unify them, we could move the logic of selecting CLT over Xcode to `macports::set_developer_dir`.

If not, we can set DEVELOPER_DIR to developer_dir if CLT not found. But it doesn't actually mattter since if DEVELOPER_DIR isn't set, it looks for `xcode-select` which is what `macports::set_developer_dir` already does.

satraul: Handled macOS < 10.9 and CLT not found in 074a501

src/port1.0/portconfigure.tcl
```diff
@@ -515,17 +515,16 @@ proc portconfigure::configure_get_sdkroot {sdk_version} {
 
 # internal function to determine DEVELOPER_DIR according to Xcode dependency
 proc portconfigure::configure_get_developer_dir {} {
-    global use_xcode macosx_version
-    if {[vercmp $macosx_version 10.9] >= 0} {
```
satraul: As @MarcusCalhoun-Lopez suggested, we should check for the existence of `/usr/lib/libxcselect.dylib` instead of OS version. Presence of libxcselect indiciates the earliest version of macOS that places CLT in `/Library/Developer/CommandLineTools`

satraul: As @MarcusCalhoun-Lopez suggested, check for `/usr/lib/libxcselect.dylib` instead of macOS version. Presence of libxcselect indiciates the earliest version of macOS that places CLT in `/Library/Developer/CommandLineTools`. This needs to be tested.

src/port1.0/portconfigure.tcl
```diff
@@ -515,17 +515,16 @@ proc portconfigure::configure_get_sdkroot {sdk_version} {
...
-        set cltpath "/Library/Developer/CommandLineTools"
+    global use_xcode developer_dir
+    set cltpath "/Library/Developer/CommandLineTools"
+    # Assume that the existence of libxcselect indiciates the earliest version of
+    # macOS that places CLT in /Library/Developer/CommandLineTools
+    # If port is Xcode-dependent or CommandLineTools directory is invalid, set to developer_dir
+    if {[tbool use_xcode] || ![file exists /usr/lib/libxcselect.dylib] || ![file executable [file join $cltpath usr bin make]]} {
```
satraul: Instead of `""`, we explicitly set `DEVELOPER_DIR` to `${developer_dir}` now if `use_xcode` or CLT is invalid. This covers the case that Xcode is installed but CLT isn't. As @jmroot pointed out, the check for `/usr/include` is unreliable so we check for only `/usr/bin/make`.

MarcusCalhoun-Lopez:
This looks good to me as well.

As expected, this change does not seem to have any affect on the compiler behavior for OS < 10.9.

satraul:
Okay, I've rebased and cleaned the commit history

src/port1.0/porttrace.tcl
```diff
@@ -213,18 +213,29 @@ namespace eval porttrace {
...
-        allow trace_sandbox [file join {*}$ddsplit]
+        lappend xcode_paths [file join {*}$ddsplit]
+
+        if {[tbool use_xcode]} {
+            foreach xcode_path $xcode_paths {
+                allow trace_sandbox $xcode_path
+            }
+        } else {
+            foreach xcode_path $xcode_paths {
+                deny trace_sandbox $xcode_path
```
jmroot: Xcode paths need to be allowed here if the system doesn't have the CLT. Not sure if it works or not on pre-xcode-select systems.

satraul: Ok I added in a check for CLT, same as in the one in configure.get_developer_dir

jmroot: Copy/pasted code isn't great, but that should work now.

PR #131, opened 2019-06-26
doc: Add use_lzip documentation

kurthindenburg:
see https://trac.macports.org/ticket/56035

### DIFF
--- a/doc/portfile.7
+++ b/doc/portfile.7
@@ -959,16 +959,6 @@ Sets
 to
 .Dq lzip .
 .br
-Sets
-.Ic extract.pre_args
-to
-.Dq -dc .
-.br
-Sets
-.Ic extract.post_args
-to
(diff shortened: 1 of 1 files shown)


doc/portfile.7
```diff
@@ -946,6 +946,34 @@ to
...
+.br
+Sets
+.Ic extract.pre_args
+to
+.Dq -dc .
+.br
+Sets
+.Ic extract.post_args
+to
+.Dq | tar -xf - .
```
satraul: Isn't this the default anyways for `extract.post_args` without use_zlip?

https://github.com/macports/macports-base/blob/1d6588d178b251c86abc37cbea5ab580e2b85fd1/src/port1.0/portextract.tcl#L58

satraul: Looks good otherwise

PR #132, opened 2019-06-26
Add note that ports with license NoMirror will not be mirrored

kurthindenburg:
see https://trac.macports.org/ticket/56025

### DIFF
--- a/doc/port-mirror.1.txt
+++ b/doc/port-mirror.1.txt
@@ -28,6 +28,9 @@ downloaded files are stored in a custom database in
 man:macports.conf[5]. This database can be deleted and re-created from scratch
 using the *--new* option.
 
+[NOTE]
+Ports that have NoMirror in their license field will not be mirrored.
+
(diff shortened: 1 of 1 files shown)


kurthindenburg:
I"m a bit confused on why port-mirror.1 is generated and also in git repo; let me know if I should commit that file as well.

satraul:
Yeah, the generated port-mirror.1 file should also be committed in a separate commit. Looks good to me otherwise

raimue:
We commit the generated man pages to the repository in order to allow building base without asciidoc. It would be a chicken-and-egg problem to require asciidoc first to build MacPorts on a new platform.

However, we also use the modification date of the latest git commit that touched the `*.1.txt` source file in the generated man page. Therefore both files cannot be committed at the same time, but the process has to be:

1. Edit *.1.txt
2. Commit *.1.txt
3. Regenerate manpages
4. Commit *.1

PR #133, opened 2019-06-30
Remove Xcode uninstalled warning message

satraul:
* ~~Changes message from “most ports” to “some ports”. This has been requested.~~
* ~~Warn only once every two weeks. This solves the complaint that the warning spams the terminal (e.g when doing `sudo port upgrade`).~~
* Remove warning since we already error out on Xcode-dependent ports.

References https://trac.macports.org/ticket/35045
References https://trac.macports.org/ticket/58016

### DIFF
--- a/src/port1.0/portutil.tcl
+++ b/src/port1.0/portutil.tcl
@@ -3300,7 +3300,6 @@ proc _check_xcode_version {} {
             }
         }
         if {$xcodeversion eq "none"} {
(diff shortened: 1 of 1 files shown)


jmroot:
I don't see why we need this warning at all if we're supporting Xcodeless operation.

satraul:
> I don't see why we need this warning at all if we're supporting Xcodeless operation.

I was under the impression that we needed a solution between entirely removing the warning and not. After reading the ticket discussions again, I see that removing it altogether is feasible.

PR #134, opened 2019-07-08
Universally export DEVELOPER_DIR

satraul:
* refactor autoreconf, automake etc to one loop
* export DEVELOPER_DIR for destroot, build, and autotools

Since we're hiding Xcode on trace mode if ports aren't Xcode-dependent, ports that do e.g `post-build { system -W "someotherdir" "make" }` fail because `/usr/bin/make` resolves to Xcode's make. It should resolve to CommandLineTools's make instead like `build.cmd make` already does.

This PR solves that problem by also setting DEVELOPER_DIR for build, destroot and all the autotools.

### DIFF
--- a/src/port1.0/portutil.tcl
+++ b/src/port1.0/portutil.tcl
@@ -425,6 +425,7 @@ proc command_exec {command args} {
     if {[option compiler.library_path] ne ""} {
         set ${varprefix}.env_array(LIBRARY_PATH) [join [option compiler.library_path] :]
     }
+    set ${varprefix}.env_array(DEVELOPER_DIR) [option configure.developer_dir]
 
     # Debug that.
     ui_debug "Environment: [environment_array_to_string ${varprefix}.env_array]"


src/port1.0/portconfigure.tcl
```diff
@@ -1367,27 +1367,23 @@ proc portconfigure::configure_main {args} {
...
-            return -code error "[format [msgcat::mc "%s failure: %s"] automake $result]"
-        }
-    }
-
-    if {[tbool use_autoconf]} {
-        if {[catch {command_exec autoconf} result]} {
-            return -code error "[format [msgcat::mc "%s failure: %s"] autoconf $result]"
+    # Export flags into destroot and build
+    foreach phase {destroot build} {
+        parse_environment ${phase}
```
MarcusCalhoun-Lopez: I could be mistaken, but I believe `parse_environment` is already called [by default](https://github.com/macports/macports-base/blob/6f8aca0267707bd4d8e122148e3f88af57029a73/src/port1.0/portutil.tcl#L415).

src/port1.0/portconfigure.tcl
```diff
@@ -1367,27 +1367,23 @@ proc portconfigure::configure_main {args} {
...
+        append_to_environment_value ${phase} DEVELOPER_DIR {*}[option configure.developer_dir]
+    }
+    # If enabled, run autoreconf/automake/autoconf
+    foreach config {autoreconf automake autoconf} {
+        if {[tbool use_${config}]} {
+            append_to_environment_value ${config} DEVELOPER_DIR {*}[option configure.developer_dir]
```
MarcusCalhoun-Lopez: If the goal is to always add DEVELOPER_DIR to the environment, there is [precedent](https://github.com/macports/macports-base/blob/6f8aca0267707bd4d8e122148e3f88af57029a73/src/port1.0/portutil.tcl#L417) for that sort of thing in `command_exec`. Perhaps we can do something similar.

satraul: Right, that looks much cleaner. I've created the solution which adds the DEVELOPER_DIR in `command_exec` here https://github.com/satraul/macports-base/commit/ad8df129ed674fee3e532fe8b7eb69f5afbc4b6f. It seems both solutions aren't working yet tho.. I'm testing both on emptyport with `build { system -W "/" "clang -v"}` but it's still going to Xcode.

jmroot: If you're running something with `system` (or `exec`) directly, that's completely separate from all the `command` machinery. It's the caller's responsibility to set any required environment variables in that case.

satraul:
I'm compiling a list of before/after (to demonstrate what has changed. Among the ports I'm testing depends on tiff, which is currently failing to install. It says `:info:configure configure: error: Libzstd headers not found at /opt/macports-test/include`. I've tried to port clean & install zstd again but it still doesn't work. Here's the gist: https://gist.github.com/satraul/d327cf63a9b31e2b716262174c4db22f

eirnym:
After this breaking change most ports I require stopped to build. I forcibly build each and every port from source for my personal reason. I had to set default `use_xcode` to `true` to make them compilable. Please, rebuild all ports as a test for such changes

mojca:
Thanks a lot for the feedback.

Can you please just point out a few ports for which you know are broken after the change? Do you use 10.14?

Maybe we should in fact try this, but one of the problems is that such building would probably take one or two weeks, and there are outdated ports that are broken without the change, so we will need a bit of extra work to figure out which ports are newly broken.

eirnym:
I've tried on both `10.14` and `10.15` with both `Xcode 10` and `Xcode 11` beta on both systems (yes, I was sure about command line tools I using). And *exactly* setting the `DEVELOPER_DIR` breaks builds for some ports, not macOs or `Xcode` version. 

I understand that building ALL ports doesn't make sense if you want just to test feature. Most ports I using are more or less recent, language sub dependencies I prefer to install as minimum as possible to be sure that my applications depends on their own versions. Try languages, compilers, shells, popular developer tools first to make the list even more shorten.

The ports tree I using is [here](https://github.com/eirnym/macports-ports/tree/backup). It's a huge pain to request an update through few tracking systems at once, so I prefer to keep updates for myself.

satraul:
Sorry for the inconvenience, we'll try to reproduce the build errors you are facing and try to fix them. For now, we have already tested installing the most popular packages (~250 ports) but not building them from source. We do know though that building e.g gcc5, gcc6, cmake works but building clang-6.0 fails (without Xcode).

Do you have any specific port/logs we can work on? Thanks.

eirnym:
Thank you, you can start with python. Perl could be built, but I haven't tested it yet.

eirnym:
on logs for Python, the same error as it was some day, `generate-posix-vars failed` with segmentation fault during early stage of `build` phase. I don't know the reasons exactly, and I wanted to install my system.

neverpanic:
@eirnym We're tried building the `python3.7` port from source on 10.12 with Xcode 9.2 and the command line tools installed and 10.13 without Xcode (but CLT) and it didn't fail for either of us.

Can you please provide the main.log of a failed build so we can debug this further?

PR #135, opened 2019-07-17
portconfigure: Prioritize returning SDK with DEVELOPER_DIR set 

satraul:
- move xcrun --sdk macosx to find macOS SDK to the top
- export DEVELOPER_DIR so xcrun can find CommandLineTools SDK

Relevant: https://trac.macports.org/ticket/57143

`xcrun --sdk macosx`, which was previously reenabled as the final fallback by @jmroot, is moved to the top of the `::configure_get_sdkroot` proc so the newest-available macOS SDK is returned first instead of the current OS version's SDKs. It was generally agreed on trac that the newest-available SDK should be used, so I assume this change is better and causes no problems.

[previously reenabled]: https://github.com/macports/macports-base/commit/8caac195a15c8a01d07cf337439fd68f4370d9ee

The DEVELOPER_DIR change is done so a CommandLineTools SDK is returned if the port isn't Xcode-dependent. This fixes gmp failing to install in tracemode with the latest changes to master.

### DIFF
--- a/src/port1.0/portconfigure.tcl
+++ b/src/port1.0/portconfigure.tcl
@@ -455,7 +455,7 @@ proc portconfigure::configure_get_ld_archflags {} {
 }
 
 proc portconfigure::configure_get_sdkroot {sdk_version} {
-    global developer_dir macosx_version xcodeversion os.arch os.platform
+    global developer_dir macosx_version xcodeversion os.arch os.platform use_xcode
 
     # This is only relevant for macOS
     if {${os.platform} ne "darwin"} {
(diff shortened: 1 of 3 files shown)


jmroot:
It definitely does cause problems, as indicated in the comment you deleted. Anything that detects functionality at build time will see things available in the SDK that are not available in the OS. This would be fine if everything used potentially weak-linked symbols correctly, but that's actually very rare in the world of software that isn't Mac-only.

satraul:
> It definitely does cause problems, as indicated in the comment you deleted. Anything that detects functionality at build time will see things available in the SDK that are not available in the OS. This would be fine if everything used potentially weak-linked symbols correctly, but that's actually very rare in the world of software that isn't Mac-only.

I don't fully understand weak-links yet but I've just read up about them. It means that we should still ideally return the target's OS version (which is the current OS) over the newest-available right?

I figure the solution would be to try `xcrun --sdk ${sdk_version}` first, then `xcrun --sdk macosx`, then the other checks. We would still be prioritizing DEVELOPER_DIR first (CLT>Xcode) over version (now current>newest-available). Prioritizing version over DEVELOPER_DIR would still cause conflicts with the Xcode tracemode changes.

src/port1.0/portconfigure.tcl
```diff
@@ -472,6 +472,20 @@ proc portconfigure::configure_get_sdkroot {sdk_version} {
...
+    # Prefer finding SDK using xcrun because it supports the DEVELOPER_DIR environment
+    # Thus CommandLinesTools SDK is returned if Xcode is unnecessary.
+    if {![catch {exec env DEVELOPER_DIR=${configure.developer_dir} xcrun --sdk macosx${sdk_version} --show-sdk-path 2> /dev/null} sdk]} {
+        return $sdk
+    }
+
+    # Return the newest-available SDK ("macosx") if it is present.
+    # This was previously the final fallback, but we want to prioritize CLT being used even if it's newer than current OS.
+    # See <https://trac.macports.org/ticket/57143>
+    if {![catch {exec env DEVELOPER_DIR=${configure.developer_dir} xcrun --sdk macosx --show-sdk-path 2> /dev/null} sdk]} {
```
neverpanic: While that's the approach we did discuss, I think you should preserve the information that was documented in the comment that you removed with this change.

We should probably also keep the `ui_warn "Unable to determine location of the macOS ${sdk_version} SDK.  Using the default macOS SDK."` message.

satraul: Added the warnings and message again

jmroot:
Looking in the CLT location first is fine, but xcrun should remain the last resort because it can be quite slow. And yes the target OS version's SDK should be tried first and generic MacOSX after.

The lack of weak linking awareness isn't just theoretical by the way, there have been a number of tickets where users ran into problems because they had Xcode installed (with a later OS version's SDK) but not the CLT. This is one example: https://trac.macports.org/ticket/54893

satraul:
Alright thanks for the feedback, I've made some changes to favor checking with `file` over xcrun. Also refactored `use_xcode` so its set to false if CLT isn't installed. gmp sucessfully installs with the latest changes.

PR #136, opened 2019-07-17
portutil: Export DEVELOPER_DIR for target {} blocks

satraul:
Doesn't work yet. We want to fix `system -W` resolving to Xcode binaries which we now hide in tracemode if the port isn't Xcode-dependent. Ports like libnetpbm, tig, and emacs now fail to install in tracemode because of this.

I've said in the mail that it was around 60 ports that use system -W, but there's actually 357 matches for system -W in macports-ports, which would be a headache to patch since we don't know which ones might resolve to an Xcode tool.

Is this the wrong way to do it? It works with exec, I'm naively assuming it'd be the same for system.
```tcl
% set ::env(DEVELOPER_DIR) /Library/Developer/CommandLineTools/
/Library/Developer/CommandLineTools/
% exec clang -v
Apple LLVM version 10.0.1 (clang-1001.0.46.4)
Target: x86_64-apple-darwin18.5.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```

### DIFF
--- a/src/port1.0/portutil.tcl
+++ b/src/port1.0/portutil.tcl
@@ -1471,6 +1471,8 @@ proc target_run {ditem} {
                     if {[llength $deptypes] > 0} {tracelib setdeps $deplist}
                 }
 
+                # For {} blocks in the Portfile, export DEVELOPER_DIR to prevent Xcode binaries if shouldn't be used
+                set ::env(DEVELOPER_DIR) [option configure.developer_dir]
                 if {$result == 0} {
                     foreach pre [ditem_key $ditem pre] {
                         ui_debug "Executing $pre"
@@ -1500,6 +1502,8 @@ proc target_run {ditem} {
                         if {$result != 0} { break }
                     }
                 }
+                # Keep the environment clean by unsetting DEVELOPER_DIR
+                unset -nocomplain ::env(DEVELOPER_DIR)
 
                 # Check dependencies & file creations outside workpath.
                 if {[tbool ports_trace]


src/port1.0/portconfigure.tcl
```diff
@@ -1372,6 +1372,9 @@ proc portconfigure::configure_main {args} {
         global configure.${flags} configure.universal_${flags}
     }
 
+    # Hack to export DEVELOPER_DIR for all child processes
+    set ::env(DEVELOPER_DIR) [option configure.developer_dir]
```
neverpanic: We might not always run the configure phase (for example if it has been completed in a previous invocation of `port`), so just setting this in configure and assuming it will stay set for all following phases is probably wrong.

satraul: Putting it in main works and I could install libnetpbm in tracemode, but I had to hardcode the path. I can't access the configure.developer_dir variable. Each case, it produces a warning/error.

- hardcode produces warning at runtime in macports::setxcodeinfo
```
set ::env(DEVELOPER_DIR) "/Library/Developer/CommandLineTools/"
...
Warning: xcodebuild exists but failed to execute
```
So maybe it is overkill to put it in main, since we ultimately just want to affect the system -W calls. Which part of the codebase invokes the e.g post-build {} blocks of a Portfile?

- configure.developer_dir errors at runtime
```
...
set ::env(DEVELOPER_DIR) [option configure.developer_dir]
...
Error: Unable to open port: can't read "configure.developer_dir": no such variable
```
The solution would be to option configure.developer_dir and default configure.developer_dir before this code on main, but the naming wouldn't make sense anymore. Or maybe explore modifying system.c instead. What do you think?

jmroot:
This is only for trace mode's benefit right? Is it really worth it if Apple so clearly intends Xcode to be used if present?

In any case, it seems like the right place to inject the env vars would be the same place as the trace mode ones.

satraul:
> This is only for trace mode's benefit right? Is it really worth it if Apple so clearly intends Xcode to be used if present?

Not really, the intended high-level goal was to ensure that CLT gets used if Xcode isn't needed, to ensure more build-reproducibility. The tracemode part is a way to assert that it doesn't get used if needed (for package maintainers).

> In any case, it seems like the right place to inject the env vars would be the same place as the trace mode ones.

If system -W not following this goal is deemed OK to be an exception, that works. Is that alright?

neverpanic:
As discussed, we should move setting the `DEVELOPER_DIR` env variable around the block https://github.com/macports/macports-base/blob/master/src/port1.0/portutil.tcl#L1474-L1502. That should give us the behavior we want, work when builds of ports are interrupted and resumed and should make trace mode happy as well.

jmroot:
> Not really, the intended high-level goal was to ensure that CLT gets used if Xcode isn't needed, to ensure more build-reproducibility. Does this need modification?

Just understand that reproducing a build is only possible if the same tools are used. If we allow a choice of tools at all, then build products won't always be identical.

Or to put it differently, reproducibility is not the same as reproduction.

If you want to reproduce the builds done by our buildbot, you have to install both Xcode and the CLT because that is the buildbot's configuration. Users can choose to install only one or the other, but they can't expect to reproduce the build results in that case.

Favouring the CLT over Xcode doesn't change the level of reproducibility overall. It just makes Xcode-only the less-tested alternative, rather than CLT-only being in that position. That can still be viewed as a win since the CLT is a smaller install.

satraul:
> Favouring the CLT over Xcode doesn't change the level of reproducibility overall. It just makes Xcode-only the less-tested alternative, rather than CLT-only being in that position.

I've just understood this because of your explanation. I've thought about this a bit but I couldn't articulate it since I don't understand too well.

>That can still be viewed as a win since the CLT is a smaller install.

Yes, I guess this is the true goal now, atleast since I realized that Xcode-only should be supported.  It's still inline with my original motivation of undertaking this project which was my past experience as a new macOS user on a new work environment with slow internet that was hindered by the preference of Xcode by package managers.

satraul:
> Can you test a few ports with this to make sure it doesn't break anything?

I've tested the changes again on 6 ports which use system -W. Those which failed to install on tracemode previously now work (tig, emacs, libnetpbm). The others still work as before (python37, gnuplot, ghostscript).

To be clear, I'm not sure if "target {} blocks" is what they're supposed called, but now the changes only affect the those blocks which are e.g `pre-build {...}`, `build {...}`, `post-build {...}` in a Portfile.

PR #140, opened 2019-08-01
macports.tcl: repeat xcode check for dependencies

satraul:
* repeat _check_xcode_version in _mportexec
* add portname in _check_xcode_version error message

Error out earlier if dependency needs Xcode.

Current behavior:
```
$ sudo port install gnuplot
--->  Computing dependencies for gnuplot
The following dependencies will be installed: 
 aquaterm
 libcaca
 libcerf
 lua
 wxWidgets-3.0
 wxWidgets-common
 wxWidgets_select
Continue? [Y/n]: Y
--->  Fetching distfiles for aquaterm
--->  Verifying checksums for aquaterm
--->  Extracting aquaterm
--->  Applying patches to aquaterm
--->  Configuring aquaterm
--->  Building aquaterm
Error: Failed to build aquaterm: command execution failed
Error: See /opt/local/var/macports/logs/_aqua_aquaterm/aquaterm/main.log for details.
Error: Follow https://guide.macports.org/#project.tickets to report a bug.
Error: Processing of port gnuplot failed
```

PR behavior:
```
$ sudo port install gnuplot
--->  Computing dependencies for gnuplot
The following dependencies will be installed: 
 aquaterm
 libcaca
 libcerf
 lua
 wxWidgets-3.0
 wxWidgets-common
 wxWidgets_select
Continue? [Y/n]: Y
Error: Port aquaterm requires the full Xcode installation, which was not found on your system. You can install Xcode from the Mac App Store or https://developer.apple.com/xcode/
Error: See /opt/local/var/macports/logs/_aqua_aquaterm/aquaterm/main.log for details.
Error: Follow https://guide.macports.org/#project.tickets to report a bug.
Error: Processing of port gnuplot failed
```

### DIFF
--- a/src/macports1.0/macports.tcl
+++ b/src/macports1.0/macports.tcl
@@ -2102,6 +2102,7 @@ proc _mportexec {target mport} {
     set workername [ditem_key $mport workername]
     $workername eval {validate_macportsuser}
     if {![catch {$workername eval "check_variants $target"} result] && $result == 0 &&
+        ![catch {$workername eval {_check_xcode_version}} result] && $result == 0 &&
         ![catch {$workername eval {check_supported_archs}} result] && $result == 0 &&
         ![catch {$workername eval "eval_targets $target"} result] && $result == 0} {
         # If auto-clean mode, clean-up after dependency install
--- a/src/port1.0/portutil.tcl
+++ b/src/port1.0/portutil.tcl
@@ -3239,7 +3239,11 @@ proc check_supported_archs {} {
 
 # check if the installed xcode version is new enough
 proc _check_xcode_version {} {
-    global os.subplatform macosx_version xcodeversion use_xcode
+    global os.subplatform macosx_version xcodeversion use_xcode subport
+
+    if {[_archive_available]} {
+        return 0
+    }
 
     if {${os.subplatform} eq "macosx"} {
         switch $macosx_version {
@@ -3309,7 +3313,9 @@ proc _check_xcode_version {} {
                 ui_warn "You downloaded Xcode from the Mac App Store but didn't install it. Run \"Install Xcode\" in the /Applications folder."
             }
             if {[tbool use_xcode]} {
-                return -code error "This port requires the full Xcode installation, which was not found on your system. You can install Xcode from the Mac App Store or https://developer.apple.com/xcode/"
+                ui_error "Port ${subport} requires a full Xcode installation, which was not found on your system."
+                ui_error "You can install Xcode from the Mac App Store or https://developer.apple.com/xcode/"
+                return 1
             }
         } elseif {[vercmp $xcodeversion $min] < 0} {
             ui_error "The installed version of Xcode (${xcodeversion}) is too old to use on the installed OS version. Version $rec or later is recommended on macOS ${macosx_version}."


mojca:
How does this behave when `gnuplot` needs to be installed from source, while `aquaterm` could be fetched in binary form from the mirror?

mojca:
Also, maybe a much much much better fix would be to write CMake files and simply build aquaterm with cmake rather than with XCode :)
I'm of course not asking you to do so, just making a remark. The only thing I'm not sure about is how to properly build the framework, but the number of files is so small that this could be achieved almost trivially.

satraul:
> How does this behave when `gnuplot` needs to be installed from source, while `aquaterm` could be fetched in binary form from the mirror?

How do I go about testing this? I'm not familiar with the build system enough to know how to simulate that `aquaterm` could be fetched in binary form.
https://github.com/macports/macports-base/blob/7ddbc50602b794cbdccdde81ee801b08c699e2b2/src/macports1.0/macports.tcl#L2150-L2168

My logic was to repeat the Xcode check like how the main port already does. `_check_xcode_version` and `check_supported_archs` are side by side here in `mportexec` which runs for the main port, so I just did the same thing for `_mportexec` which runs for dependencies.

mojca:
> > How does this behave when `gnuplot` needs to be installed from source, while `aquaterm` could be fetched in binary form from the mirror?
> 
> How do I go about testing this? I'm not familiar with the build system enough to know how to simulate that `aquaterm` could be fetched in binary form.

That's a question for one of the base experts like @jmroot, @neverpanic, @MarcusCalhoun-Lopez, ...

I'm afraid that our support for that is somewhat limited, and we just build ports as requests come in. For example, I always wanted to be able to fetch everything from web first in the fastest possible time, and then take the laptop on the train and keep building the ports that might take hours to build while I'm offline. I'm not aware that this would be possible.

I once tried installing macports on machine with no xcode and no command-line tools and a surprising number of ports were simply installed successfully from binary archives (actually everything I needed at that time). Throwing an error in case when something might actually succeed is a bit of an overkill, but I agree that current situation is not ideal either. We could probably throw a warning or ask the user to tell whether to manually proceed or not. Not ideal either, but ...

I leave that up to experts.

src/macports1.0/macports.tcl
```diff
@@ -2102,6 +2102,7 @@ proc _mportexec {target mport} {
     set workername [ditem_key $mport workername]
     $workername eval {validate_macportsuser}
     if {![catch {$workername eval "check_variants $target"} result] && $result == 0 &&
+        ![catch {$workername eval {_check_xcode_version}} result] && $result == 0 &&
```
jmroot: If you're going to use _check_xcode_version in this way, it needs to check _archive_available first.

src/port1.0/portutil.tcl
```diff
@@ -3309,7 +3309,8 @@ proc _check_xcode_version {} {
                 ui_warn "You downloaded Xcode from the Mac App Store but didn't install it. Run \"Install Xcode\" in the /Applications folder."
             }
             if {[tbool use_xcode]} {
-                return -code error "This port requires the full Xcode installation, which was not found on your system. You can install Xcode from the Mac App Store or https://developer.apple.com/xcode/"
+                ui_error "Port ${subport} requires the full Xcode installation, which was not found on your system. You can install Xcode from the Mac App Store or https://developer.apple.com/xcode/"
```
jmroot: I think "a full Xcode installation" would sound more natural than "the full Xcode installation".

satraul: Changed

src/macports1.0/macports.tcl
satraul: I've moved the _check_xcode_version call to portfetch now instead of _mportexec, that should work right?

jmroot: No, Xcode is not required to fetch or extract. I'd recommend going back to the previous version of the change, but adding this at the start of _check_xcode_version:
```
if {[_archive_available]} {
    return 0
}
```

satraul: Right, I've done that. When is _archive_available used before?

jmroot: > When is _archive_available used before?

Several places, for example in the calculation of dependencies, since if there's an archive then build-time deps are not needed.

Further on the point that Xcode is not needed to fetch, I think that _target_needs_deps is no longer a sufficient check. We really need another check for whether the target needs a toolchain.

neverpanic: @jmroot Would you suggest adding

```tcl
proc macports::_target_needs_toolchain {target} {
    switch -- $target {
         patch -
         configure -
         build -
         test -
         destroot -
         install -
         activate -
         dmg -
         mdmg -
         pkg -
         mpkg {
              return 1
         }
         default {
              return 0
         }
    }
}
```

instead then and replacing `_target_needs_deps $target` in the beginning of `mportexec`, plus guarding the `_check_xcode_version` check in `_mportexec` with a `if {_target_needs_toolchain $target}` then?

In any case, the current behavior is definitely wrong and needs to be fixed, and this patch is better in requiring Xcode a little too often rather than not requiring it often enough, so I'm going to merge it as-is now. Please let me know whether my proposal outlined above is what you had in mind – I'll follow-up with a patch then.

jmroot: In _mportexec, the target is always activate, so the Xcode check is always needed if an archive is not available. In mportexec, the target may need Xcode even if an archive is available (e.g. build). So there need to be three cases in _target_needs_toolchain: needs toolchain, doesn't need toolchain, and needs toolchain if there is no archive.

So the _archive_available check needs to be moved outside of _check_xcode_version, and the new proc would look more like this:
```tcl
proc macports::_target_needs_toolchain {target} {
    switch -- $target {
         configure -
         build -
         test -
         destroot {
             return yes
         }
         install -
         activate -
         dmg -
         mdmg -
         pkg -
         mpkg {
              return if_no_archive
         }
         default {
              return no
         }
    }
}
```
I excluded patch as well because /usr/bin/patch comes with the base OS. Technically a portfile could do something that needs a toolchain in post-extract etc., but I think it's better to err on the side of allowing those targets to try to run.

neverpanic: Right, that sounds good. I'll try to follow-up with this change soonish.

neverpanic: See #143.

PR #141, opened 2019-08-01
porttrace: allow /var/db/timezone/zoneinfo/

satraul:
```
Warning: The following existing file were hidden from the build system by trace mode:
  /var/db/timezone/zoneinfo/Asia/Jakarta
```
When installing 261 ports with tracemode, this message popped up [232] and [233] times respectively on 10.13 w/o Xcode and 10.14 w/ Xcode.

`/var/root/.CFUserTextEncoding` also popped up, but I don't know if that's as harmless as timezone.

[232]: https://paste.macports.org/8d4a12615e02
[233]: https://paste.macports.org/ba4702689b08

### DIFF
--- a/src/port1.0/porttrace.tcl
+++ b/src/port1.0/porttrace.tcl
@@ -207,6 +207,9 @@ namespace eval porttrace {
             }
         }
 
+        # Allow timezone info
+        allow trace_sandbox "/var/db/timezone/zoneinfo/"
+
         # Allow access to SDK if it's not inside the Developer folder.
         if {${configure.sdkroot} ne ""} {
             allow trace_sandbox "${configure.sdkroot}"


jmroot:
Did hiding the timezone files cause build failures? Does un-hiding them change the built files?

neverpanic:
Hiding them does not cause build failures, but false positive warnings. I don't think there's harm in un-hiding them, because even with those files hidden, the time zone of the machine is available to the builds (otherwise it wouldn't know which files in `/var/db/timezone/zoneinfo` to check).

If there is an effect of timezone on the builds, we should apply a fix to all builds, not only to trace mode builds, and that fix is probably exporting the `TZ` environment variable.

PR #143, opened 2019-08-26
macports1.0: Only check Xcode if building from src

neverpanic:
We don't need to check the Xcode version (or presence) if we're not going to build a port from source. This was previously achieved with a shortcut in `_check_xcode_version` that would always return 0 if an archive was available.

This fixes a situation where `_check_xcode_version` would indicate Xcode isn't required because an archive is available even though a user specifically asked for a build from source (e.g. when running 'sudo port destroot' on a port that sets `use_xcode yes`).

The newly introduced `_target_needs_toolchain` now returns `yes` for all cases where a toolchain is required regardless of archive availability and returns `yes` iff an archive is available for those targets that only need a toolchain when no prebuilt archive is being used.

Tested by applying these wonderfully hackish patches and attempting to install gnuplot (which depends on aquaterm, which – depending on whether you force archive availability to `yes` or `no` will throw an error about requiring Xcode):

```patch
diff --git a/src/macports1.0/macports.tcl b/src/macports1.0/macports.tcl
index 80e75700..91327e29 100644
@@ -426,6 +426,7 @@ proc macports::setxcodeinfo {name1 name2 op} {
     trace remove variable macports::xcodebuildcmd read macports::setxcodeinfo

     try -pass_signal {
+        error "fail immediately"
         set xcodebuild [findBinary xcodebuild $macports::autoconf::xcodebuild_path]
         if {![info exists xcodeversion]} {
             # Determine xcode version
diff --git a/src/port1.0/portutil.tcl b/src/port1.0/portutil.tcl
index abe349be..3f373926 100644
--- a/src/port1.0/portutil.tcl
+++ b/src/port1.0/portutil.tcl
@@ -3359,6 +3359,10 @@ proc _archive_available {} {
         return $archive_available_result
     }

+    if {[option subport] eq "aquaterm"} {
+        return 0
+    }
+
     if {[tbool ports_source_only]} {
         set archive_available_result 0
         return 0
```

See: https://github.com/macports/macports-base/pull/140

### DIFF
--- a/src/macports1.0/macports.tcl
+++ b/src/macports1.0/macports.tcl
@@ -2101,8 +2101,12 @@ proc _mportexec {target mport} {
     # xxx: set the work path?
     set workername [ditem_key $mport workername]
     $workername eval {validate_macportsuser}
+
+    # If the target doesn't need a toolchain (e.g. because an archive is
+    # available and we're not going to build it), don't check for the Xcode
+    # version (and presence for use_xcode yes ports).
     if {![catch {$workername eval "check_variants $target"} result] && $result == 0 &&
-        ![catch {$workername eval {_check_xcode_version}} result] && $result == 0 &&
+        (![macports::_target_needs_toolchain $workername $target] || (![catch {$workername eval {_check_xcode_version}} result] && $result == 0)) &&
         ![catch {$workername eval {check_supported_archs}} result] && $result == 0 &&
         ![catch {$workername eval "eval_targets $target"} result] && $result == 0} {
         # If auto-clean mode, clean-up after dependency install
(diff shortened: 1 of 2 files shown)


src/macports1.0/macports.tcl
```diff
@@ -3644,6 +3647,44 @@ proc macports::_target_needs_deps {target} {
     }
 }
 
+##
+# Determine if the given target needs a toolchain of the given port needs
+# a toolchain. Returns true iff the target will require a compiler (or
```
jmroot: Editing error: the first "needs a toolchain" should be deleted.

neverpanic: Done.

PR #145, opened 2019-09-08
use ##__VA_ARGS__ instead of __VA_ARGS__

MihirLuthra:
On calling `debug_printf(format, …)`  if no variable args are passed, the current definition of the macro would cause errors.

Using ##__VA_ARGS__ would handle such a scenario by removing the trailing comma.


neverpanic:
Normally I would advise using the more modern `__VA_OPT__(,)` macro, but since base needs to keep working with very old compilers, let's use this instead.

PR #146, opened 2019-09-22
bump: Error out if checksums lines are missing from Portfile

kurthindenburg:
Fixes error 'both_checksums": no such variable' when there are no
checksums lines in the Portfile prior to a 'port bump'.


kurthindenburg:
I also don't think anything in bump should fail; it should just warn instead.  I didn't address that with this request.

mojca:
How does this affect ports like `cross/<something>-gcc/`? Neither gcc not binutils ports have checksums in the portfiles and I wouldn't call this an error, the checksums are provided by the portgroup.

kurthindenburg:
Really all this patch does is prevent the error 'both_checksums": no such variable'  - it should not change any other behavior.   Does bump even work for those ports?

PR #147, opened 2019-09-22
set $SDKROOT if configure.sdkroot is set

kencu:
current and recent versions of xcrun look to $SDKROOT
to decide which macOS SDK to choose. If none is set, then
xcrun defaults to the current macOS SDK.

if the user has selected a specific SDK to use by setting
configure.sdkroot, honour that setting by setting $SDKROOT

this has to be done in all port phases where xcrun might be called,
ie configure, build, and destroot

### DIFF
--- a/src/port1.0/portbuild.tcl
+++ b/src/port1.0/portbuild.tcl
@@ -195,6 +195,12 @@ proc portbuild::build_start {args} {
 proc portbuild::build_main {args} {
     global build.cmd
 
+    global configure.sdkroot
+    if { ${configure.sdkroot} ne "" } {
+        ui_debug "setting build SDKROOT to ${configure.sdkroot}"
+        build.env-append "SDKROOT=${configure.sdkroot}"
+    }
+
     set jobs_suffix [build_getjobsarg]
 
     set realcmd ${build.cmd}
--- a/src/port1.0/portconfigure.tcl
+++ b/src/port1.0/portconfigure.tcl
@@ -1477,6 +1477,8 @@ proc portconfigure::configure_main {args} {
                 append_to_environment_value configure $env_var -isysroot${configure.sdkroot}
             }
             append_to_environment_value configure "LDFLAGS" -Wl,-syslibroot,${configure.sdkroot}
+            ui_debug "setting configure SDKROOT to ${configure.sdkroot}"
+            append_to_environment_value configure "SDKROOT" ${configure.sdkroot}
         }
 
         # add extra flags that are conditional on whether we're building universal
--- a/src/port1.0/portdestroot.tcl
+++ b/src/port1.0/portdestroot.tcl
@@ -121,6 +121,13 @@ proc portdestroot::destroot_start {args} {
 }
 
 proc portdestroot::destroot_main {args} {
+
+    global configure.sdkroot
+    if { ${configure.sdkroot} ne "" } {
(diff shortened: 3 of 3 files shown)


kencu:
For your consideration.

This appears to work reliably to specify an SDK when using newer versions of Apple's toolchain, which calls tools via xcrun. 

OTOH, it will likely not affect the build with any toolchain that doesn't look to $SDKROOT during the build. That likely means all non-Apple toolchains (gcc, macports-clang-*, etc) and older OS systems. So it's part of a fix, and perhaps improves the sdk-handling situation sufficiently to consider it.

kencu:
I hope plugging into the build and destroot command scripts was not too "hacky". If there is a more elegant method of setting environment variables for port stages, I was not able to spot it so far.

kencu:
Hmm. Looking through the Portfiles, there do seem to be no shortage of ports that either manually set the SDKROOT as this PR would do globally, or otherwise modify the SDKROOT during the build to some other value.

kencu:
Jeremy has wrapped all the macports-clang-* drivers with xcrun calls as well, so this SDKROOT approach will work there as well, on the supported systems.

cjones051073:
Note this should once merged be cherry picked to the release-2.6 branch as well, if you want it to come out sooner as part of a 2.6.x patch base release.

kencu:
Thanks for the clue -- I knew there had to be a better place to set an env var. It's all done in command_exec in portutil, I see. I'll add it like this one
```
    if {[option macosx_deployment_target] ne ""} {
        set ${varprefix}.env_array(MACOSX_DEPLOYMENT_TARGET) [option macosx_deployment_target]
    }
```

kencu:
Well, that didn't work.... trying some other ideas.

jmroot:
What didn't work exactly? This seems to work fine in initial testing:
```diff
diff --git a/src/port1.0/portutil.tcl b/src/port1.0/portutil.tcl
index edce7ca7..3b0905db 100644
--- a/src/port1.0/portutil.tcl
+++ b/src/port1.0/portutil.tcl
@@ -426,6 +426,9 @@ proc command_exec {command args} {
         set ${varprefix}.env_array(LIBRARY_PATH) [join [option compiler.library_path] :]
     }
     set ${varprefix}.env_array(DEVELOPER_DIR) [option configure.developer_dir]
+    if {[option configure.sdkroot] ne ""} {
+        set ${varprefix}.env_array(SDKROOT) [option configure.sdkroot]
+    }
 
     # Debug that.
     ui_debug "Environment: [environment_array_to_string ${varprefix}.env_array]"
```

kencu:
Maybe I just typo'd it, or brain-fuzzed. I'll try it again, as it should work indeed.

kencu:
As you say, it works fine. Either I had a typo, or I had it somehow in a bad spot -- I'll go with the former. And - no coding after dinner on Saturday night :>

kencu:
Now -- this will replace any SDKROOT that is set by the portfile or by the software's Makefile / system guessing, right? I think that would be a good thing, if it did.

kencu:
And by the way, **Thank You** for taking the time to look this idea over and eventually incorporate it once it was fixed up properly. I think this is going to be a great help!

jmroot:
We can't stop Makefiles from changing variables, but if they set their own value only when none is already set, they'll use ours. The value set by the new code will overwrite any that was set via one of the *.env options.

PR #148, opened 2019-09-26
don't use Apple clang if libc++ on 10.6.8

kencu:
No Apple clang can build on 10.6.8 against libc++
only macports-clang-* compilers will so so

This fixes my corner case where I used Xcode 4.2 with MacPorts 2.6 on 10.6.8, and portconfigure.tcl was returning "clang" for the compiler (which failed).


cjones051073:
I guess once merged this also should be cherry picked to the 2.6 branch, otherwise it will be some time before it makes a release...

kencu:
I will admit I didn't really understand why the initial version of the test was so tricky. Perhaps there was some reason I am not aware of.

PR #151, opened 2019-10-09
gcc_min_version: always allow macports-gcc on PowerPC

kencu:
on Intel systems, macports-gcc is allowed only if the stdlib selection allows it.

on PowerPC, we have to be more flexible and allow macports-gcc versions
even if there might be a stdlib issue, as there are often no other options.

Luckily no issues have been raised with this in the several years it has been allowed.

This selection can be revisited if & when a suitable clang compiler and libc++
implementation become available on PowerPC

### DIFF
--- a/src/port1.0/portconfigure.tcl
+++ b/src/port1.0/portconfigure.tcl
@@ -837,8 +837,12 @@ proc portconfigure::get_min_clang {} {
 proc portconfigure::get_min_gcc {} {
     global compiler.c_standard compiler.cxx_standard compiler.openmp_version compiler.thread_local_storage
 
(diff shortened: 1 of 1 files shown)


kencu:
this needs to be double-checked on some Intel systems, but appears to do the job on PowerPC.

kencu:
So I closed this briefly because I'm still not certain that there are no situations that exist when you might want this function to return a gcc compiler on an Intel system even when the stdlib doesn't match up, but be that as it may, this at least keeps the existing behaviour on intel systems, and allows the PowerPC systems to fall back to gcc7, 6, and 5. So it brings all the PowerPC systems back to functional, and all of mine are building away again.

PR #152, opened 2019-10-11
postflight: Support zsh

neverpanic:
Apple switched the default Shell on macOS Catalina to zsh, so our installer now no longer correctly configures $PATH in the default configuration.

Add support for zsh to fix this. Write the PATH configuration to `~/.zshenv`, which seems to be the default location for this kind of configuration data on zsh.

### DIFF
--- a/portmgr/dmg/postflight.in
+++ b/portmgr/dmg/postflight.in
@@ -221,7 +221,6 @@ USHELL=${USHELL#*shell: }
 case "${USHELL}" in
     */tcsh)
         echo "Detected the tcsh shell."
-        LOGIN_FLAG=""
         ENV_COMMAND="setenv"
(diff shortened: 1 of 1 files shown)


neverpanic:
This pull request was supported by the [MacPortsMeeting 2019](https://trac.macports.org/wiki/Meetings/MacPortsMeeting2019).

neverpanic:
Right, thanks for the tip. Do we need to do the if dance for zshrc and zprofile again or does zshrc automatically get executed even if zprofile exists?

jmroot:
The manual says that zprofile is always executed for login shells.

neverpanic:
@jmroot Should we build a new installer for Catalina for this?

PR #153, opened 2019-10-12
mportinit: Disable running unknown cmds in $SHELL

neverpanic:
Tcl defaults to running unknown commands in a shell automatically. This can inadvertently lead to running things in the shell instead of using a Tcl builtin, e.g. due to a typo.

It does not seem like we are currently using this functionality, and we really shouldn't, so disable it on startup to ensure it is not used.

### DIFF
--- a/src/macports1.0/macports.tcl
+++ b/src/macports1.0/macports.tcl
@@ -587,6 +587,10 @@ proc macports::_is_valid_developer_dir {dir} {
 
 
 proc mportinit {{up_ui_options {}} {up_options {}} {up_variations {}}} {
+    # Disable unknown(n)'s behavior of running unknown commands in the system
+    # shell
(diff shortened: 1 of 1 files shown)


raimue:
Did this actually cause any problems for you? To me, this setting seems a bit out of scope of `mportinit`. What you describe is only true for interactive tclsh, while a non-interactive tclsh will not search and execute external binary files.

```
$ tclsh
% ls
foo bar baz
$ echo "ls" | tclsh
invalid command name "ls"
```

neverpanic:
No, this didn't cause problems, I just happed to read the `unknown(n)` manpage, which doesn't state that this behavior isn't enabled in non-interactive tclsh, so I assumed it was.

I guess we should revert this then.

PR #154, opened 2019-10-12
add variable-replace to portfile(7)

ra1nb0w:
this PR add `variable-replace` extension to portfile(7)

### DIFF
--- a/doc/portfile.7
+++ b/doc/portfile.7
@@ -2578,6 +2578,16 @@ from the
 .br
 .Sy Example:
 .Dl configure.args-delete --with-gnomedb
+.It Va variable Ns - Ns Ic replace Ar old_item new_item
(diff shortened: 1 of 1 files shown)


doc/portfile.7
```diff
@@ -2578,6 +2578,16 @@ from the
 .br
 .Sy Example:
 .Dl configure.args-delete --with-gnomedb
+.It Va variable Ns - Ns Ic replace Ar old_item new_item
+Replace
+.Ar old_item
+with
+.Ar new_item
+on the
```
jmroot: Saying "on" sounds a bit strange to me. I would use "in" instead.

jmroot:
Thanks.

ra1nb0w:
thank you

PR #156, opened 2019-10-13
Only check OS/platform on MacOS

satraul:
Temporary fix for Linux not working

References: http://trac.macports.org/ticket/58497

### DIFF
--- a/src/macports1.0/macports.tcl
+++ b/src/macports1.0/macports.tcl
@@ -706,7 +706,7 @@ proc mportinit {{up_ui_options {}} {up_options {}} {up_variations {}}} {
     }
 
     # Check that the current platform is the one we were configured for, otherwise need to do migration
-    if {($os_platform ne $macports::autoconf::os_platform) || ($os_major != $macports::autoconf::os_major)} {
+    if {($os_platform ne $macports::autoconf::os_platform) || ($os_platform eq "darwin" && $os_major != $macports::autoconf::os_major)} {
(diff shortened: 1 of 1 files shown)


jmroot:
Seems like an os_platform mismatch should always be fatal. I'd suggest this instead:
```diff
index 63439d85..b1172076 100644
--- a/src/macports1.0/macports.tcl
+++ b/src/macports1.0/macports.tcl
@@ -706,7 +706,7 @@ proc mportinit {{up_ui_options {}} {up_options {}} {up_variations {}}} {
     }
 
     # Check that the current platform is the one we were configured for, otherwise need to do migration
-    if {($os_platform ne $macports::autoconf::os_platform) || ($os_major != $macports::autoconf::os_major)} {
+    if {($os_platform ne $macports::autoconf::os_platform) || ($os_platform eq "darwin" && $os_major != $macports::autoconf::os_major)} {
         ui_error "Current platform \"$os_platform $os_major\" does not match expected platform \"$macports::autoconf::os_platform $macports::autoconf::os_major\""
         ui_error "If you upgraded your OS, please follow the migration instructions: https://trac.macports.org/wiki/Migration"
         return -code error "OS platform mismatch"
```

mojca:
@raimue additionally noted:
> Changing just this one line will not help much, as the installed ports also store the platform information in the registry and will all be considered outdated. I guess the real fix would be not to store the Linux kernel version in `os_major`, but I am also not sure what else to use.

However I'm not as much concerned if we end up with "outdated" ports and just rebuild them. At the moment we are not even able to run anything.

PR #157, opened 2019-10-14
WIP: port.tcl: Offer automated switch to git-based source

satraul:
Reference: https://trac.macports.org/wiki/Meetings/MacPortsMeeting2019/ContributingUXImprovements


src/port/port.tcl
```diff
@@ -3927,6 +3927,128 @@ proc action_echo { action portlist opts } {
...
 
+proc prompt { message default variable } {
+    puts -nonewline $message
+    flush stdout
+    gets stdin variable
+    if {[string trimleft $variable] eq ""} {
+        set variable $default
+    }
+    return 0
+}
```
neverpanic: This should probably go if it's unused.

src/port/port.tcl
```diff
@@ -3927,6 +3927,128 @@ proc action_echo { action portlist opts } {
...
+
+proc offer_clone {} {
+    global macports::os_platform macports::os_major macports::os_arch \
+        macports::sources macports::autoconf::rsync_path \
+        macports:rsync_options macports::getindex macports::sources_default
+
+    set msg4 {Rsyncing a PortIndex file...}
```
neverpanic: The variable names need some love and consideration.

src/port/port.tcl
```diff
@@ -3984,6 +4106,13 @@ proc action_portcmds { action portlist opts } {
                     array unset env *
                     array set env [array get boot_env]
 
+                    # Offer setting up the switch to git-based source etc. for first time contributors
+                    set first_time 1
```
neverpanic: Needs figuring out when `first_time` is actually true.

src/port/port.tcl
```diff
@@ -3984,6 +4106,13 @@ proc action_portcmds { action portlist opts } {
...
+                    if {$first_time} {
+                        offer_clone
+                    }
+                    return 0
```
neverpanic: We still want to edit the file after doing the setup.

src/port/port.tcl
```diff
@@ -3927,6 +3927,128 @@ proc action_echo { action portlist opts } {
...
+
+    # OFFER GIT-BASED SOURCE AND ASK FOR PATH
+    set msg1 "You do not seem have a working copy of the macports-ports git repository, which is required to contribute this change."
+    set retvalue [$macports::ui_options(questions_yesno) $msg1 "clone" {} {y} 0 "Create one now?"]
+    if {$retvalue == 1} {
+        # quit as user answered 'no'
+        return 0
+    }
+
+    set path [$macports::ui_options(questions_input) "Where do you want to create your working copy?" [file join [file normalize ~] macports-ports] 0]
```
neverpanic: I'd show "~/macports-ports" to the user and then run the `[file normalize]` on the return value. This would give the user the option to enter paths containing `~` themselves.

src/port/port.tcl
```diff
@@ -3927,6 +3927,128 @@ proc action_echo { action portlist opts } {
...
+
+    # WALK DOWN FROM PATH TO ROOT AND ASK FOR PERMISSIONS IF NEEDED
+    set paths {}
+    set temp $path
+    while {$temp != [file normalize /]} {
+        ui_msg "$temp [file readable $temp]"
+        if {[file readable $temp]} {
```
neverpanic: I think we should test whether the other permission bits contain the "executable" bit instead, because that would be independent of which user we're running this under, while `file readable` isn't.

Let's use this:
```tcl
expr {[string index [file attributes /Users/clemens -permissions] end] & 1}
```

src/port/port.tcl
```diff
@@ -3927,6 +3927,128 @@ proc action_echo { action portlist opts } {
...
+        # quit as user answered 'no'
+        return 0
+    }
+
+    set path [$macports::ui_options(questions_input) "Where do you want to create your working copy?" [file join [file normalize ~] macports-ports] 0]
+
+    # WALK DOWN FROM PATH TO ROOT AND ASK FOR PERMISSIONS IF NEEDED
+    set paths {}
+    set temp $path
+    while {$temp != [file normalize /]} {
```
neverpanic: `[file normalize /]` is always `/`, so that's not needed (unless you wanted to do the normalization on `$temp`?

Also, string comparisons should use `eq` in Tcl, because that doesn't trigger integer comparison if the arguments look like integers.

src/port/port.tcl
```diff
@@ -3927,6 +3927,128 @@ proc action_echo { action portlist opts } {
...
+        ui_msg "$temp [file readable $temp]"
+        if {[file readable $temp]} {
+            break
+        }
+        lappend paths $temp
+        set temp [file dirname $temp]
+    }
+    set msg2 "The 'macports' user needs search access to ~/macports-ports to use ports in
```
neverpanic: Instead of `~/macports-ports` this should be the actual path specified.

src/port/port.tcl
```diff
@@ -3927,6 +3927,128 @@ proc action_echo { action portlist opts } {
...
+"
+    if {[llength $paths] > 0} {
+        set allow [$macports::ui_options(questions_yesno) $msg2 "allow" $paths {y} 0 "Do you want to allow this?"]
+        if {$allow == 1} {
+            return 0
+        }
+    }
+
+    # CREATE GIT SOURCE BY CLONING MACPORTS-PORTS
+    if {[catch {system "${macports::findBinary git} clone --depth 100 $url $path"} result]} {
```
neverpanic: This should be `[macports::findBinary git]`. `${macports::findBinary git}` would look for the variable named `findBinary git` in the `macports` namespace. No idea whether Tcl variables can contain spaces – wouldn't be surprised if it's possible.

src/port/port.tcl
```diff
@@ -3927,6 +3927,128 @@ proc action_echo { action portlist opts } {
...
+        ui_debug $::errorInfo
+        return 1
+    }
+
+    # TODO: CHANGE PERMISSIONS OF $PATH TO ROOT
```
neverpanic: Let's not. ;-)

src/port/port.tcl
```diff
@@ -3927,6 +3927,128 @@ proc action_echo { action portlist opts } {
...
+
+    # COPY INDEXFILE FROM OLD SOURCE TO GIT SOURCE
+    set default_source_url [lindex $sources_default 0]
+    set indexfile [macports::getindex $default_source_url]
+    if {[catch {file copy -force $indexfile $path} error]} {
+        ui_debug $::errorInfo
+        ui_error "file copy: $error"
+        return -code error "Copying Rsync IndexFile to new git source failed"
```
neverpanic: `s/Rsync IndexFile/PortIndex/`.

Also we need to copy both `PortIndex` and `PortIndex.quick`, so we need two copy operations here.

src/port/port.tcl
```diff
@@ -3927,6 +3927,128 @@ proc action_echo { action portlist opts } {
...
+    # Replace sources.conf with sources.conf.bump
+    if {[catch {move -force $tmpfile $indexfile} error]} {
+        ui_debug $::errorInfo
+        ui_error "bump: $error"
+        file delete "$tmpfile"
+        return -code error "bump overwrite failed"
+    }
+
+    # Restore sources.conf attributes
+    file attributes $portfile {*}$attributes
```
neverpanic: `$portfile` sounds wrong here, shouldn't this be `$default_source_url`?

src/port/port.tcl
```diff
@@ -3927,6 +3927,128 @@ proc action_echo { action portlist opts } {
...
+
+    # syncing ports tree.
+    if {![info exists options(ports_selfupdate_no-sync)]} {
```
neverpanic: You can drop this check.

src/port/port.tcl
```diff
@@ -3927,6 +3927,128 @@ proc action_echo { action portlist opts } {
...
+        if {$comp > 0} {
```
neverpanic: `$comp` is set when selfupdate updated base to a newer version of MacPorts base, in which case Portfiles might need the newer version to parse them, so it tells users to re-run selfupdate to trigger the indexing and doesn't do it in this run.

This doesn't apply here because we're not selfupdating, so this entire branch can go.

src/port/port.tcl
```diff
@@ -3927,6 +3927,128 @@ proc action_echo { action portlist opts } {
...
+        }
+        try {
+            mportsync $optionslist
+        }  catch {{*} eCode eMessage} {
+            return -code error "Couldn't sync the ports tree: $eMessage"
+        }
+    }
+
+    ui_msg "Your /opt/local/etc/macports/sources.conf has been updated to use this repository."
+}
```
neverpanic: You may want a `return yes/1/true/truthyvalue` here.

You should probably also take a second look at your return values in this function.

src/port/port.tcl
```diff
@@ -3927,6 +3927,161 @@ proc action_echo { action portlist opts } {
...
+        file delete "$tmpfile"
+        return -code error "sources.conf overwrite failed"
+    }
+
+    # Restore sources.conf attributes
+    file attributes $confpath {*}$attributes
+
+    # Update the ports tree
+    try {
+        mportsync
```
satraul: Need help with this one. At this point we've setup the new git source including updating sources.conf, but calling `mportsync` still updates the old source.

src/port/port.tcl
```diff
@@ -3927,6 +3927,161 @@ proc action_echo { action portlist opts } {
...
 
+proc offer_clone {} {
+    global macports::os_platform macports::os_major macports::os_arch \
+        macports::sources macports::autoconf::rsync_path \
+        macports:rsync_options macports::getindex macports::sources_default \
+        macports::sudo_user macports::prefix macports::portdbpath
+
+    # CHECK IF SOURCES.CONF IS STILL DEFAULT (INDICATING THAT USER IS FIRST TIME CONTRIBUTOR)
+    set confpath "${macports::prefix}/etc/macports/sources.conf"
+    if {[catch {exec [macports::findBinary diff] "-q" $confpath "$confpath.default"} _]} {
```
satraul: We offer this change if `sources.conf` is identical with `sources.conf.default` or this proc has ran before without unexpectedly failing. We keep track that by creating a file `${macports::portdbpath}/contrib`

src/port/port.tcl
neverpanic: I see two options:
 - Call `mportshutdown` and `mportinit` again, then run `mportsync`
 - Use `execl` to re-execute the same command after the changes were made, see https://github.com/macports/macports-base/blob/gsoc17-migrate/src/port/port.tcl#L2829

PR #158, opened 2019-10-14
Add Dockerfile

mojca:
We used @Korusuke's work as the basis to create a slightly more improved Docker image for MacPorts base and suggest to merge the image to the base where it makes most sense.

What's left to be done is:
* fix the `configure` arguments
* reduce the number of runtime dependencies to just the required ones (most of dependencies are probably just build dependencies, but this needs to be checked)
* fix `Portfiles` which fail at `portindex`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,34 @@
+FROM ubuntu:18.04 AS builder
+
+RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y \
+        build-essential autotools-dev debhelper \
+        gobjc gobjc++ gobjc-multilib gobjc++-multilib \
+        libgnustep-base-dev gnustep-core-devel sqlite \
+        libsqlite3-dev openssl libcurl4-openssl-dev curl \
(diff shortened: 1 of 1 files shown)


cceneag:
This pull request introduces a new `Dockerfile` to set up a multi-stage build process for an Ubuntu-based development environment. The `Dockerfile` includes the installation of required dependencies, configuration, and compilation of the project, as well as environment setup.

### Key Changes:

#### Multi-stage Docker Build:
* Added a `builder` stage to the `Dockerfile` that installs necessary build tools and dependencies, configures the project with GNU Objective-C runtime and foundation, and compiles it using `make`.
* Added a second stage in the `Dockerfile` to create a lightweight runtime image by copying the compiled binaries and setting up the environment.

#### Dependency Management:
* Installed a comprehensive list of dependencies, including GNUstep libraries, SQLite, OpenSSL, Tcl, and other essential tools, for both the build and runtime stages.

#### Environment Configuration:
* Configured the `PATH` environment variable to include `/opt/local/bin` for accessing the installed binaries.

PR #159, opened 2019-10-14
portindex.tcl: Drop support for specifying stdlib

mojca:
While trying to fix https://trac.macports.org/ticket/59318 I only noticed a duplicate at https://trac.macports.org/ticket/55471 when I tried to make a pull request (I was initially working on the wrong/old/generated file).

Maybe we nevertheless want to drop support for specifying stdlib as it's unlikely we'll need it again.
See: https://trac.macports.org/ticket/59069


kencu:
It's quite likely that 10.4 and 10.5 intel might use libc++, and yet 10.4 and 10.5 PPC might not. So having the capability to change it still seems useful to me.

I have my 10.5 Intel setup as stdlib=libc++ now, for example. Just in case anyone wondered if it was useful at all...

PR #161, opened 2019-11-23
restrict compiler for C++11 and thread-local storage

MarcusCalhoun-Lopez:
See https://lists.macports.org/pipermail/macports-dev/2019-November/thread.html#41503

### DIFF
--- a/src/port1.0/portconfigure.tcl
+++ b/src/port1.0/portconfigure.tcl
@@ -788,7 +788,15 @@ proc portconfigure::get_min_command_line {compiler} {
             } elseif {${compiler.cxx_standard} >= 2014} {
                 set min_value [max_version $min_value 600.0.54]
             } elseif {${compiler.cxx_standard} >= 2011} {
-                set min_value [max_version $min_value 500.2.75]
+                if {${compiler.thread_local_storage}} {
+                    # macOS has supported thread-local storage since Mac OS X Lion.
+                    # So __thread (GNU extension) and _Thread_local (C11) could be used.
+                    # However, the C++11 keyword was not supported until Xcode 8
+                    #    (https://developer.apple.com/videos/play/wwdc2016-405/?time=354).
+                    set min_value [max_version $min_value 800.0.38]
+                } else {
+                    set min_value [max_version $min_value 500.2.75]
+                }
             }
         }
         llvm-gcc-4.2 -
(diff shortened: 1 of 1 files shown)


kencu:
It looks like compiler.thread_local_storage will only function properly if the compiler.cxxstandard is set to 2011 or greater?

We'd have to add that to the guide.

Or we could unhide it from the cxxstandard block, and then it could always work.

up to you.

jmroot:
There is no thread local storage prior to C++11. Asking for it with an earlier C++ standard can't be expected to work.

MarcusCalhoun-Lopez:
I have added a `${compiler.thread_local_storage}` check for the older Apple-provided compilers.
The change might be more for the sake of completeness since [a previous check](https://github.com/macports/macports-base/blob/2249c806dafa35c0bd2b2bce0bec29c94fa79856/src/port1.0/portconfigure.tcl#L768) alone might be sufficient.

jmroot:
It's possible to be using llvm-gcc-4.2 on Lion without the added check. Thanks.

kencu:
So be sure to add to the guide then that this command only works if the cxx standard is also set.

MarcusCalhoun-Lopez:
You point about proper documentation is well-taken.
If this pull request is accepted, I will add a few words on the [Trac](https://trac.macports.org/wiki/CompilerSelection).

However, I would respectively suggest that this command does not **only** work when `compiler.cxx_standard` is set.
`compiler.thread_local_storage yes` is proper and correct if the code uses `__thread` or `_Thread_local`.

Apple even [suggests](https://developer.apple.com/videos/play/wwdc2016-405/?time=354) that `_Thread_local` is to be preferred over `thread_local` for performance reasons.

kencu:
for a c++ port that gets a thread_local error when building, adding 

"compiler.thread_local_storage yes"

will not fix the port unless a cxxstandard is also set >= 2011.

yes, that's precise. Just imho in would be less confusing, less tickets, less time wasted, and simpler if it just blacklisted Xcode clangs <800 no matter what, because then it would always work.

But I know my "practical just-works" approach bangs heads with you guys sometimes.

chrstphrchvz:
See #214 and https://trac.macports.org/ticket/61418: the original approach here only affected C++11 ports, but should also affect C++14 ports.

PR #162, opened 2019-12-02
Fix C++14 minimum clang version

ryandesign:
Ports that specify `compiler.cxx_standard 2014` fail on OS X 10.9 (see https://trac.macports.org/ticket/59323 or https://build.macports.org/builders/ports-10.9_x86_64-builder/builds/107245) because its clang 600.x.x gets used, even though only clang 602 and later support C++14 (see https://trac.macports.org/ticket/57682).


MarcusCalhoun-Lopez:
This seems entirely reasonable.
When this PR is merged, I (or whoever gets to it first) need to remember to update [CompilerSelection](https://trac.macports.org/wiki/CompilerSelection).

PR #163, opened 2019-12-04
allow Portfile to turn off verbose logging

MarcusCalhoun-Lopez:
Setting CC_PRINT_OPTIONS and CC_PRINT_OPTIONS_FILE changes how Clang
handles the compiler flags -v and -Wl,-v.
If a port relies on the default behavior (e.g. FindOpenMP in CMake),
then errors can ensue.


szhorvat:
Wow ... This issue should be clearly documented! It just spent an inordinate about of time trying to figure out why a port wouldn't compile with OpenMP enabled.

PR #165, opened 2019-12-08
allow compiler to use environment variables only

MarcusCalhoun-Lopez:
For many ports, the compiler flags
  -I${prefix}/include,
  -L${prefix}/lib,
  -isysroot${configure.sdkroot}, and
  -Wl,-syslibroot,${configure.sdkroot}
are unnecessary because of the environment variables
  CPATH,
  LIBRARY_PATH, and
  SDKROOT.
compiler.limit_flags prevents MacPorts from attempting to add
unnecessary compiler flags, which can cause problems.

See https://trac.macports.org/wiki/CompilerEnvironmentVariables
for details.

### DIFF
--- a/src/port1.0/portconfigure.tcl
+++ b/src/port1.0/portconfigure.tcl
@@ -234,8 +234,24 @@ options configure.optflags \
 default configure.optflags      -Os
 default configure.cflags        {${configure.optflags}}
 default configure.objcflags     {${configure.optflags}}
-default configure.cppflags      {-I${prefix}/include}
-default configure.ldflags       {-L${prefix}/lib -Wl,-headerpad_max_install_names}
+default configure.cppflags      {[portconfigure::configure_get_cppflags]}
+proc portconfigure::configure_get_cppflags {} {
+    global prefix
+    if {[option compiler.limit_flags]} {
+        return ""
+    } else {
+        return -I${prefix}/include
+    }
+}
+default configure.ldflags       {[portconfigure::configure_get_ldflags]}
+proc portconfigure::configure_get_ldflags {} {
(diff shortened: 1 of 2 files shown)


pmetzger:
Just FYI, the proper term is "environment", not "environmental" — I edited the PR to reflect that but we should make sure we're consistent in code and documentation.

MarcusCalhoun-Lopez:
Thank you for the information.
It would seem I have using the wrong vocabulary.

jmroot:
Is it intended that the Portfile using this feature are responsible for ensuring that the compiler version in use supports these variables?

MarcusCalhoun-Lopez:
The base should handle  compiler selection.
I need to fix this.
Thank you for pointing this out.

CPATH and LIBRARY_PATH seem to be widely supported (except Clang 1.x).
SDKROOT has issues:
* on older systems where `${configure.sdkroot}` is non-empty but `/usr/lib/libxcselect.dylib` does not exist
* with older versions of GCC

I have attempted to gather my results so far in the [wiki](https://trac.macports.org/wiki/CompilerEnvironmentVariables).
As always, any input and/or corrections would be appreciated.

kencu:
This repo <https://github.com/b-man/xcode-tools> has a reverse-engineered copy of xcrun and xcode-select that mimics the behaviour of newer versions of these tools on older systems. It could be used to make versions of these tools that would run on 10.4 to 10.8 and allow those systems to also use the same current xcrun syntax and function.

kencu:
This plan does appear to be the logical next step down the path transitioning to using SDKROOT and the other envvars instead of the problematic baked-in flags.

PR #167, opened 2020-01-26
Fixes to HTTP compression behavior

dgilman:
- Fix build on old MacOS X
- Leave compression OFF by default
- For livecheck, turn it ON by default
- Expose livecheck compression toggle to portfile via livecheck.compression

### DIFF
--- a/src/pextlib1.0/curl.c
+++ b/src/pextlib1.0/curl.c
@@ -44,6 +44,8 @@
 #include <sys/select.h>
 #include <utime.h>
 
+#include <AvailabilityMacros.h>
+
 #include <curl/curl.h>
 
 #include <tcl.h>
@@ -59,10 +61,15 @@
 #define _CURL_MINIMUM_XFER_TIMEOUT	((long)(60))		/* 1 minute */
 #define _CURL_MINIMUM_PROGRESS_INTERVAL ((double)(0.2)) /* 0.2 seconds */
 
+/* pre-10.8 curl does not appear to support compression out of the box */
+#if MAC_OS_X_VERSION_MIN_REQUIRED >= MAC_OS_X_VERSION_10_8
+
 #if defined CURLOPT_ACCEPT_ENCODING
 #define _CURL_ENCODING CURLOPT_ACCEPT_ENCODING
 #elif defined CURLOPT_ENCODING
 #define _CURL_ENCODING CURLOPT_ENCODING
+#endif
+
 #endif
 
 /* ========================================================================= **
--- a/src/pextlib1.0/curl.c
+++ b/src/pextlib1.0/curl.c
@@ -163,7 +163,7 @@ SetResultFromCurlMErrorCode(Tcl_Interp *interp, CURLMcode inErrorCode)
 /**
  * curl fetch subcommand entry point.
  *
- * syntax: curl fetch [--disable-epsv] [--ignore-ssl-cert] [--remote-time] [-u userpass] [--effective-url lasturlvar] [--progress "builtin"|callback] [--disable-compression] url filename
+ * syntax: curl fetch [--disable-epsv] [--ignore-ssl-cert] [--remote-time] [-u userpass] [--effective-url lasturlvar] [--progress "builtin"|callback] [--enable-compression] url filename
  *
  * @param interp		current interpreter
  * @param objc			number of parameters
@@ -204,7 +204,7 @@ CurlFetchCmd(Tcl_Interp* interp, int objc, Tcl_Obj* CONST objv[])
 		struct curl_slist *headers = NULL;
 		struct CURLMsg *info = NULL;
 		int running; /* number of running transfers */
-		char* acceptEncoding = "";
+		char* acceptEncoding = NULL;
 
 		/* we might have options and then the url and the file */
 		/* let's process the options first */
@@ -290,8 +290,8 @@ CurlFetchCmd(Tcl_Interp* interp, int objc, Tcl_Obj* CONST objv[])
 					theResult = TCL_ERROR;
 					break;
 				}
-         } else if (strcmp(theOption, "--disable-compression") == 0) {
-				acceptEncoding = NULL;
+         } else if (strcmp(theOption, "--enable-compression") == 0) {
+				acceptEncoding = "";
 			} else {
 				Tcl_ResetResult(interp);
 				Tcl_AppendResult(interp, "curl fetch: unknown option ", theOption, NULL);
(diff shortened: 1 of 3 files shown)


src/pextlib1.0/curl.c
```diff
@@ -283,8 +290,8 @@ CurlFetchCmd(Tcl_Interp* interp, int objc, Tcl_Obj* CONST objv[])
 					theResult = TCL_ERROR;
 					break;
 				}
-         } else if (strcmp(theOption, "--disable-compression") == 0) {
-				acceptEncoding = NULL;
+         } else if (strcmp(theOption, "--enable-compression") == 0) {
```
tobypeterson: This line is indented with spaces instead of tabs.

dgilman:
I've fixed up the indentation across the entire file and added a "modeline".  It is the vim settings I have to get the thing to indent according to the style used in the rest of the file but I am having trouble in my personal vim setup getting it to properly read the modeline on start.  If there are any vim wizards here I'd appreciate feedback, but even if the modeline is non functional it at least serves as documentation of the correct formatting.

raimue:
I think the modeline and whitespace re-indentation is not related to this feature and I do not see the necessity to make these changes across the whole file. Let's discuss this separately. Could you please move this change to a new pull request instead?

src/pextlib1.0/curl.c
```diff
@@ -1,5 +1,6 @@
 /*
  * curl.c
+ * vim: set noet ts=8 sw=8 tw=80 sts=-1 cino=t0 :
```
jmroot: If you add a modeline, please use the standard MacPorts one.

src/pextlib1.0/curl.c
```diff
@@ -59,12 +62,17 @@
 #define _CURL_MINIMUM_XFER_TIMEOUT	((long)(60))		/* 1 minute */
 #define _CURL_MINIMUM_PROGRESS_INTERVAL ((double)(0.2)) /* 0.2 seconds */
 
+/* pre-10.8 curl does not appear to support compression out of the box */
+#if MAC_OS_X_VERSION_MIN_REQUIRED >= MAC_OS_X_VERSION_10_8
```
jmroot: This would be better done as a curl version check.

src/port1.0/portlivecheck.tcl
```diff
@@ -56,11 +56,13 @@ default livecheck.name default
 default livecheck.distname default
 default livecheck.version {$version}
 default livecheck.ignore_sslcert no
+default livecheck.compression yes
```
jmroot: Adding an option for this may be overkill since livecheck will always want decompressed text to search. I guess it doesn't hurt though.

src/pextlib1.0/curl.c
jmroot: Actually, it looks like curl supports compression back to at least 10.6. The use of https in the tests is likely the problem.

jmroot: Hmm, nope, `/usr/bin/curl --compressed https://www.whatsmyip.org/http-compression-test/` works fine, so it's not the TLS. But it isn't lack of CURLOPT_ENCODING either.

dgilman: My hunch was that the Apple curl wasn't compiled with zlib or something like that. If you have an old Mac around can you see what kind of request headers it's producing?  If that is the case it makes sense to just ban on OSX version.

dgilman:
The indentation is fixed

src/pextlib1.0/curl.c
jmroot:
```
% /usr/bin/curl -v --compressed https://www.whatsmyip.org/http-compression-test/
* About to connect() to www.whatsmyip.org port 443 (#0)
*   Trying 204.11.35.98... connected
* Connected to www.whatsmyip.org (204.11.35.98) port 443 (#0)
* SSLv3, TLS handshake, Client hello (1):
* SSLv3, TLS handshake, Server hello (2):
* SSLv3, TLS handshake, CERT (11):
* SSLv3, TLS handshake, Server key exchange (12):
* SSLv3, TLS handshake, Server finished (14):
* SSLv3, TLS handshake, Client key exchange (16):
* SSLv3, TLS change cipher, Client hello (1):
* SSLv3, TLS handshake, Finished (20):
* SSLv3, TLS change cipher, Client hello (1):
* SSLv3, TLS handshake, Finished (20):
* SSL connection using DHE-RSA-AES256-SHA
* Server certificate:
* 	 subject: CN=random.whatsmyip.org
* 	 start date: 2019-12-24 19:43:20 GMT
* 	 expire date: 2020-03-23 19:43:20 GMT
* 	 subjectAltName: www.whatsmyip.org matched
* 	 issuer: C=US; O=Let's Encrypt; CN=Let's Encrypt Authority X3
* 	 SSL certificate verify ok.
> GET /http-compression-test/ HTTP/1.1
> User-Agent: curl/7.19.7 (universal-apple-darwin10.0) libcurl/7.19.7 OpenSSL/0.9.8y zlib/1.2.3
> Host: www.whatsmyip.org
> Accept: */*
> Accept-Encoding: deflate, gzip
> 
< HTTP/1.1 200 OK
< Date: Fri, 31 Jan 2020 02:02:58 GMT
< Server: Apache
< X-Powered-By: Web Server Built Entirely of Legos
< Cache-Control: max-age=0
< Expires: Fri, 31 Jan 2020 02:02:58 GMT
< Vary: Accept-Encoding
< Content-Encoding: gzip
< Content-Length: 3177
< Content-Type: text/html; charset=UTF-8
< 
<!DOCTYPE html>
```
(actual page content trimmed)

jmroot: With current master (i.e. none of the changes in this PR applied), running both `curl fetch` and `curl fetch  --disable-compression` in Tcl sends the exact same headers in each case:
```
GET / HTTP/1.1
User-Agent: MacPorts/2.6.99 libcurl/7.19.7
Host: x.x.x.x:1234
Accept: */*
```

jmroot: The problem is that CURLOPT_ENCODING is not defined as a macro, but as part of an enum. Changing the #elif to an #else should fix it.

dgilman:
I've updated the c preprocessor to handle the old definition. Thanks to jmroot for investigating it, I don't have an old mac anywhere.

src/pextlib1.0/curl.c
```diff
@@ -44,6 +44,8 @@
 #include <sys/select.h>
 #include <utime.h>
 
+#include <AvailabilityMacros.h>
```
jmroot: Looks like this is no longer used.

dgilman: Done

jmroot:
Thanks!

PR #168, opened 2020-02-01
Pass -Wl,-dead_strip,-bind_at_load

rex4539:
From `man ld`:
```
-dead_strip
                 Remove functions and data that are unreachable by the entry point or exported symbols.
```
```
-bind_at_load
                 Sets a bit in the mach header of the resulting binary which tells dyld to bind all symbols when the binary is loaded,
                 rather than lazily.
```
Before:
```
% wc -c MacPorts.dylib
   20160 MacPorts.dylib
% wc -c Pextlib.dylib 
  131396 Pextlib.dylib
% wc -c daemondo 
   41264 daemondo
% wc -c registry.dylib 
   85320 registry.dylib
```
After:
```
% wc -c MacPorts.dylib 
   20112 MacPorts.dylib
% wc -c Pextlib.dylib 
  130556 Pextlib.dylib
% wc -c daemondo            
   39336 daemondo
% wc -c registry.dylib
   85112 registry.dylib
```


jmroot:
Why would we want to use -bind_at_load? It seems like it's just going to increase startup time by spending time binding symbols that we don't end up using.

rex4539:
`-bind_at_load` shaves some binary size.

I have been using this for a few days and haven't seen any drawbacks in speed.

pmetzger:
I don't think the "I haven't seen a reduction in speed" is an acceptable answer. Unless you've benchmarked the change in run time carefully on a variety of platforms, you don't really know.

Generally: I don't think we want this. A trivial space reduction in exchange for potential mysterious problems and for an increase in run time seems unreasonable.

rex4539:
Fair enough. Thanks for taking the time to test this.

jmroot:
Thanks for taking the time to contribute. If the dead code can be identified (maybe with the clang static analyzer or similar tools?) we can still remove it at the source level.

raimue:
As `-dead_strip` works on unused symbols, it might already help to just compare the list of symbols reported by `otool` for the libraries for both _before_ and _after_ as shown in the original post.

PR #169, opened 2020-04-19
Abort when a port includes a nonexistent PortGroup

kurthindenburg:
Currently, only a warning is printed and the processing of the port
continues. This typically results in a port not completing later.
After this patch, an error is printed with the port name added and
then the port processing aborts.

closes https//trac.macports.org/ticket/59787

### DIFF
--- a/src/port1.0/portutil.tcl
+++ b/src/port1.0/portutil.tcl
@@ -2580,7 +2580,7 @@ proc PortGroup {group version} {
         ui_debug "Sourcing PortGroup $group $version from $groupFile"
     } else {
         ui_error "${subport}: PortGroup ${group} ${version} could not be located. ${group}-${version}.tcl does not exist."
-        return -code error "Invalid PortGroup found"
+        return -code error "PortGroup not found"
     }
 }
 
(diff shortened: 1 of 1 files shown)


kurthindenburg:
Ideally we would want to check the entire portfile before aborting but that doesn't seem trivial.  The return is needed to abort processing when invalid PortGroups are in dependencies.

src/port1.0/portutil.tcl
```diff
@@ -2579,8 +2579,8 @@ proc PortGroup {group version} {
         uplevel "source $groupFile"
         ui_debug "Sourcing PortGroup $group $version from $groupFile"
     } else {
-        lappend PortInfo(portgroups) [list $group $version ""]
-        ui_warn "PortGroup ${group} ${version} could not be located. ${group}-${version}.tcl does not exist."
+        ui_error "${subport}: PortGroup ${group} ${version} could not be located. ${group}-${version}.tcl does not exist."
+        return -code error "Invalid PortGroup found"
```
jmroot, line 2583: I would rather say "PortGroup not found". We don't know for certain that it's invalid; the file could just be missing.

kurthindenburg:
jmr approved on IRC

PR #171, opened 2020-04-26
portconfigure: Add support for Macports Clang versions 10+

cjones051073:
Minor update to `portconfigure.tcl` to extend the regex used to detect macports provide clang compiler to support the dropping of the minor version from the port name, starting with `clang-10` (https://github.com/macports/macports-ports/pull/6902).

Note that the changes here do not require any changes in the ports tree, either to add the `clang-10` compiler or to the resources to add it to the known lists, so can be applied before they are done. However the changes here are required to be available before `macports-clang-10` can be used as a compiler option.

Once merged I would propose cherry-picking this to the 2.6 release branch, such as to make it available in the next release from that branch.

### DIFF
--- a/src/port1.0/portconfigure.tcl
+++ b/src/port1.0/portconfigure.tcl
@@ -340,30 +340,26 @@ proc portconfigure::configure_start {args} {
 
     set compiler [option configure.compiler]
     set valid_compilers {
-        {^apple-gcc-(4\.[02])$}             {MacPorts Apple GCC %s}
-        {^cc$}                              {System cc}
-        {^clang$}                           {Xcode Clang}
-        {^gcc$}                             {System GCC}
-        {^gcc-(3\.3|4\.[02])$}              {Xcode GCC %s}
-        {^llvm-gcc-4\.2$}                   {Xcode LLVM-GCC 4.2}
-        {^macports-clang$}                  {MacPorts Clang (port select)}
-        {^macports-clang-(\d+\.\d+)$}       {MacPorts Clang %s}
-        {^macports-gcc$}                    {MacPorts GCC (port select)}
-        {^macports-gcc-(\d+(?:\.\d+)?)$}    {MacPorts GCC %s}
-        {^macports-llvm-gcc-4\.2$}          {MacPorts LLVM-GCC 4.2}
-        {^macports-g95$}                    {MacPorts G95}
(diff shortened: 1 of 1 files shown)


src/port1.0/portconfigure.tcl
```diff
@@ -340,25 +340,25 @@ proc portconfigure::configure_start {args} {
...
-        {^macports-openmpi-clang$}          {MacPorts Open MPI Wrapper for Xcode Clang}
-        {^macports-mpich-clang-(\d+\.\d+)$}
+        {^apple-gcc-(4\.[02])$}                   {MacPorts Apple GCC %s}
+        {^cc$}                                    {System cc}
+        {^clang$}                                 {Xcode Clang}
+        {^gcc$}                                   {System GCC}
+        {^gcc-(3\.3|4\.[02])$}                    {Xcode GCC %s}
+        {^llvm-gcc-4\.2$}                         {Xcode LLVM-GCC 4.2}
+        {^macports-clang$}                        {MacPorts Clang (port select)}
+        {^macports-clang-(\d+\.\d+|[1-9][0-9])$}  {MacPorts Clang %s}
```
jmroot, line 350: If I understand the intent correctly, the parenthetical expression would be better written as `(\d+(\.\d+)?)` or perhaps even `([\d.]+)`, as there's very little harm being more flexible than strictly necessary.

cjones051073: I considered something like that, but went with the above to enforce the fact that the 'dotless' clang versions start with 10. i.e. `macports-clang-10` is valid, but `macports-clang-9` is not.

cjones051073: i.e. I am not sure I agree that being more flexible with the expression is necessarily a good idea here.

jmroot: Well, I don't think that base should be enforcing what the names of the ports can be, beyond the bare minimum that it's "macports-clang" and something identifiable as a version. The downside of allowing more is minimal, since if you write something that matches this regex but doesn't correspond to an existing port you just get a "port not found" error instead of "Invalid value for configure.compiler".

cjones051073: OK, I made the change.

cjones051073:
@jmroot Thanks for your review. I'll give this a bit longer to allow others to comment, maybe a week or so, but then I think it can be merged.

PR #175, opened 2020-05-13
reclaim: remove portdbpath/build directory

kurthindenburg:
This removes the entire build folder which also removes the default
ccache location.

Closes: https://trac.macports.org/ticket/60126

### DIFF
--- a/src/macports1.0/reclaim.tcl
+++ b/src/macports1.0/reclaim.tcl
@@ -112,13 +112,22 @@ namespace eval reclaim {
         # The root build folder location
         set root_build      [file join ${macports::portdbpath} build]
 
-        try -pass_signal {
-            if {[macports::global_option_isset ports_dryrun]} {
-                ui_msg "Deleting... (dry run)"
-                ui_info [msgcat::mc "Skipping deletion of %s (dry run)" $root_build]
-            } else {
-                ui_info [msgcat::mc "Deleting %s" $root_build]
-                file delete -force $root_build
+        ui_msg "$macports::ui_prefix Build location: ${root_build}"
+
+        set retval 0
+        if {[info exists macports::ui_options(questions_yesno)]} {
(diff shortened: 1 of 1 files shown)


ryandesign:
> which also removes the default ccache location

Is that desirable? A user who has configured MacPorts to use ccache wants to use it to speed up their compilations. Deleting the ccache directory defeats that.

On the other hand, a user who enabled ccache and then later disabled it again might want MacPorts to delete the now-unused ccache directory.

kurthindenburg:
I think reclaim should delete the entire build including the. cache.   I'm open for other ways to handle it or perhaps ask the user.

ryandesign:
That's fine as a first step. To reduce surprise, we could find wherever we document MacPorts ccache support and mention there that reclaim will delete the directory.

src/macports1.0/reclaim.tcl
```diff
@@ -112,13 +112,22 @@ namespace eval reclaim {
...
-            if {[macports::global_option_isset ports_dryrun]} {
-                ui_msg "Deleting... (dry run)"
-                ui_info [msgcat::mc "Skipping deletion of %s (dry run)" $root_build]
-            } else {
-                ui_info [msgcat::mc "Deleting %s" $root_build]
-                file delete -force $root_build
+        ui_msg "$macports::ui_prefix Build location: ${root_build}"
+
+        set retval 0
+        if {[info exists macports::ui_options(questions_yesno)]} {
```
neverpanic, line 118: I realize we're wandering into make-a-wish land here, but should we only ask when ccache is enabled?

ryandesign:
I had forgotten that the ccache directory path can be changed in macports.conf. So users who want to reclaim their build directory but don't want to reclaim the ccache directory could just move the ccache directory out of the build directory.

kurthindenburg:
I did this manually  [b9fdb847990736cae8080dcd8f47b33a6fe1c6b3/macports-base]

PR #176, opened 2020-05-18
Move cltversion PG into base

MarcusCalhoun-Lopez:


jmroot:
Sorry, I forgot there was an open PR for this when I implemented d05ebde.

PR #177, opened 2020-05-20
Process ping output without using grep or cut

ryandesign:
Process ping output using Tcl rather than by spawning grep and cut processes.

This is one of several steps that may help alleviate https://trac.macports.org/ticket/60509.


ryandesign:
Doesn't that overwrite the earlier
```
                        set pingtimes($host) 5000
```

ryandesign:
No it doesn't because we're only looping over the hosts for which we could spawn the ping.

PR #178, opened 2020-05-27
reclaim: separate removing build and ccache locations

kurthindenburg:
Do not actually remove the build and ccache locations, only the files
and folders underneath.
A followup to the discussions in (#175)

See https://trac.macports.org/ticket/60126

### DIFF
--- a/src/macports1.0/reclaim.tcl
+++ b/src/macports1.0/reclaim.tcl
@@ -110,7 +110,7 @@ namespace eval reclaim {
         # Delete portdbpath/build directories
         global macports::portdbpath
 
-        # The root build folder location
+        # The root build directory location
         set root_build      [file join ${macports::portdbpath} build]
 
         ui_msg "$macports::ui_prefix Build location: ${root_build}"
@@ -127,7 +127,7 @@ namespace eval reclaim {
                     ui_info [msgcat::mc "Skipping deletion of %s (dry run)" $root_build]
                 } else {
                     ui_info [msgcat::mc "Deleting %s" $root_build]
-                    set builddirs [glob -nocomplain -types d [file join $root_build *]]
+                    set builddirs [glob -nocomplain -directory $root_build *]
                     catch {file delete -force -- {*}$builddirs}
                 }
             }
@@ -159,7 +159,7 @@ namespace eval reclaim {
                 }
             }
         } else {
-            ui_info [msgcat::mc "Skipping deletion of ccache folders %s not found." $macports::ccache_dir]
+            ui_info [msgcat::mc "Skipping deletion of ccache directory: %s does not exist." $macports::ccache_dir]
         }
     }
 
--- a/src/macports1.0/reclaim.tcl
+++ b/src/macports1.0/reclaim.tcl
@@ -128,7 +128,14 @@ namespace eval reclaim {
                 } else {
                     ui_info [msgcat::mc "Deleting %s" $root_build]
                     set builddirs [glob -nocomplain -directory $root_build *]
-                    catch {file delete -force -- {*}$builddirs}
+                    if {[llength $builddirs] > 0} {
+                        try -pass_signal {
+                            file delete -force -- {*}$builddirs
+                        } catch {{*} eCode eMessage} {
+                            ui_debug "$::errorInfo"
+                            ui_error "$eMessage"
+                        }
+                    }
                 }
             }
         }
@@ -154,7 +161,14 @@ namespace eval reclaim {
                     } else {
                         ui_info [msgcat::mc "Deleting %s" $macports::ccache_dir]
                         set ccachedirs [glob -nocomplain [file join $macports::ccache_dir *]]
-                        catch {file delete -force -- {*}$ccachedirs}
+                        if {[llength $ccachedirs] > 0} {
+                            try -pass_signal {
+                                file delete -force -- {*}$ccachedirs
+                            } catch {{*} eCode eMessage} {
+                                ui_debug "$::errorInfo"
+                                ui_error "$eMessage"
+                            }
+                        }
                     }
                 }
             }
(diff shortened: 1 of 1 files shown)


src/macports1.0/reclaim.tcl
```diff
@@ -116,7 +117,7 @@ namespace eval reclaim {
 
         set retval 0
         if {[info exists macports::ui_options(questions_yesno)]} {
-            set retval [$macports::ui_options(questions_yesno) "" "" "" "y" 0 "Would you like to delete the entire build directory (including the default ccache location)?"]
+            set retval [$macports::ui_options(questions_yesno) "" "" "" "y" 0 "Would you like to delete everything under the build directory (except the default ccache location)?"]
```
jmroot, line 120: "everything under the build directory" might be better phrased as "the contents of the build directory" or "all port build directories". The latter could remove the need to specify that the ccache dir is excepted.

src/macports1.0/reclaim.tcl
```diff
@@ -126,12 +127,42 @@ namespace eval reclaim {
                     ui_info [msgcat::mc "Skipping deletion of %s (dry run)" $root_build]
                 } else {
                     ui_info [msgcat::mc "Deleting %s" $root_build]
-                    file delete -force $root_build
+                    set builddirs [glob -nocomplain -types d [file join $root_build *]]
+                    catch {file delete -force -- {*}$builddirs}
```
jmroot, line 131: Is the catch just in case there are no build dirs and the glob thus matches nothing? Would it be worth reporting if the deletion fails for other reasons?

src/macports1.0/reclaim.tcl
```diff
@@ -126,12 +127,42 @@ namespace eval reclaim {
                     ui_info [msgcat::mc "Skipping deletion of %s (dry run)" $root_build]
                 } else {
                     ui_info [msgcat::mc "Deleting %s" $root_build]
-                    file delete -force $root_build
+                    set builddirs [glob -nocomplain -types d [file join $root_build *]]
```
jmroot, line 130: Probably better to use the `-directory` option and leave out the `file join`.

src/macports1.0/reclaim.tcl
```diff
@@ -126,12 +127,42 @@ namespace eval reclaim {
...
+                        ui_info [msgcat::mc "Skipping deletion of %s (dry run)" $macports::ccache_dir]
+                    } else {
+                        ui_info [msgcat::mc "Deleting %s" $macports::ccache_dir]
+                        set ccachedirs [glob -nocomplain [file join $macports::ccache_dir *]]
+                        catch {file delete -force -- {*}$ccachedirs}
+                    }
+                }
+            }
+        } else {
+            ui_info [msgcat::mc "Skipping deletion of ccache folders %s not found." $macports::ccache_dir]
```
jmroot, line 162: There should probably be a semicolon or something after "folders". Also we should probably pick either "folder" or "directory" and stick to it.

src/macports1.0/reclaim.tcl
kurthindenburg: yes, it will fail if there is nothing there; are there other reasons it might fail?

kurthindenburg: I see portclean checks for errors; it would have to ignore if there are no dirs there.

kurthindenburg: I copy/pasted the code that is in a lot of other places - I couldn't figure how to trigger it to test.

jmroot: Come to think of it, you could do the glob earlier and, if it matches nothing, report that there are no build dirs to delete and return immediately. That would avoid having to answer the interactive question in that case.

Error handling is still good to have. Look at the ERRORS section in `man 2 unlink` for a (probably not exhaustive) list of things that can go wrong.

src/macports1.0/reclaim.tcl
```diff
@@ -126,9 +127,53 @@ namespace eval reclaim {
...
+            }
+
+            if {${retval} == 0} {
+                try -pass_signal {
+                    if {[macports::global_option_isset ports_dryrun]} {
+                        ui_msg "Deleting... (dry run)"
+                        ui_info [msgcat::mc "Skipping deletion of %s (dry run)" $macports::ccache_dir]
+                    } else {
+                        ui_info [msgcat::mc "Deleting %s" $macports::ccache_dir]
+                        set ccachedirs [glob -nocomplain [file join $macports::ccache_dir *]]
```
jmroot, line 163: You can use -directory here too.

kurthindenburg: You are correct - I thought it would miss the ccache.conf file.  -directory doesn't work like I thought it would.

src/macports1.0/reclaim.tcl
kurthindenburg: Ok bypassing the question sounds good if there's nothing to delete.

kurthindenburg:
I think I've handled all your suggestions - I can't figure out why github says there's some one review outstanding.

src/macports1.0/reclaim.tcl
```diff
@@ -106,28 +107,77 @@ namespace eval reclaim {
...
 
-        # The root build folder location
+        # The root build directory location
         set root_build      [file join ${macports::portdbpath} build]
 
         ui_msg "$macports::ui_prefix Build location: ${root_build}"
 
+        if {[macports::global_option_isset ports_dryrun]} {
+            ui_msg "Deleting... (dry run)"
+            ui_info [msgcat::mc "Skipping deletion of %s (dry run)" $root_build]
```
jmroot, line 120: Nitpick: This and line 136 still make it sound like $root_build itself will be deleted.

kurthindenburg: fixed thanks

kurthindenburg:
[d08d261f93844cea87f602e1f295e8a0ddc83bde/macports-base] done thanks

PR #179, opened 2020-05-28
Fix c++17 compiler selection to blacklist incomplete mac10.13 versions

cjones051073:
Triggered by a discussion at

https://github.com/macports/macports-ports/pull/7243

The clang blacklisting is currently allowing some versions on 10.13 (Xcode 9/10) that are either missing std::optional or for which the implementation is incomplete or broken.

Consider the basic test application, using `std::optional`

```
 > cat test.cpp
#include <optional>
int main() {
  std::optional<int> i = 0;
  return i.value();
}
```

on macOS 10.13 with Xcode 9.4.1 this fails to build with

```
 > clang++ -std=c++17 ./test.cpp
./test.cpp:1:10: fatal error: 'optional' file not found
#include <optional>
         ^~~~~~~~~~
1 error generated.

 > clang++ --version
Apple LLVM version 9.1.0 (clang-902.0.39.2)
Target: x86_64-apple-darwin17.7.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

then, again on macOS 10.13 but with Xcode 10.1

```
  > clang++ -std=c++17 ./test.cpp 
./test.cpp:4:12: error: call to unavailable member function 'value': introduced in macOS 10.14
  return i.value();
         ~~^~~~~
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1/optional:942:27: note: candidate function has been
      explicitly made unavailable
    constexpr value_type& value() &
                          ^
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1/optional:933:33: note: candidate function has been
      explicitly made unavailable
    constexpr value_type const& value() const&
                                ^
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1/optional:951:28: note: candidate function not viable: no
      known conversion from 'optional<...>' to 'optional<...>' for object argument
    constexpr value_type&& value() &&
                           ^
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1/optional:960:34: note: candidate function not viable: no
      known conversion from 'optional<...>' to 'const optional<...>' for object argument
    constexpr value_type const&& value() const&&
                                 ^
1 error generated.

> clang++ --version
Apple LLVM version 10.0.0 (clang-1000.11.45.5)
Target: x86_64-apple-darwin17.7.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

Finally, on macOS 10.14, Xcode 10.3

```
MacVM1014 ~/Downloads > clang++ -std=c++17 ./test.cpp 
MacVM1014 ~/Downloads > ./a.out 
MacVM1014 ~/Downloads > 
```

As a compiler cannot be relied upon as a fully functional c++17 compiler if it fails to implement (a commonly used) part of that standard, this MR increases the versions blacklisted to exclude these versions.

### DIFF
--- a/src/port1.0/portconfigure.tcl
+++ b/src/port1.0/portconfigure.tcl
@@ -849,7 +849,7 @@ proc portconfigure::max_version {verA verB} {
 #| 1998 (C++98) |     -     |       -       |     -     |     -     |
 #| 2011 (C++11) |    3.3    |   500.2.75    |    5.0    |   4.8.1   |
 #| 2014 (C++14) |    3.4    |   602.0.49    |    6.3    |     5     |
-#| 2017 (C++17) |    5.0    |   902.0.39.1  |    9.3    |     7     |
+#| 2017 (C++17) |    5.0    |  1000.11.45.2 |   10.0    |     7     |
 #--------------------------------------------------------------------
 #
 # https://openmp.llvm.org
@@ -909,7 +909,7 @@ proc portconfigure::get_min_command_line {compiler} {
                 set min_value [max_version $min_value 500.2.75]
             }
             if {${compiler.cxx_standard} >= 2017} {
-                set min_value [max_version $min_value 902.0.39.1]
+                set min_value [max_version $min_value 1000.11.45.2]
             } elseif {${compiler.cxx_standard} >= 2014} {
                 set min_value [max_version $min_value 602.0.49]
             } elseif {${compiler.cxx_standard} >= 2011} {


ryandesign:
I'll make the same comment here that I made in macports/macports-ports#7243: Did you actually verify that this works on macOS 10.13, or did you only test on macOS 10.14? Because I thought the issue with `std::optional` was that it requires changes in libc++, changes which are in macOS 10.14 but not in 10.13 and earlier. That's why the mkvtoolnix port degrades to an earlier version on macOS 10.13 and earlier.

cjones051073:
> I'll make the same comment here that I made in [macports/macports-ports#7243](https://github.com/macports/macports-ports/pull/7243): Did you actually verify that this works on macOS 10.13, or did you only test on macOS 10.14? Because I thought the issue with `std::optional` was that it requires changes in libc++, changes which are in macOS 10.14 but not in 10.13 and earlier. That's why the mkvtoolnix port degrades to an earlier version on macOS 10.13 and earlier.

Yes, I should have clarified what happens with the macports clang compilers.

Yes, it does work, but not with all clang versions. i.e. 5.0 to 7.0 are all just fine.

```
Titan ~/cernbox/MacPorts/c++17 > clang++-mp-7.0 -std=c++17 ./test.cpp 
Titan ~/cernbox/MacPorts/c++17 > 
```

however 8.0 and newer fail with similar messages to the Xcode clang.

So yes, we also should blacklist macports-clang 8.0 and newer on macOS10.13 and older, when c++17 is requested.

cjones051073:
Just to illustrate the point a bit more, the `mkvtoolnix` port will actually build the latest 46.0.0 version on macOS10.13 just fine. You just need to add an additional blacklist to avoid the broken compilers

```
compiler.blacklist-append {clang < 1001.0.46.4} {macports-clang-[8-9].0}
```

with that, the build fallbacks to using macports clang 7.0, which works just fine. e.g. from the configure step

```
checking for support for C++17 feature "attribute 'maybe_unused'"... yes
checking for support for C++17 feature "nested namespace definition"... yes
checking for support for C++17 feature "structured bindings"... yes
checking for support for C++17 feature "std::optional"... yes
```

the build goes through just fine, and I can start the gui etc.

This MR just adds by default to base the `clang` blacklist when c++17 is requested. A second MR will be needed against 

https://github.com/macports/macports-ports/blob/master/_resources/port1.0/compilers/clang_compilers.tcl

to just provide a minor tweak to the logic, to also veto macports clang 8.0 or newer, on macOS 10.13 or older, if c++17 is requested.

with these two changes, just setting the c++ standard will be enough, which I think should ultimately be the aim here.

jmroot:
> however 8.0 and newer fail with similar messages to the Xcode clang.
> 
> So yes, we also should blacklist macports-clang 8.0 and newer on macOS10.13 and older, when c++17 is requested.

Really? That's awful.

cjones051073:
> > however 8.0 and newer fail with similar messages to the Xcode clang.
> > So yes, we also should blacklist macports-clang 8.0 and newer on macOS10.13 and older, when c++17 is requested.
> 
> Really? That's awful.

Awful it’s necessary or awful I am proposing it ?

jmroot:
> Awful it’s necessary or awful I am proposing it ?

That newer clangs are broken.

cjones051073:
> > Awful it’s necessary or awful I am proposing it ?
> 
> That newer clangs are broken.

I agree, but its inevitable I guess that newer clangs will eventually drop older systems, although this seems more like a specific bug. @kencu any ideas

kencu:
I presume this is related to the new  __config Availability tests, that I disable only on 10.6 and less because we have a newer libc++ there: <https://github.com/macports/macports-ports/blob/master/lang/llvm-9.0/files/5005-MacPorts-only-patch-libcxx-includes-disable-availability-tests.diff>.

It is trivial to build and install a newer libc++ on all systems, btw. How exactly we would use it is more complicated. wrapper with dyly_library_path set to it is the only useful idead, I suppose.

kencu:
makes no sense that the libc++ is OK when using clang-7.0 but not when using clang-8.0, though.

kencu:
one more: as of llvm 8, there was an ABI change in libc++ . See <https://trac.macports.org/ticket/57576>. This might be relevant as well.

kencu:
to be considered: requiring our "compiler.cxx_standard 2017" to only point to a compiler that has complete and total 100% support for the entire c++17 standard is a pretty obtuse criteria for the compilers that may actually be capable of building a port. I don't think any clangs quailfy yet.

see <https://clang.llvm.org/cxx_status.html#cxx17> and <https://libcxx.llvm.org/cxx1z_status.html> 

Our old method of blacklisting failing compilers would likely be better than this.

cjones051073:
I agree whether 100% coverage is feasible, is a good question. No compiler has absolute perfect support, but I would argue some parts are more important that others, and std::optional is IMHO one of the parts of the standard more likely to be used. Its one of the reasons we now target this standard where I work, and I am pretty sure I have run into the issue that triggered this with a few other ports. Its just this time I decided to look into it. So I would still maintain tweaking the blacklist to fix this is a good idea.

I also agree its a bit odd how things break for 8+ but are fine for 7, but for whatever the reason it seems to be the case. Luckily the macports clang blacklisting is part of the ports tree, not base, so much easier to fine tune again later on if things can be improved. Base is where the Xcode clangs need blacklisting.

kencu:
I think we should modify the clang/llvm-10+ ports to install their libcxx/libcxxabi somewhere, like we do with libgcc. Then we can figure out how to use it -- there is already wrapper technology in legacysupport 1.1 that could be leveraged...

ryandesign:
> the `mkvtoolnix` port will actually build the latest 46.0.0 version on macOS10.13 just fine. You just need to add an additional blacklist to avoid the broken compilers

Thank you for finding that out. That does work, I'm just really baffled as to why. On macOS 10.13.6 with Apple clang 9.1.0 from Xcode 9.4.1, the `std::optional` test fails with:

```
conftest.cpp:25:10: fatal error: 'optional' file not found
```

With MacPorts clang 9.0.1 or 10.0.0 it fails with:

```
conftest.cpp:32:20: error: 'value' is unavailable: introduced in macOS 10.14
```

With MacPorts clang 8.0.1 it fails with:

```
conftest.cpp:32:20: error: call to unavailable member function 'value': introduced in macOS 10.14
```

And with MacPorts clang 7.1.0 it builds and launches with no problem. That makes it sound like it's not a libc++ issue after all.

kencu:
>  That makes it sound like it's not a libc++ issue after all.

It's not libc++, exactly -- it's Apple's availability tests for libc++ that have been added to newer clangs. It builds fine with new clangs if you turn that off, as we sorted out with mkvtoolnix three months ago <https://trac.macports.org/ticket/60148#comment:4>

kencu:
of course, those libc++ availability tests were added for a reason. So even though mkvtoolnix appears to work, the only resilient fix is to upgrade libc++ and then disable the availability tests (like I have done above on 10.6, which builds mkvtoolnix just fine using clang-9.0).

kencu:
Here is a definitive bit about std::optional::value on macos from the Apple libc++ lead, and why it's only available on 10.14 and newer. <https://twitter.com/LouisDionne/status/1184807412483723265> . In fact, it seems the fact that clang-7.0 compiles it without error would probably be an error in clang-7.0, but you have to actually trigger the exception to throw to see what happens in real life on an older system.

kencu:
Funny thing -- bad_optional_access is supposed to be available on 10.14 according to the clang-9.0 _config header:
```
#  define _LIBCPP_AVAILABILITY_BAD_OPTIONAL_ACCESS                             \
     __attribute__((availability(macosx,strict,introduced=10.14)))             \
     __attribute__((availability(ios,strict,introduced=12.0)))                 \
     __attribute__((availability(tvos,strict,introduced=12.0)))                \
     __attribute__((availability(watchos,strict,introduced=5.0)))
```
but on 10.13, there it is:
```
$ nm /usr/lib/libc++.1.dylib | grep optional
0000000000014fc6 T __ZNKSt19bad_optional_access4whatEv
0000000000014fe8 T __ZNSt12experimental19bad_optional_accessD0Ev
0000000000014fde T __ZNSt12experimental19bad_optional_accessD1Ev
0000000000014fd4 T __ZNSt12experimental19bad_optional_accessD2Ev
0000000000014faa T __ZNSt19bad_optional_accessD0Ev
0000000000014fa0 T __ZNSt19bad_optional_accessD1Ev
0000000000014f96 T __ZNSt19bad_optional_accessD2Ev
0000000000057750 S __ZTINSt12experimental19bad_optional_accessE
0000000000057790 S __ZTISt19bad_optional_access
0000000000051850 S __ZTSNSt12experimental19bad_optional_accessE
0000000000051880 S __ZTSSt19bad_optional_access
0000000000057720 S __ZTVNSt12experimental19bad_optional_accessE
0000000000057768 S __ZTVSt19bad_optional_access

$ c++filt  __ZNSt19bad_optional_accessD0Ev
std::bad_optional_access::~bad_optional_access()
```

kencu:
Issue has been taken upstream for opinions <http://lists.llvm.org/pipermail/libcxx-dev/2020-May/000815.html>

cjones051073:
Many thanks @kencu Your analysis makes sense to me, it does seem like support was added to the 10.13 libc++ at some point, and  the checks in clang just haven't caught up. But lets see what upstream says. In the meantime I think patching our clang 8+ builds to fix the check to allow 10.13 is probably a good idea. Then https://github.com/macports/macports-ports/pull/7263 will not be required.

kencu:
That patch should fix this issue on clang-devel. Once we confirm it works as expected, I'll backport it into clang-8.0 through clang-10. Upstream has not responded as yet.

kencu:
I'm trying to hold off on pushing this fix through until we need to revbump clang for some other reason...hundreds of MB to download for everyone for what amounts to a single digit change in one header to affect a niche case...there's a revbump coming for clang for libffi's update, and it would be nice to coordinate it with that.

cjones051073:
seems very reasonable.

kencu:
and <https://github.com/macports/macports-ports/commit/6ce821918ccbd829438a95b6efa72f7bbad76a99> should about do it.

I think you can close this @cjones051073 , and there is another somewhere...

ryandesign:
Don't we still need the change in this PR to blacklist the Xcode clang versions that don't have the fix you added?

We also should do something for OS X 10.7–10.12. No amount of compiler blacklisting will provide compatibility with this C++17 feature there, unless libc++ is upgraded, which we have no supported way to do yet.

Looks like there wasn't any response yet to your question on the llvm mailing list.

kencu:
Yes, you're right about the Xcode clangs -- they are broken, right up to the current Xcode on 10.13. So they are "pretty good" at c++17 but not 100%.

I personally would take those failures as they come and blacklist ports that fail. It is not very many ports that even use optional.value at present to trigger this error. Being overly dogmatic about the purity of c++17 support a compiler offers could tie our hands pretty solidly at some point for bootstrapping. But some have strong feelings about this.

There is something else that might still be done for the older libc++.dylibs that don't have the bad_optional_access symbol -- look at this patch that was in WebKit2 for the longest time <https://github.com/macports/macports-ports/blob/master/www/webkit2-gtk-devel/files/unused/patch_bad_optional_access.diff> -- that I had to pull out at a certain point. It looks like we might consider adding something like that replacement function to our <optional> header on systems < 10.13....

ryandesign:
I thought someone else had told me that "optional" was a more popular C++17 feature that everyone was going to be using. But if it's actually not used that often then I agree with you that ports can additionally blacklist non-supporting implementations manually. However we should have a recipe in PortfileRecipes that explains how to do so correctly so that we have consistency.

I can't say anything about the webkit2 patch you referenced. I don't know anything about that.

kencu:
could be that optional is popular. As I understand it, using that "optional.value" construct is the thing that causes trouble, because you need the custom handler for the exception.

The patch inlines that handler into the header statically.

Or hey -- yet another thing to go into legacysupport :>. Override <optional> for the definitions, and add in the handlers in legacysupport. ... The hits just keep on coming :>

cjones051073:
Personally, I would still include the updated blacklisting here. Yes, we could do it on a case by case basis as ports that need optional fail, but that requires people to spot those failures, and manually fix them. The changes here do it automatically. The only 'expense' is we use Macports clang by default on 10.13. I'm have no problem with this.

kencu:
there are other, more serious troubles, though. 

for example, with that plan 10.13 is likely to run into bootstrapping loops much sooner. the current llvm toolchain with libtapi etc requires some c++17 to build, but does not exercise the complete spec. If you blacklist 10.13's xcode, all of a sudden you're in a much trickier mess than you need to be in, as 10.13's Xcode builds it all just fine.

Those kinds of issues become more common. But there are advantages too, in less interventions, which I have advocated for in the past.

kencu:
upstream committed a similar fix <https://github.com/llvm/llvm-project/commit/7fb40e1569dd66292b647f4501b85517e9247953#diff-a020123e9cf2d0b9d0d6f0ca236b6521>

ryandesign:
Ok so shall we merge this PR then?

Ken, will you take care of updating our llvm patch to match upstream's?

ryandesign:
Also Ken, macports/macports-ports@6ce8219 didn't bump the revisions, so this change is not yet available to users. I believe you were waiting to do a revbump of the llvms and clangs along with the libffi 3.3 update. I'm not sure what we're waiting on there: macports/macports-ports/pull/6971

kencu:
There is a very simple patch from the webkit project that would likely make c++17 work for a number of older libc++ implementations. We would stick the patch in ```<optional.h>``` or similar, and then disable the test.

92fca5b526a7ff65bbff5763b5d7737712abef69/macports-ports

kencu:
my older browser won't let me edit that link -- I'll fix it up later -- but that's the commit with the (simple) patch. It might be a lot easier to use that patch than to figure out how to provide a new libc++ to all the 10.7 to 10.12 systems (although sooner or later I will have to figure out how to do that).

ryandesign:
I fixed the link. I don't understand the patch so I'll have to leave it to someone else to decide.

kencu:
Here is my understanding of this:

newer clangs support std::optional on all systems.

std::optional::value, which is at present occasionally used, requires a function to process the exception generation when there is a bad optional access call.

This function was patched into the std::optional headers when it was still under experimental, but the Apple part of llvm team didn't include that patch in the libc++ release, and instead compiled it into libc++, and marked it as unavailable before 10.13 (now that we fixed it).

If we patched the bad optional access function back into the header on older systems, and then disabled at least that particular test in libc++'s ```__config``` then all the systems, probably back to 10.7, could use std::optional with full support. 10.4 to 10.6 already work, because they have a newer libc++, and I patch out the config test.

That needs a bit of testing on 10.7 through 10.12, but it is very likely to work.

And then -- this PR would probably not be needed.

kencu:
So as it is right now, all of std::optional works with clang-9.0 back to 10.7, except for the one missing bit with std::optional::value, that has to thow to bad_optional_access() if there is an error.

This little patch:
```
$ sudo cp optional optional.orig
Password:
Kens-107-VM-Parallels:v1 cunningh$ diff -u optional.orig optional
--- optional.orig	2020-07-04 18:51:05.000000000 -0700
+++ optional	2020-07-04 18:51:13.000000000 -0700
@@ -168,13 +168,10 @@
 namespace std  // purposefully not using versioning namespace
 {
 
-class _LIBCPP_EXCEPTION_ABI _LIBCPP_AVAILABILITY_BAD_OPTIONAL_ACCESS bad_optional_access
-    : public exception
-{
+// 20.5.8, class bad_optional_access
+class bad_optional_access : public logic_error {
 public:
-    // Get the key function ~bad_optional_access() into the dylib
-    virtual ~bad_optional_access() _NOEXCEPT;
-    virtual const char* what() const _NOEXCEPT;
+  explicit bad_optional_access() : logic_error{"bad optional access"} {}
 };
 
 }  // std
```
and disabling apple's availability tests in ```__config``` like this:
```
#define _LIBCPP_DISABLE_AVAILABILITY
```

and all the test files of std::optional and std::optional::value I can throw at it work fine.

kencu:
finally officially fixed by 8ffd5598ef15de3d4b260693ce552ef762658efe

ryandesign:
> finally officially fixed by [macports/macports-ports@8ffd559](https://github.com/macports/macports-ports/commit/8ffd5598ef15de3d4b260693ce552ef762658efe)

Thanks, that fixed clang-10. https://github.com/macports/macports-ports/commit/aefdb6c4596db427f5ee26ba23a77d37373ee095 fixed clang-9.0 and https://github.com/macports/macports-ports/commit/23add25b0a9c9a3112bc467bb5a6785043b69c0b fixed clang-8.0.

Looks like clang-devel still has the early version of the patch. Maybe llvm-devel could be updated to a newer upstream version that includes the final fix; then the patch can be removed.

kencu:
You don't think llvm-devel will pick it up from upstream's commit?

ryandesign:
...Yes, I do. That's why I'm recommending llvm-devel be updated to a newer commit so that our local patch can be removed.

kencu:
Yeah, you replied before I could delete that comment. You’re quite right, there were indeed two other commits for other versions of clang, and clang-devel will pick up the patch from upstream. I’ll have to see if the presently-imaginary clang-11 will need the patch, or whether the fork was late enough and / or updated to puck up the upstream fix, or whether it will need an updated patch.

chrstphrchvz:
It might be worth mentioning here that support for `<optional>` was [introduced in Xcode 10 according to its release notes](https://developer.apple.com/documentation/xcode-release-notes/xcode-10-release-notes).

>## Apple Clang Compiler
>### New Features
>* Xcode 10 adds support for the C++17 headers `<any>`, `<optional>`, and `<variant>`. (39271859)

cjones051073:
reopening this, as I just ran into it, again....

https://build.macports.org/builders/ports-10.13_x86_64-builder/builds/110131/steps/install-port/logs/stdio

the fix above to macports clang does not help when the base selection still allows the broken Xcode clangs to be used when c++17 is requested, so I still maintain this change should be made.

kencu:
So we established that the system libc++ in 10.13 has all the needed functions, but the libc++ headers wrongly thought they didn't have it. We fixed that in all our macports-clangs that matter, and moved the fix upstream so Xcode has also picked it up.

I have not verified which released version of Xcode has the fixed headers, but on 10.13, any Xcode without the corrected headers will continue to show this error.

cjones051073:
> So we established that the system libc++ in 10.13 has all the needed functions, but the libc++ headers wrongly thought they didn't have it. We fixed that in all our macports-clangs that matter, and moved the fix upstream so Xcode has also picked it up.
> 
> I have not verified which released version of Xcode has the fixed headers, but on 10.13, any Xcode without the corrected headers will continue to show this error.

https://trac.macports.org/wiki/XcodeVersionInfo#macOS10.13

It is fixed with Xcode 10(.1). My 10.13 VM is using Xcode 10.1 / clang 1000.11.45.5 and does not show the problem.

The build bot is using Xcode 9.4.1 and has the issue.

The only thing I am not sure on is if Xcode 10 (not 10.1) fixes it as well. I suspect it does.

chrstphrchvz:
I would also note that the [Xcode 10.1 release notes](https://developer.apple.com/documentation/xcode-release-notes/xcode-10_1-release-notes) do not mention anything about `<optional>`.

cjones051073:
I rolled my VM back to Xcode 10 (clang 1000.11.45.2) and can confirm that is also OK, so I have updated this MR accordingly. Interestingly this means the c++17 requirement is then the same as C17, which I doubt is a coincidence.

Going to merge, as I really see no reason not to, we have discussed this enough...

PR #180, opened 2020-05-30
Set macosx_version to the full macOS version (e.g. 10.15.5)

kurthindenburg:
Previously macosx_version was set to the major version (e.g. 10.5). A
new variable macosx_major_version is added to avoid numerious splits
and joins.
Any checks using vercmp will still work.

Closes: https://trac.macports.org/ticket/56252

### DIFF
--- a/src/macports1.0/diagnose.tcl
+++ b/src/macports1.0/diagnose.tcl
@@ -91,7 +91,7 @@ namespace eval diagnose {
 
         array set config_options    [list]
         set parser_options          [list macports_location profile_path shell_location \
-                                    xcode_version_${macports::macosx_version} xcode_build]
+                                    xcode_version_${macports::macosx_major_version} xcode_build]
 
         set user_config_path        "${macports::autoconf::macports_conf_path}/port_diagnose.ini"
         set xcode_config_path       [macports::getdefaultportresourcepath "macports1.0/xcode_versions.ini"]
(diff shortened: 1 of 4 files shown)


kurthindenburg:
Putting this up for comments on if this is the way to proceed.  There are a number of tests that set macosx_version that will need to be investigated.

jmroot:
Adding a new variable is fine, but changing the semantics of an existing one is an API break and should be avoided if at all possible. I don't think there's a compelling reason to break API here.

kurthindenburg:
So you'd rather have macosx_version stay the same and add a new variable macosx_full_version?

mojca:
I also prefer the other one

ryandesign:
What's the compelling reason for not fixing how the existing variable works? Why perpetuate the mistake of having a variable called macosx_version that does not contain the macOS version but only a part of it? What use case are you envisioning where changing macosx_version to the full version would break something? I already went through the ports tree years ago to fix ports so that they would use macosx_version in a way that would work if it were the full version. Looks like this PR has gone through base to do the same there, or rather introduces a new macosx_major_version variable and uses that instead, which at least makes sense and names the variables in a way that clearly conveys what they contain.

We have previously made breaking changes without apparently caring about API stability. For example, platform blocks were changed so that they executed immediately (like an if statement) rather than deferred (like a variant) which broke several ports. We removed the cd command, requiring ports to adapt to that. We've renamed variables (livecheck.type used to be called livecheck.check). And quoting of env variables was changed, requiring some compatibility code to be added to portfiles for a time but that worked out fine.

PR #181, opened 2020-05-31
Add new variable to contain full macOS version (e.g. 10.15.5)

kurthindenburg:
This will allow the full version to be displayed in the log files.

Closes: https://trac.macports.org/ticket/56252


kurthindenburg:
See #180 for other discussion

src/macports1.0/macports.tcl
```diff
@@ -711,7 +713,8 @@ proc mportinit {{up_ui_options {}} {up_options {}} {up_variations {}}} {
             set os_subplatform macosx
             if {[file executable /usr/bin/sw_vers]} {
                 try -pass_signal {
-                    set macosx_version [exec /usr/bin/sw_vers -productVersion | cut -f1,2 -d.]
+                    set macosx_full_version [exec /usr/bin/sw_vers -productVersion]
+                    set macosx_version [join [lrange [split $macosx_full_version "."] 0 end-1] "."]
```
neverpanic, line 717: Should macOS ever start using four-tupled version numbers, this is a change of behavior. I'd prefer

```suggestion
                    set macosx_version [join [lrange [split $macosx_full_version "."] 0 1] "."]
```

jmroot:
If it would be helpful, you could also add variables like `macosx_version_major` and `macosx_version_minor` which would be e.g. 14 and 6 respectively if macosx_version_full is 10.14.6.

ryandesign:
`[macports_version]` was implemented as a proc rather than an option/variable. When I questioned this, I think it was Clemens who said it made more sense for it to be a proc since it should be read-only. You wouldn't want someone trying to set `macports_version` to some other value. By that same logic, should `macosx_full_version` be a proc too instead of an option/variable?

Tcl supports having a proc and a variable of the same name, so the proc could be named `[macosx_version]` the same as the old variable, though that might be confusing. Since the OS is called macOS now, it could be called `[macos_version]`.

If it were a proc, it could accept args like "full" or "major" or "minor". But a "major" value of "14" for macOS 10.14 would no longer be a good idea, now that macOS 11 has been announced. In fact, macOS 11 probably makes our existing `$macosx_version` variable's value incorrect as well, since on macOS 11 it would report the version as `11.0` when it should be `11` (if I understand Apple's new versioning scheme correctly).

On the other hand, Tcl does support read-only variables, so I don't know why that couldn't have been used for `macports_version`.

My suggestion would be that the new read-only variable or proc should be `macos_version`. I don't think it's burdensome for ports that want to examine its value to use `vercmp` or string manipulation so I wouldn't bother with major or minor variables. Once this is released we'd deprecate `macosx_version` and transition all remaining uses of it to the new variable or proc and later we can remove `macosx_version`.

jmroot:
I don't think the variable/constant thing is a significant issue, so far we've generally made things that are supposed to be read-only variables but not options, and that seems to work pretty well.

I agree we might as well use the opportunity to switch to "macos" rather than "macosx". So add a new variable `macos_version` which contains the full version, and another new variable `macos_version_major` that contains the same value as the existing `macosx_version`. Then just update all the documentation and examples to use the new variables, add a lint warning for using `macosx_version`, but at least for the near future keep it available for compatibility.

ryandesign:
Now there are 63 commits in this PR...

kurthindenburg:
lol yea I messed up - I plan on creating a new MR

mojca:
You can force push correct contents into this branch (as long as you don't accidentally push to master).

PR #182, opened 2020-06-01
Fix machinsta1.0 test by removing ', reexport' from the otool output

kurthindenburg:
In Catalina 10.15, the output of otool changed.
This fixes the machista1.0 test 'parsing libSystem.B.dylib'.

Library current version mismatch. Expected `83.0.0, reexport', was
    `83.0.0'

10.13 example:
/usr/lib/system/libxpc.dylib (compatibility version 1.0.0,
    current version 1205.70.15)

10.15 example:
/usr/lib/system/libxpc.dylib (compatibility version 1.0.0,
    current version 1738.120.8, reexport)

Tested in 10.13 and 10.15.

Closes: https://trac.macports.org/ticket/60580

### DIFF
--- a/src/machista1.0/tests/libmachista-test.c
+++ b/src/machista1.0/tests/libmachista-test.c
@@ -174,18 +174,23 @@ static bool compare_to_otool_output(char *path, const macho_t *ref) {
 			char lib_path[_POSIX_PATH_MAX];
 			char lib_comp_version[256];
 			char lib_curr_version[256];
+			char lib_curr_version_tmp[256];
 
 			// read loadcommand output line from otool
-			if (3 != fscanf(tmpf, "%*[\n]%*[\t]%255s (compatibility version %255[^,], current version %255[^)]))", lib_path, lib_comp_version, lib_curr_version)) {
+			if (3 != fscanf(tmpf, "%*[\n]%*[\t]%255s (compatibility version %255[^,], current version %255[^)]))", lib_path, lib_comp_version, lib_curr_version_tmp)) {
 				// error out silently, probably been the last line
 				break;
 			}
 
 			nullterminate(lib_path);
 			nullterminate(lib_comp_version);
-			nullterminate(lib_curr_version);
+			nullterminate(lib_curr_version_tmp);
 
-			//printf("\t\t%s, %s, %s\n", lib_path, lib_comp_version, lib_curr_version);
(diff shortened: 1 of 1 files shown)


kurthindenburg:
I could not figure how to just change the fscanf regex so I took this route.   I don't have a 10.14 system to check but don't think this test fails on it.

jmroot:
I think this should do it, though I can't test on 10.15.
```diff
--- a/src/machista1.0/tests/libmachista-test.c
+++ b/src/machista1.0/tests/libmachista-test.c
@@ -138,7 +138,7 @@ static bool compare_to_otool_output(char *path, const macho_t *ref) {
 		}
 
 		// get install name
-		if (3 != fscanf(tmpf, "%*[\n]%*[\t]%255s (compatibility version %255[^,], current version %255[^)])", install_name, compatibility_version, current_version)) {
+		if (3 != fscanf(tmpf, "%*[\n]%*[\t]%255s (compatibility version %255[^,], current version %255[^),]", install_name, compatibility_version, current_version)) {
 			fprintf(stderr, "\tError getting install name from otool output\n");
 			goto error_out;
 		}
```

kurthindenburg:
Yea I thought that should work as well but it doesn't.

	otool didn't return all libs found by libmachista. Was: 0, expected 34

jmroot:
Ah of course, because it keeps reading from tmpf afterwards. I'll try again later, unless @neverpanic already knows how to do it?

neverpanic:
I don't think that's possible with a single `fscanf(3)` invocation. The only other alternative I see is reading the line and then testing it against a regex, but that's probably even more effort.

The current solution lg2m? Considering that the CI on this repo still seems to be on 10.13, I guess we can merged this if you can confirm it works on 10.15?

kurthindenburg:
yes it works on 10.13 and 10.15 here

jmroot:
OK, I got it. The necessary format string is: `"%*[\n]%*[\t]%255s (compatibility version %255[^,], current version %255[^),]%*[^\n]"`. Could someone test that it works on 10.15?

kurthindenburg:
It fails


+               if (3 != fscanf(tmpf, "%*[\n]%*[\t]%255s (compatibility version %255[^,], current version %255[^),]%*[^\n]", install_name, compatibility_version, current_version)) {
</code>

	Library current version mismatch. Expected `83.0.0, reexport', was `83.0.0'
	Process was terminated with non-zero return value 1: Error parsing libSystem.B.dylib

PR #183, opened 2020-06-02
portconfigure.tcl - set sdk to "/" if that's what we're using

kencu:
leaving it as an empty string causes each individual port
to check whether it is empty and set it to "/" if so

closes: https://trac.macports.org/ticket/59798

### DIFF
--- a/src/port1.0/portconfigure.tcl
+++ b/src/port1.0/portconfigure.tcl
@@ -483,7 +483,7 @@ proc portconfigure::configure_get_sdkroot {sdk_version} {
 
     # Use the DevSDK (eg: /usr/include) if present and the requested SDK version matches the host version
     if {$sdk_version eq $macosx_version && [file exists /usr/include]} {
-        return {}
+        return {"/"}
     }
 
     set cltpath /Library/Developer/CommandLineTools


jmroot:
This breaks usage to construct paths like `${configure.sdkroot}/usr/include/foo.h`. I would suggest a new variable to avoid breaking API.

kencu:
Ah -- I knew there had to be some reason this was being left as an empty string, but I could not see what it was :>

kencu:
it seemed to be working -- I guess the OS accepts that construction, at least sometimes :
```
$ ls -la //usr/lib/libc++.dylib
lrwxr-xr-x  1 root  wheel  14 25 Sep  2018 //usr/lib/libc++.dylib -> libc++.1.dylib
```
And almost every portfile does set it to "/" if it's empty, I notice -- must be dozens of them doing that -- almost every portfile that uses configure.sdkroot has to do it.

I will see if I can find where a new variable might happily exist.

jmroot:
Don't forget this is used in base as well.

jmroot:
> And almost every portfile does set it to "/" if it's empty, I notice -- must be dozens of them doing that -- almost every portfile that uses configure.sdkroot has to do it.

I just had a look and it seems like there are more that don't set it to / when it's empty.

kencu:
I will admit never actually counting them -- just fixing the broken ones. 

I actually could not find a lot of uses of this in base.


So -- close this I guess, then, and we can fix it later on when Marcus gets around to it? I don't see a lot of point to adding another variable, as it is configure.sdkroot that we are having trouble with.

kencu:
I might as well leave this open, as it needs to be fixed.

PR #184, opened 2020-06-06
HTTP selfupdate

neverpanic:
*Note*: A few things still need to happen before this can be merged:

- [ ] Generate and upload signatures for the current release to GitHub and distfiles
- [ ] Possibly adjust CI to generate the new signature format
- [ ] Adjust the releasing guide to cover the new steps required
- [ ] Add a changelog entry
- [ ] Test signify compilation on older macOS (help wanted!)
- [ ] Commit `_resources/port1.0/fetch/base_mirror_sites.list`

I'm uploading this so you can have a look and see if you notice any issues with my approach. Note that this should be reviewed commit-by-commit – otherwise you'll get a huge patchset that's not simple to review.

# Description
Refactor the selfupdate package into smaller functions and switch to a control flow of:

1. checking for newer versions by downloading a single URL
2. downloading the new version using HTTP when available
3. verifying the signature using EdDSA with ed25519 elliptic curve crypto with the help of OpenBSD's signify(1)
4. automatically re-executing selfupdate for the sync step with the new base version when base was updated

This allows us to automatically fall back to other mirrors should our main mirror not be available. We did not have this functionality available for our previous rsync-based mechanism.

Additionally, the use of our standard curl wrapper enables displaying a progress bar for the download and automatically uses any proxies that might already be configured for HTTP.

Since we currently do not have a standard mechanism to sign the source code tarballs uploaded to our distfiles server and github releases page (we only sign the tarball pushed to rsync), I've taken the liberty to introduce a new modern signature scheme for this channel – if we have to touch this anyway, we might as well use modern tools and algorithms.

Previously, we would instruct users to re-run selfupdate manually when MacPorts base was updated, because Portfiles might already require the new base version to index correctly. This step has now been automated using an idea that lives on the migration branch, where the port.tcl client automatically re-executes itself in this situation.

This patch also adds support for a list of mirrors specified in the ports tree at _resources/port1.0/fetch/base_mirror_sites.list, so that we can add and remove mirrors without releasing a new version of MacPorts base. Base itself comes preconfigured with the GitHub releases URL if this list is not available.

See: https://trac.macports.org/ticket/60608


src/macports1.0/macports.tcl
```diff
@@ -775,11 +882,29 @@ proc mportinit {{up_ui_options {}} {up_options {}} {up_variations {}}} {
...
+            # specifying a line multiple times to specify a list, but still
+            # allows starting from scratch in the next configuration file
+            array set options_in_file {}
             while {[gets $fd line] >= 0} {
                 if {[regexp $conf_option_re $line match option ignore val] == 1} {
                     if {$option in $bootstrap_options} {
+                        # standard single-value options
                         set macports::$option [string trim $val]
                         global macports::$option
+                    } elseif {$option in $bootstrap_multivalue_options} {
```
jmroot, line 897: I'm not terribly happy about having two different behaviours for options in the config file. Is it just to avoid long lines?

vendor/signify-osx/src/CVS/Entries
```diff
@@ -0,0 +1 @@
+D
```
jmroot, line 1: Might want to add the CVS stuff to .gitignore.

jmroot (commented):
This all looks pretty reasonable for the most part. Two issues are:
 * The scope of this PR is quite large, covering multiple features: switching selfupdate from rsync to HTTPS, adding signify and using it for signature verification, and re-executing after updating base to continue with syncing the ports tree. Some of it can't be merged yet. It might be better to split it up so some can be merged sooner?
 * The new selfupdate method is going to be very difficult to adequately test without doing an actual release. I think it would be safer to start by adding the HTTPS method but use rsync by default for the first release, so developers and interested users can try out HTTPS selfupdate under real world conditions, but everyone else doesn't break if there are bugs. Then for the next release, make HTTPS the default, then if all goes well, remove rsync in the next release.

src/macports1.0/macports.tcl
neverpanic: Yes. Do you have a better suggestion on how to achieve this?

vendor/signify-osx/src/CVS/Entries
neverpanic: Good idea, yes.

neverpanic (commented):
We could merge everything up to and including "macports1.0: Sort and format list of options", and merge the rest later?

I like the idea of having this available selectively first, but I fear it would not really give us good test coverage unless we make it the default and fall back to rsync if it fails. I can take a look at implementing that, if you want?

jmroot:
> We could merge everything up to and including "macports1.0: Sort and format list of options", and merge the rest later?

Sounds good. Remember to add the .gitignore changes first.

> I like the idea of having this available selectively first, but I fear it would not really give us good test coverage unless we make it the default and fall back to rsync if it fails. I can take a look at implementing that, if you want?

That would be even better I guess. Probably still have a command-line option to make it use rsync, just in case.

neverpanic:
This needs spitting into smaller parts, and some refactoring to keep the rsync selfupdate as fallback. I haven't found the time to implement that yet, unfortunately.

neverpanic:
> @neverpanic Would it be OK if I start merging parts of this?

Yes!

pmetzger:
BTW, maybe this is a stupid question, but why don't we just switch to using Git off of Github for updates? It's fast, it's efficient, we already depend on Github.

ryandesign:
I'm not sure what reasons Clemens had for that, so I'll give my own reasons:

Using git is slower and takes more bandwidth and disk space than just downloading a release tarball. I haven't verified this but it just stands to reason.

The oldest versions of macOS that MacPorts supports do not come with git, and some of the old ones that do have git programs that cannot access GitHub anymore due to the set of TLS algorithms GitHub now uses. We would not want the selfupdate process to require the installation of the git port.

Downloading the release tarball from GitHub is similarly no good because the versions of curl that come with the oldest versions of macOS that we support can't talk to GitHub anymore either and we would not want to require the installation of the curl port.

neverpanic:
In addition to what Ryan already outlined, a git working copy also takes more space on disk, since it keeps history that we don't need. Now, this is probably not a huge problem for base, it would be for the ports tree, and it is a frequent annoyance reported by users of Homebrew, which does use git.

pmetzger:
I always keep the full thing anyway (since I have a checked out version of the repo for working on it.) It might be nice to give that as an option to people as it's a lot faster (at least for me) for updating.

PR #185, opened 2020-06-11
Abort 'port test' early if test.run is set to no

kurthindenburg:
This will avoid installing dependencies if tests is not enabled.

Closes: https://trac.macports.org/ticket/45010

### DIFF
--- a/src/macports1.0/macports.tcl
+++ b/src/macports1.0/macports.tcl
@@ -2227,6 +2227,14 @@ proc mportexec {mport target} {
         set log_needs_pop yes
     }
 
+    if {$target eq "test"} {
+        # Before continuing with dependents, verify test.run is present and yes
+        set test_valid [catch {_mportkey $mport test.run} res]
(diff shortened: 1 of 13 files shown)


neverpanic:
Hm, it seems that this actually breaks a couple of tests. Please take a look.

kurthindenburg:
Yea I missed that.... strange a lot of tests under tests/test fail even though they don't set test.run.

kurthindenburg:
When running those tests, the target is "test" and they don't have test.run in the Portfiles so it fails.  I wonder since the tests Portfiles all have 'test {}' in them, that this is why these fail.

kurthindenburg:
I'm not sure this is the best or even correct way to fix this.

kurthindenburg:
Before I work on this again, does anything object or have suggesting for a better way to handle the test portion? It does work as-is, just seems hackish.

jmroot:
Shouldn't be needed any more after ddc2655.

PR #186, opened 2020-06-12
pextlib1.0: init err buffer, use err code if empty

neverpanic:
curl < 7.60.0 does not initialize the error buffer, and cases have been reported where a curl download failed without the error buffer being set. This was the case on older systems that did not yet have curl 7.60.0, and caused random garbage to be printed.

0-initialize the buffer and check whether the buffer has a useful message before returning it. Generate a substitute message using `curl_easy_strerror()` from the returned error code if the error buffer is not set.

Closes: https://trac.macports.org/ticket/60581

### DIFF
--- a/src/pextlib1.0/curl.c
+++ b/src/pextlib1.0/curl.c
@@ -254,14 +254,14 @@ CurlFetchCmd(Tcl_Interp* interp, int objc, Tcl_Obj* CONST objv[])
 				/* check we also have the parameter */
 				if (optioncrsr < lastoption) {
 					optioncrsr++;
-					if ( numHTTPHeaders < MAXHTTPHEADERS ) {
-					  httpHeaders[numHTTPHeaders++] = Tcl_GetString(objv[optioncrsr]);
+					if (numHTTPHeaders < MAXHTTPHEADERS) {
+						httpHeaders[numHTTPHeaders++] = Tcl_GetString(objv[optioncrsr]);
 					} else {
-					  Tcl_SetResult(interp,
-					    "curl fetch: Too many --append-http-header options",
-					    TCL_STATIC);
-					  theResult = TCL_ERROR;
-					  break;
+						Tcl_SetResult(interp,
+							"curl fetch: Too many --append-http-header options",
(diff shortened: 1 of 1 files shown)


src/pextlib1.0/curl.c
```diff
@@ -645,8 +650,21 @@ CurlFetchCmd(Tcl_Interp* interp, int objc, Tcl_Obj* CONST objv[])
 		}
 		
 		if (info->data.result != CURLE_OK) {
-			/* execution failed, use the error string */
-			Tcl_SetResult(interp, theErrorString, TCL_VOLATILE);
+			/* execution failed, use the error string if it is set */
+			if (theErrorString[0] != '\0') {
+				Tcl_SetResult(interp, theErrorString, TCL_VOLATILE);
+			} else {
+				/* When the error buffer doesn not hold useful information,
```
ryandesign, line 657: "doesn't"

ryandesign: Well, either "does not" or "doesn't". Just not "doesn not".

ryandesign:
Oh and the "tests: Fix curl isnewer test broken by braeburn update" commit shouldn't be in this PR. That's already been merged to master.

neverpanic:
I pushed the "curl isnewer" commit to this branch as well to allow the tests to pass. It would have disappeared automatically when merging, but since I'm now pushing another version of the PR for your review comments anyway, I've rebased on top of master.

ryandesign:
> It would have disappeared automatically when merging

Ah. I still have no idea how git works. 😐

neverpanic:
I couldn't reproduce and thus not verify that this does what I wrote it to do, but Travis just triggered such a case and it seems to work as expected: https://paste.macports.org/75dc136dfd83

```
Error: Failed to fetch py38-lxml: curl_multi_info_read() returned {.msg = CURLMSG_DONE, .data.result = 7 (!= CURLE_OK)}, but the error buffer is not set. curl_easy_strerror(.data.result): Couldn't connect to server
```

PR #187, opened 2020-06-12
ReleaseProcess: Update buildbot job for manpages, drop Google+

neverpanic:


neverpanic:
Ignoring the failed check, since it does not affect the ReleaseProcess document.

PR #191, opened 2020-06-14
Don't add extra quoting to startupitems key values

ryandesign:
This allows multi-command start, stop and init values to work and allows executable values with arguments to work and gives the startupitems keyword feature parity with the older startupitem keywords.

Unfortunately it is also a change in syntax for existing users of the start, stop, init and executable keys of the startupitems keyword if the values of those keys contain whitespace. Additional quoting is now required.

See my question on the mailing list:

https://lists.macports.org/pipermail/macports-dev/2020-June/041973.html

Fortunately the list of affected ports using this syntax is small. The only affected ports I found are:

* [adblock2privoxy](https://github.com/macports/macports-ports/blob/master/www/adblock2privoxy/Portfile#L107-L139)
* [clamav-server](https://github.com/macports/macports-ports/blob/master/sysutils/clamav-server/Portfile#L47-L133)
* [macos-fortress](https://github.com/macports/macports-ports/blob/master/net/macos-fortress/Portfile#L287-L314)
* [mail-server](https://github.com/macports/macports-ports/blob/master/mail/mail-server/Portfile#L827-L877)

### DIFF
--- a/src/port1.0/portstartupitem.tcl
+++ b/src/port1.0/portstartupitem.tcl
@@ -1,7 +1,7 @@
 # -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:filetype=tcl:et:sw=4:ts=4:sts=4
 # portstartupitem.tcl
 #
-# Copyright (c) 2004-2012 The MacPorts Project
+# Copyright (c) 2004-2014, 2016-2018, 2020 The MacPorts Project
 # Copyright (c) 2006-2007 James D. Berry
 # Copyright (c) 2004,2005 Markus W. Weissman <mww@macports.org>
 # All rights reserved.
@@ -146,7 +146,7 @@ proc portstartupitem::foreach_startupitem {body} {
                 lappend startupitems_dict($curname) uniquename $uniquename
                 lappend startupitems_dict($curname) plist ${uniquename}.plist
             }
-            lappend startupitems_dict($curname) $key [list $val]
+            lappend startupitems_dict($curname) $key $val
         }
     } else {
         global startupitem.name
--- a/src/port1.0/tests/portstartupitem.test
+++ b/src/port1.0/tests/portstartupitem.test
@@ -20,25 +20,81 @@ macports_worker_init
 package require port 1.0
 package require portstartupitem 1.0
 
-test startupitems {
-    startupitems unit test. Check that the name seen in foreach_startupitem\
-    is the same whether set via the startupitem.name option or via key+value\
-    in the startupitems option.
+
+test startupitems_executable {
+    Check that setting the startupitem.executable option behaves the\
+    same as setting executable via the startupitems option.
 } -setup {
     set system_options(startupitem_install) yes
 } -body {
-    startupitem.name    testname
+    startupitem.executable  ${prefix}/sbin/mydaemon -flag1 "quoted string"
     portstartupitem::foreach_startupitem {
-        set single_result $si_name
+        set single_result [list {*}$si_executable]
     }
-    startupitems        name testname
+    startupitems \
+                name        testname \
+                executable  "${prefix}/sbin/mydaemon -flag1 \"quoted string\""
     portstartupitem::foreach_startupitem {
-        set multi_result $si_name
+        set multi_result [list {*}$si_executable]
     }
     if {$single_result ne $multi_result} {
         return "ERROR: single_result = '$single_result' but multi_result = '$multi_result'"
     }
-    return "startupitems successful."
-} -result "startupitems successful."
+    return "startupitems_executable successful."
+} -cleanup {
+    startupitems
+} -result "startupitems_executable successful."
+
+
+test startupitems_name {
+    Check that setting the startupitem.name option behaves the\
+    same as setting name via the startupitems option.
+} -setup {
+    set system_options(startupitem_install) yes
+} -body {
+    startupitem.name        testname
+    portstartupitem::foreach_startupitem {
+        set single_result [list {*}$si_name]
+    }
+    startupitems \
+                name        testname
+    portstartupitem::foreach_startupitem {
+        set multi_result [list {*}$si_name]
+    }
+    if {$single_result ne $multi_result} {
+        return "ERROR: single_result = '$single_result' but multi_result = '$multi_result'"
+    }
+    return "startupitems_name successful."
+} -cleanup {
+    startupitems
+} -result "startupitems_name successful."
+
+
+test startupitems_start {
+    Check that setting the startupitem.start option behaves the\
+    same as setting start via the startupitems option.
+} -setup {
+    set system_options(startupitem_install) yes
+} -body {
+    startupitem.start       "${prefix}/sbin/daemon1 start -flag1 \"quoted string\"" \
+                            "${prefix}/sbin/daemon2 start -flag2 -flag3"
+    portstartupitem::foreach_startupitem {
+        set single_result [list {*}$si_start]
+    }
+    startupitems \
+                name        testname \
+                start       "\"${prefix}/sbin/daemon1 start -flag1 \\\"quoted string\\\"\" \
+                            \"${prefix}/sbin/daemon2 start -flag2 -flag3\""
+    portstartupitem::foreach_startupitem {
+        set multi_result [list {*}$si_start]
+    }
+    if {$single_result ne $multi_result} {
+        return "ERROR: single_result = '$single_result' but multi_result = '$multi_result'"
+    }
+    return "startupitems_start successful."
+} -cleanup {
+    startupitems
+} -result "startupitems_start successful."
+
 
 cleanupTests


ryandesign:
By way of example, here's a Portfile. With the change in this PR, the a1 and a2 variants now do the same thing, and the b1 and b2 variants do too.

```
PortSystem 1.0
name test
version 1
distfiles
use_configure no
build {}
destroot {}
startupitem.create yes

variant a1 conflicts a2 b1 b2 {
    startupitem.start "echo hello ${name} >>/tmp/log" "echo start ${name} >>/tmp/log"
    startupitem.stop "echo goodbye ${name} >>/tmp/log" "echo stop ${name} >>/tmp/log"
}

variant a2 conflicts a1 b1 b2 {
    startupitems \
        name ${name} \
        start "\"echo hello ${name} >>/tmp/log\" \"echo start ${name} >>/tmp/log\"" \
        stop "\"echo goodbye ${name} >>/tmp/log\" \"echo stop ${name} >>/tmp/log\""
}

variant b1 conflicts a1 a2 b2 {
    startupitem.executable echo hello ${name}
}

variant b2 conflicts a1 a2 b1 {
    startupitems \
        name ${name} \
        executable "echo hello ${name}"
}
```

essandess:
A specific few comments (inline below).

This PR addresses  the case where multiple program arguments are required in a launchd.plist. Note that the current code already works when the launch item creates a shell `.wrapper` script. Also, the `post-activate` code to edit several other keys beyond `ProgramArguments` will still be necessary.

Details:

> Fortunately the list of affected ports using this syntax is small. The only affected ports I found are:
> 
> * [adblock2privoxy](https://github.com/macports/macports-ports/blob/master/www/adblock2privoxy/Portfile#L107-L139)

This is a case where `startupitems` works because it creates a shell `.wrapper` file. Correctly working script:
`/opt/local/etc/LaunchDaemons/org.macports.adblock2privoxy/adblock2privoxy.wrapper`
```bash
Start()
{
        ( /bin/test -f "${prefix}/etc/adblock2privoxy/privoxy/ab2p.task" \
        && "${prefix}/bin/adblock2privoxy" -t "${prefix}/etc/adblock2privoxy/privoxy/ab2p.task" \
        || "${prefix}/bin/adblock2privoxy" -p "${prefix}/etc/adblock2privoxy/privoxy" \
                -w "${prefix}/etc/adblock2privoxy/css" \
                -d 127.0.0.1:8119 \
                https://easylist.to/easylist/easyprivacy.txt \
                https://easylist.to/easylist/easylist.txt \
                https://easylist.to/easylist/fanboy-annoyance.txt \
                https://easylist.to/easylist/fanboy-social.txt \
                https://easylist-downloads.adblockplus.org/antiadblockfilters.txt \
                https://easylist-downloads.adblockplus.org/malwaredomains_full.txt \
                https://raw.githubusercontent.com/ryanbr/fanboy-adblock/master/fanboy-antifacebook.txt \
                https://raw.githubusercontent.com/Dawsey21/Lists/master/adblock-list.txt \
        ) && "${prefix}/bin/port" reload privoxy
}
```

Also, note that a `post-activate` block with `reinplace` commands will still be necessary to set the plist key `StartCalendarInterval `.

> * [clamav-server](https://github.com/macports/macports-ports/blob/master/sysutils/clamav-server/Portfile#L47-L133)

Ditto here for some of the startupitems; the main ones `clamd` and `freshclam` don't need multiple executable arguments. The `startupitems` correctly creates the `.wrapper` shell script.  And a `post-activate` block with `reinplace` commands will still be necessary to set the plist keys `UserName`, `GroupName`, `Nice`, `StartCalendarInterval `, `RunAtLoad`, `ProcessType`, and remove the `KeepAlive` key.

> * [macos-fortress](https://github.com/macports/macports-ports/blob/master/net/macos-fortress/Portfile#L287-L314)

In addition to the keys mentioned above, this ones also uses `StandardErrorPath` and `StandardOutPath`.

> * [mail-server](https://github.com/macports/macports-ports/blob/master/mail/mail-server/Portfile#L827-L877)

This Portfile replaces `ProgramArguments` to code around the issue of this PR, but also edits many of the other launchd.plist keys mentioned above.

Finally, startupitems are edited like this in other ports as well that do not necessarily use `startupitems`, e.g., [macos-vpn-server](https://github.com/macports/macports-ports/blob/2aea590ca3d2be9c69ce4c306cc7102ce139c3ab/net/macos-vpn-server/Portfile#L115-L122).

ryandesign:
> This PR addresses the case where multiple program arguments are required in a launchd.plist.

Yes, with regard to the "executable" key. It also fixes the case where multiple separate commands are to be used in the "init", "start", "stop" or "reload" keys.

> Note that the current code already works when the launch item creates a shell `.wrapper` script.

The way that the portfiles are currently written works with the way that MacPorts base currently interprets the syntax of the `startupitems` option. I am stating that the way that MacPorts base interprets the `startupitems` option is flawed because it does not allow all use cases of the `startupitem.*` options to be expressed. This PR fixes that problem and that will require all ports that use the old flawed syntax to be updated to use the new syntax, e.g. by adding extra layers of quoting.

> Also, the `post-activate` code to edit several other keys beyond `ProgramArguments` will still be necessary.

That's true. This PR is not trying to address the fact that MacPorts doesn't provide a good way to customize plists. I've filed other tickets for those enhancements.

src/port1.0/portstartupitem.tcl
```diff
@@ -146,7 +146,7 @@ proc portstartupitem::foreach_startupitem {body} {
                 lappend startupitems_dict($curname) uniquename $uniquename
                 lappend startupitems_dict($curname) plist ${uniquename}.plist
             }
-            lappend startupitems_dict($curname) $key [list $val]
```
jmroot, line 149: I don't remember why I used `list` here, but I doubt it was for no reason. It might have been an intermediate stage of development that should have been removed after `list` was added on line 161, or it might have been because of a problem with some other key.

What's really needed here is some tests that check that each `startupitems` key behaves identically to its corresponding `startupitem.*` option with a variety of values (single word, multiple words, characters that need quoting in a list).

ryandesign: Your implementation matches what you described in the commit message for 99d5fb30eb4c1ba8b7924807a052291ba31b792c:

```
Example:

startupitem.create  yes
startupitems        name        mydaemon \
                    executable  ${prefix}/bin/mydaemon \
                    name        anotherthing \
                    start       "${prefix}/bin/another --someflag"
                    stop        "${prefix}/bin/another shutdown"
```

I assumed that your implementation was intentional but that you just didn't consider the fact that startupitem.executable supported arguments or that startupitem.start et al supported specifying multiple commands. (I didn't realize that MacPorts supported the latter until I looked at the code two days ago; it's not documented in the guide.)

Adding tests is a good idea... but I've never understood our test suite. Do you know of a specific other test that would be a good model to copy for the startupitems tests, or would you be able to write one test or a placeholder that I could start from?

jmroot: > would you be able to write one test or a placeholder that I could start from?

Sure: 9c74d13

ryandesign: Thanks. And would I add the additional tests in the `-body` part of `test startupitems` or should I add additional `test` blocks?

ryandesign:
Here's another question. The four ports mentioned in this PR's description will need to be changed after this PR is merged. And I submitted an update for salt which also relies on this fix. Since the bug doesn't result in build failure but in a silently incorrect build, I would like to prevent incorrect builds from happening. How should these ports detect whether they are running in a version of MacPorts that contains the fix? In https://github.com/macports/macports-ports/pull/7440 I had salt check the MacPorts version, but that's not really reliable unless we delay the fix to MacPorts 2.7.0. If we release it in 2.6.3, then checking for a MacPorts version >= 2.6.3 won't necessarily guarantee a compatible version, since the user could be running an older version of 2.6.99 from master. Checking the contents of the plist can't be done until activation time because [MacPorts doesn't create the plist until the destroot phase is already over](https://trac.macports.org/ticket/60653). One solution could be to introduce a new MacPorts variable, set to any value; ports could then use `[info exists ...]` on the variable to verify that the bug is fixed.

src/port1.0/portstartupitem.tcl
ryandesign: How can I run just this test? The [instructions in the guide](https://guide.macports.org/#internals.tests.running) are unclear and nothing I've tried works.

jmroot: > Thanks. And would I add the additional tests in the `-body` part of `test startupitems` or should I add additional `test` blocks?

Ideally each test should test one independent thing. It can make sense to combine multiple related tests into one if the number of tests gets too unwieldy.

> How can I run just this test? The [instructions in the guide](https://guide.macports.org/#internals.tests.running) are unclear and nothing I've tried works.

```
cd src/port1.0
port-tclsh ../tests/test.tcl -t portstartupitem.test
```

jmroot:
> If we release it in 2.6.3, then checking for a MacPorts version >= 2.6.3 won't necessarily guarantee a compatible version, since the user could be running an older version of 2.6.99 from master.

We don't (and really can't) guarantee the ports tree is compatible with arbitrary versions of master. Anyone who runs master is responsible for keeping their installation up to date, and updating should be the first thing they try when there are problems.

ryandesign:
I dunno, I think we can, at least in this case, using a sentinel variable as I suggested above. We can even add a comment that the variable can be removed from base after awhile, once everyone has upgraded to that version. Isn't detecting problems before they happen and telling users how to solve them better than just silently letting the user build a nonfunctional port which they might then file a ticket about which we would then have to spend time diagnosing?

ryandesign:
Whenever we introduce a new feature in MacPorts base, we can phase its use into ports even before that version of base is released using `[info exists ...]` or `[info command ...]`. Remember how we introduced `use_tar`? (f10e0e371631a1a973d8988addaf98c3afb3f1db, https://github.com/macports/macports-ports/commit/3b0521a7e2060aaf4cf7cb8c744a60f4fdc6659a) I'm just wanting us to be able to do something similar for bugfixes.

src/port1.0/portstartupitem.tcl
ryandesign: Thanks. I changed your test a little and added a couple more. Is this a reasonable way to do it? Do we need more tests? (Testing *every* key seems redundant since for example init, start, stop, and restart should behave identically.)

essandess:
> the bug doesn't result in build failure but in a silently incorrect build, I would like to prevent incorrect builds from happening.

I’m still behind the curve about why this is a bug, why it produces “incorrect builds,” and what if anything would need to be changed in the resulting launchd.plists.

From my perspective as a Portfile author, I just need to make sure that the correct plist keys are set. There are several plist keys that must be modified; `ProgramArguments` is one of many.

Aside from using the new syntax in this PR that allows multiple program arguments within the `startupitems` directive, why would any Portfile that edits the `ProgramArguments` key need to be changed at all, and why is this an “incorrect build”?

I use the resulting plists all the time on many different machines and they all work flawlessly.

src/port1.0/tests/portstartupitem.test
```diff
@@ -20,25 +20,81 @@ macports_worker_init
...
+
+test startupitems_executable {
+    Check that setting the startupitem.executable option behaves the\
+    same as setting executable via the startupitems option.
+} -setup {
+    set system_options(startupitem_install) yes
+} -body {
+    startupitem.executable  ${prefix}/sbin/mydaemon -flag1 "quoted string"
+    portstartupitem::foreach_startupitem {
+        set single_result [list {*}$si_executable]
```
jmroot, line 32: Why pull apart and then reassemble the list?

ryandesign: It converts each list to its canonical form. I got that from [this tcl wiki page](https://wiki.tcl-lang.org/page/listcomp+-Compare+the+Contents+of+two+Lists) on the topic of how to compare lists for equality. Without that, the test failed because in one instance the quoted string was seen as being quoted with double quotes and in the other case was seen as being quoted with curly braces.

ryandesign:
> I use the resulting plists all the time on many different machines and they all work flawlessly.

Yes, I have no doubt that the way your ports are written *right now* works fine with the way MacPorts base works *right now*. But the way MacPorts base works *right now* does not allow startupitems executable to be set with program arguments, among other flaws as described in my mailing list post and in this pull request. The pull request will fix that deficiency but by doing so it will change the syntax of the startupitems option so that additional quoting will be required. Once this PR is merged existing portfiles that use the startupitems option will silently build incorrectly with that new version of base. Conversely, if the portfiles are fixed to use the new syntax then they would silently build incorrectly for users who have not updated to the new base. I've described the nature of the incorrect build in my mailing list post. Therefore I would like to introduce a way for portfiles to be able to detect that they are running in a fixed version of base and to issue an error if the user hasn't updated base yet. That will prevent any incorrect builds from occurring.

ryandesign:
@jmroot are you happy with this now or would you like me to make further changes?

neverpanic:
@ryandesign As for the version check in Portfiles, I think that's what `[macports_version]` is for. Just check for a newer version than 2.6.2 and expect the new behavior (of course ensuring that this gets merged into 2.6.3). This is what we have done in similar cases previously.

PR #194, opened 2020-06-23
Support MacOS version 10.16 and 11.0

koumaza:
Added support for MacOSBeta 10.16
Updated latest Xcode versions from "11.4.1" to "11.5"

### DIFF
--- a/src/port1.0/portutil.tcl
+++ b/src/port1.0/portutil.tcl
@@ -3336,17 +3336,22 @@ proc _check_xcode_version {} {
             10.15 {
                 set min 11.0
                 set ok 11.3
-                set rec 11.5
+                set rec 11.6
             }
             10.16 {
                 set min 11.0
                 set ok 11.3
-                set rec 11.5
+                set rec 11.6
+            }
+            11.0 {
+                set min 11.0
+                set ok 11.3
+                set rec 11.6
             }
             default {
                 set min 11.0
                 set ok 11.3
-                set rec 11.5
+                set rec 11.6
             }
         }
         if {$xcodeversion eq "none"} {
--- a/src/port1.0/portutil.tcl
+++ b/src/port1.0/portutil.tcl
@@ -3336,12 +3336,17 @@ proc _check_xcode_version {} {
             10.15 {
                 set min 11.0
                 set ok 11.3
-                set rec 11.4.1
+                set rec 11.5
+            }
+            10.16 {
+                set min 11.0
+                set ok 11.3
+                set rec 11.5
             }
             default {
                 set min 11.0
                 set ok 11.3
-                set rec 11.4.1
+                set rec 11.5
             }
         }
         if {$xcodeversion eq "none"} {
--- a/configure
+++ b/configure
@@ -2782,7 +2782,7 @@ $as_echo "$as_me: WARNING: This version of macOS is out of date" >&2;}
     { $as_echo "$as_me:${as_lineno-$LINENO}: WARNING: Please run Software Update to update it" >&5
 $as_echo "$as_me: WARNING: Please run Software Update to update it" >&2;}
     ;;
-  10.4.*|10.5.*|10.6.*|10.7.*|10.8.*|10.9.*|10.10.*|10.11.*|10.12.*|10.13.*|10.14.*|10.15.*)
+  10.4.*|10.5.*|10.6.*|10.7.*|10.8.*|10.9.*|10.10.*|10.11.*|10.12.*|10.13.*|10.14.*|10.15.*|10.16.*)
         ;;
   *)
     ;;
--- a/configure
+++ b/configure
@@ -2782,7 +2782,7 @@ $as_echo "$as_me: WARNING: This version of macOS is out of date" >&2;}
     { $as_echo "$as_me:${as_lineno-$LINENO}: WARNING: Please run Software Update to update it" >&5
 $as_echo "$as_me: WARNING: Please run Software Update to update it" >&2;}
     ;;
-  10.4.*|10.5.*|10.6.*|10.7.*|10.8.*|10.9.*|10.10.*|10.11.*|10.12.*|10.13.*|10.14.*|10.15.*|10.16.*)
+  10.4.*|10.5.*|10.6.*|10.7.*|10.8.*|10.9.*|10.10.*|10.11.*|10.12.*|10.13.*|10.14.*|10.15.*|10.16.*|11.0)
         ;;
   *)
     ;;
--- a/configure.ac
+++ b/configure.ac
@@ -48,7 +48,7 @@ case "$MACOSX_VERSION" in
     AC_MSG_WARN([This version of macOS is out of date])
     AC_MSG_WARN([Please run Software Update to update it])
     ;;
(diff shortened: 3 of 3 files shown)


ryandesign:
I thought Big Sur was called macOS 11, not macOS 10.16?

There seem to be weird `¥n` characters in your commit message.

koumaza:
> I thought Big Sur was called macOS 11, not macOS 10.16?
I think "10.16" is "Big Sur" UI/UX
> There seem to be weird ¥n characters in your commit message.
Fixed it.

ryandesign:
> I think "10.16" is "Big Sur" UI/UX

I'm not sure what you mean exactly. Do you know for certain that macOS Big Sur, which Apple also calls macOS 11, reports its version number as 10.16? I thought they were going to use the version number 11.0, but I haven't seen the beta yet so I don't know for certain. I do see one mention of "10.16" referring to the deployment target in the [Xcode 12 beta release notes](https://developer.apple.com/documentation/xcode-release-notes/xcode-12-for-macos-universal-apps-beta-release-notes).

koumaza:
>I'm not sure what you mean exactly. Do you know for certain that macOS Big Sur, which Apple also calls macOS 11, **reports its version number as 10.16?**

Yes. (`sw_vers` outputs 10.16)
It means that the UI/UX that Apple will release in the next term has arrived at 10.16 as β. I think it will be 11 as soon as β is over.

>the Xcode 12 beta release notes.
>> Xcode 12 for macOS Universal Apps beta requires a Mac running macOS 11 or later.

This is sure strange.

src/port1.0/portutil.tcl
```diff
@@ -3336,12 +3336,17 @@ proc _check_xcode_version {} {
             10.15 {
                 set min 11.0
                 set ok 11.3
-                set rec 11.4.1
+                set rec 11.5
+            }
+            10.16 {
+                set min 11.0
+                set ok 11.3
+                set rec 11.5
```
ryandesign, line 3344: Surely for 10.16/11 the min/ok/rec should be 12.0?

koumaza: My guess is,
(diff)
- 10.16
	- `set rec 12.0`
- 11.x
	- `set min 12.0`
	- `set ok 12.0`
	- `set rec 12.x`

koumaza:
MacOS10.16ScreenShot

saagarjha:
It's probably worth waiting a couple weeks to see if Apple settles on a specific value; I think part of the dichotomy was a late name change/some teams lacking disclosure. So perhaps they'll unify on one value later. The two Xcode 12 beta 1s we have right now is already a bit of a headache…

koumaza:
> I just checked on the current public beta version, and it looks like they've changed 10.16 to 11.0 now.
> Screen Shot 2020-08-10 at 10 47 16 AM

[`753516a` (#194)](https://github.com/macports/macports-base/pull/194/commits/753516a280419e7b110e6482f0163c7d3311aa56)
>Updated to 11.0.

saagarjha:
Apple has settled on 11.0, FWIW. You'll get 10.16 if you're running in compatibility mode (using an environment variable or  an old build SDK) so it's probably still worth keeping that around.

zhiyuanzhai:
Well... The upcoming Xcode version is 12.0, which is currently named `Xcode-beta`
截屏2020-09-11 上午10 39 51

koumaza:
I'm a little confused about MacPorts rules.
Using latest stable release?

ryandesign:
Xcode 11 does not support macOS 11. Xcode 12 beta did support macOS 11. Xcode 12 final does not support macOS 11. Xcode 12.2 beta does support macOS 11. We don't yet know which final version of Xcode will support macOS 11.

ryandesign:
> Xcode 12 has been released.
> 
> I'm not too familiar with the build system for MacPorts, but is this enough to continue support for macOS 11? MacPorts is the only thing holding me back from upgrading.

The fact that macOS 11 is in beta should be what's holding you back from upgrading. But if that doesn't deter you and you would like to assist us in reporting and/or fixing any MacPorts problems you find, you can certainly install macOS 11 beta and Xcode 12.2 beta and build macports-base master from git; that's what what we're doing, those of us who are already working on it. The changes in this PR are not needed to do that; you'll just see one line printed when you run ./configure that warns you your version of Xcode is not supported, which you can ignore.

zhiyuanzhai:
> Xcode 11 does not support macOS 11. Xcode 12 beta did support macOS 11. Xcode 12 final does not support macOS 11. Xcode 12.2 beta does support macOS 11. We don't yet know which final version of Xcode will support macOS 11.

@ryandesign It seems that Xcode supports only public versions of macOS... Maybe Xcode 12.2?

ryandesign:
> Maybe Xcode 12.2?

We don't know yet.

zhiyuanzhai:
@ryandesign Maybe we have to wait until the first Mac on ARM is launched. After that Big Sur will definitely be released. And then we know which Xcode version is correct.

IComplainInComments:
Is there a way we can contact Apple, and ask these questions to their engineers? I'm sure its nothing 'top secret' and we can sort this out?

ryandesign:
That seems unlikely but you are welcome to attempt to do so.

zhiyuanzhai:
Now the new MacBooks have been released. The latest Xcode is claimed to support macOS Big Sur. I believe everything has got an answer.

ryandesign:
@jmroot made the necessary modifications to portutil.tcl in 26e90b4c43269cf9c97d46979e073ff7a4107b53 and to xcode_versions.ini in macports/macports-ports@bb7cc5aa25fa8469bf895f540663b4265df6dc3e. The macOS version checks in configure and configure.ac have not yet been updated.

koumaza:
Is this PR still needed?

ryandesign:
Not the modifications to portutil.tcl since Josh's modifications supersede them. But some change is still needed in configure.ac and configure.

koumaza:
@ryandesign Thanks
Why did he only update `portutil.tcl`?

zhiyuanzhai:
@koumaza Yes, this is the current system, on which my MacBook is running. It's macOS 11.0.1 (20B28).

koumaza:
@zhiyuanzhai Thank you for telling me that.

jmroot:
> Not the modifications to portutil.tcl since Josh's modifications supersede them. But some change is still needed in configure.ac and configure.

It's kind of silly to have an ever-growing list of version patterns where matching them results in a noop, BTW. There's already a default case that is a noop. The important part is where it warns about outdated versions.

ryandesign:
That's true, I'm not sure why we have that.

The list of outdated versions is out of date...

koumaza:
https://github.com/macports/macports-base/commit/1a4266ba2feeee465b5cd5d5c0d3cd6d7332f8ec
This commit supports `11.0`. 
Stable latest version is `11.0.1`, so I'll do a PR corresponding to it.

ryandesign:
No need for a further PR. The above commit already correctly marks 11.0 as outdated.

koumaza:
Hi @ryandesign 
I create the PR to #215 before seeing your comment.
Should the PR be closed?

PR #195, opened 2020-06-25
portconfigure.tcl: use clang5+ for cxx11+ if not libc++

kencu:
only macports-clang-5+ clang compilers will accept macports-libstdc++

closes: https://trac.macports.org/ticket/59717

Presently, when a build calls for cxx_standard 2011, base will try to use clang-3.4 or clang-3.7 on PPC and on Intel if defaults are not to libc++.

This always fails, and either tries to call in libcxx on PPC (fails) or tries to send ```-stdlib=macports-libstdc++``` to clang 3.4 or clang-3.7 (fails).

### DIFF
--- a/src/port1.0/portconfigure.tcl
+++ b/src/port1.0/portconfigure.tcl
@@ -903,9 +903,19 @@ proc portconfigure::get_min_clang {} {
     if {${compiler.cxx_standard} >= 2017} {
         set min_value [max_version $min_value 5.0]
     } elseif {${compiler.cxx_standard} >= 2014} {
-        set min_value [max_version $min_value 3.4]
+        if {[option configure.cxx_stdlib] eq "libc++"} {
+            set min_value [max_version $min_value 3.4]
+        } else {
+            # macports-libstdc++ only macports-clang compilers >= 5.0 support this
+            set min_value [max_version $min_value 5.0]
+        }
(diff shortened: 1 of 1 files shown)


kencu:
@michaelld @ballapete @programmingkid

please have a look

ballapete:
Alright with me! As long as such a "clang5+" builds to be of use…

kencu:
@noloader try this out

kencu:
> Alright with me! As long as such a "clang5+" builds to be of use…

clang-5.0 does not build on PPC, but this PR stops base from selecting clang-3.4 or 3.7, and sends to gcc7, etc on PPC instead. 

Likewise on 10.4 and 10.5 Intel, it stops base trying to send ```-stdlib=macports-libstdc++``` to clang-3.4 and clang-3.7, which of course always fails as those compilers will not accept that.

What I would hope people might do is actually put this PR into their PPC / 10.4 Intel / 10.5 Intel systems and confirm that it does a comprehensive fix for compiler selection for them.

A new base is coming pretty soon, and if this is not in there before that, we're stuck with the current troubles for six more months.

ballapete:
I understand! A work-around. Still OK with me! It's a good plan.

kencu:
OK. Looks like happiness all around.

PR #196, opened 2020-06-26
Require clang >= 500 for C11

ryandesign:
Currently MacPorts base considers clang 318 to be new enough to support C11. However in https://trac.macports.org/ticket/58919 fontforge failed to build because:

```
scripting.c:10698:1: error: unknown type name '_Noreturn'
```

This predated the existence of `compiler.c_standard` so clang < 318 was manually blacklisted, but this was not enough; it was fixed by blacklisting clang < 500.

Similarly in https://trac.macports.org/ticket/59066 it was determined that ipmitool requires C11 and `compiler.c_standard 2011` was added, but [this still fails to build](https://build.macports.org/builders/ports-10.7_x86_64-builder/builds/24322/steps/install-port/logs/stdio):

```
/opt/local/include/openssl/crypto.h:322:1: error: unknown type name '_Noreturn'
```

This PR proposes changing the minimum clang version for `compiler.c_standard 2011` to 500 so that support for `_Noreturn` (which seems to be [a standard part of C11](https://en.wikipedia.org/wiki/C11_%28C_standard_revision%29#Changes_from_C99)) is guaranteed to be available.

### DIFF
--- a/src/port1.0/portconfigure.tcl
+++ b/src/port1.0/portconfigure.tcl
@@ -1,6 +1,6 @@
 # -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:filetype=tcl:et:sw=4:ts=4:sts=4
 #
-# Copyright (c) 2007 - 2015 The MacPorts Project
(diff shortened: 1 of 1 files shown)


ryandesign:
After this is merged, update https://trac.macports.org/wiki/CompilerSelection

PR #197, opened 2020-06-27
Fix shellescape to work on Tiger

ryandesign:
Fix shellescape to work on Tiger

Closes: https://trac.macports.org/ticket/56626

The previous code relied on the fact that we use `sh` for `system` and `sh` is supposed to be a POSIX shell but on Tiger it isn't; Leopard was the first POSIX compliant Mac OS X version


ryandesign:
I chose the list of safe characters based on the chart shown here:

https://stackoverflow.com/questions/15783701/which-characters-need-to-be-escaped-when-using-bash#answer-27817504

The test had to be changed from `echo` to `printf` because `echo` would interpret some of the escaped characters.

I verified the test now passes on Tiger and High Sierra.

PR #198, opened 2020-06-28
Do not interpose functions before initialization

saagarjha:
Static initializers have no defined order in which they are run; interposed functions, however, take effect at program initialization. In some cases, our interposed functions would be called by libSystem's constructors and prior to *our* constructors, and this would lead to crashes in trace mode as we would be in an uninitialized state. All our interposers now immediately delegate to their interposees until the are sure initialization has been performed.

### DIFF
--- a/src/darwintracelib1.0/access.c
+++ b/src/darwintracelib1.0/access.c
@@ -39,6 +39,10 @@
 #include <unistd.h>
 
 static int _dt_access(const char *path, int amode) {
+	if (!__darwintrace_initialized) {
+		return access(path, amode);
+	}
+
 	__darwintrace_setup();
 
 	int result = 0;
--- a/src/darwintracelib1.0/close.c
+++ b/src/darwintracelib1.0/close.c
@@ -48,6 +48,10 @@
  * will be set to the FD to be closed when closing should be allowed.
  */
 static int _dt_close(int fd) {
+	if (!__darwintrace_initialized) {
+		return close(fd);
+	}
+
 	__darwintrace_setup();
 
 	FILE *stream = __darwintrace_sock();
--- a/src/darwintracelib1.0/darwintrace.c
+++ b/src/darwintracelib1.0/darwintrace.c
@@ -88,7 +88,6 @@ volatile int __darwintrace_close_sock = -1;
 FILE *__darwintrace_stderr = NULL;
 
 static inline void __darwintrace_log_op(const char *op, const char *path);
-static void __darwintrace_setup_tls() __attribute__((constructor));
 static char *__send(const char *buf, uint32_t len, int answer);
 
 /**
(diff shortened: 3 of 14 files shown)


ryandesign:
> this would lead to crashes in trace mode

Has that ever actually been observed? Is there a ticket in Trac about it?

saagarjha:
Yeah, I linked to it in the commit message and you commented on it :P

I started seeing this crashing in [redacted], no doubt changes went into either dyld or libSystem to make this crop up now. I’m running trace mode, of course, to rule out other confounding factors because all the ports no longer build on [redacted] ;)

ryandesign:
Ok good to know. The ticket sounded theoretical; it didn't have any actual crash information attached but I understand why not.

neverpanic:
I'll backport this to the release-2.6 branch, too.

saagarjha:
Thanks! Is there a specific IRC channel or something of people who are disclosed on prerelease software, or have signed the DTK NDA, where I could share problems/crash reports when I run into them?

ryandesign:
See https://trac.macports.org/wiki/FAQ#prerelease

saagarjha:
Right, but there's nobody on that list for Big Sur :( Although, I decided re-read the Developer Program License Agreement and it actually seems to say that posting technical information is fine as long as it's not screenshots:

> You agree that all pre-release versions of the Apple Software and Apple Services (including pre- release Documentation), pre-release versions of Apple hardware, the FPS Deployment Package, any terms and conditions contained herein that disclose pre-release features, and the terms and conditions of Schedule 2 and Schedule 3 will be deemed “Apple Confidential Information”; provided however that upon the commercial release of the Apple Software the terms and conditions that disclose pre-release features of the Apple Software or services will no longer be confidential. Notwithstanding the foregoing, Apple Confidential Information will not include:
(i) information that is generally and legitimately available to the public through no fault or breach of Yours, (ii) information that is generally made available to the public by Apple, (iii) information that is independently developed by You without the use of any Apple Confidential Information, (iv) information that was rightfully obtained from a third party who had the right to transfer or disclose it to You without limitation, or (v) any FOSS included in the Apple Software and accompanied by licensing terms that do not impose confidentiality obligations on the use or disclosureofsuchFOSS. Further,AppleagreesthatYouwillnotbeboundbytheforegoing confidentiality terms with regard to technical information about pre-release Apple Software and services disclosed by Apple at WWDC (Apple’s Worldwide Developers Conference), except that You may not post screen shots of, write public reviews of, or redistribute any pre-release Apple Software, Apple Services or hardware.

So I guess it might be possible to drop that requirement now from the FAQ? However, since the DTK still has similar requirements I'd still be curious if anyone is part of that program.

ryandesign:
> there's nobody on that list for Big Sur

You could, for example, add yourself to the list, and write a post to the mailing list reminding others that they could add themselves to the list.

> So I guess it might be possible to drop that requirement now from the FAQ?

I am not a lawyer. I have not read the full agreement. I assume that we should stick with our current guidance.

jmroot:
> I am not a lawyer.

Likewise. But I will say that while it's not 100% clear given the phrasing, it seems more likely that what you are "not bound by the foregoing confidentiality terms with regard to" is "technical information [...] disclosed by Apple at WWDC", not "technical information about pre-release Apple Software and services".

saagarjha:
I would expect that specific verbiage is there to apply to the betas that Apple releases for its flagship operating systems (that is, the ones we're talking about right now) but to prevent it from automatically applying to some arbitrary software that they decide to preview to developers on an ad-hoc basis. Nonetheless, to be safe I just went ahead and added myself to the list in the FAQ and send a reminder to the macports-dev so that others might do so as well.

PR #199, opened 2020-07-10
Ping no more than 50 hosts per invocation

ryandesign:
The goal is to prevent a very large number of (ping) processes from being spawned all at the same time since that can cause the computer to become nonfunctional until it is restarted.

See https://trac.macports.org/ticket/60509#comment:9

The limit of 50 may not be right; we could pick a different limit.

### DIFF
--- a/src/port1.0/fetch_common.tcl
+++ b/src/port1.0/fetch_common.tcl
@@ -234,7 +234,6 @@ proc portfetch::sortsites {urls default_listvar} {
             }
         }
         set urllist $urlmap($url_var)
-        set hosts {}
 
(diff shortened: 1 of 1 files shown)


neverpanic (approved):
lg2m. 50 seems fine to me. Spawning 50 processes at the same time is already quite a lot.

jmroot:
TBH we should keep the mirror lists well below this length anyway. A couple dozen geographically diverse entries should be plenty.

ryandesign:
I've been seeing kernel panics on the 10.6 i386 worker again, so I've applied this patch there to see if it helps.

PR #200, opened 2020-07-26
Portindex: Add extended mode

arjunsalyan:
This commit adds a flag `-x` (for extended mode) which will allow to include more details in the Portindex, like notes, variant description etc.

**Why is this needed?**
We are trying to add variant `description` and port `notes` to the ports webapp. By faking the platform (`portindex -p`), we generate portindex for the latest MacOS and store it in the database. But the server and docker both run Linux, and hence the local ports tree is different from that generated by `portindex -p`. This renders `port variants` and `port notes` commands displaying incomplete/inconsistent results.

In addition to this, an extended portindex would also simplify the process of updating the database. We would be able to follow a single route, instead of three (which is being done right now).

### DIFF
--- a/src/port/portindex.tcl
+++ b/src/port/portindex.tcl
@@ -12,6 +12,7 @@ set permit_error 0
 set stats(total) 0
 set stats(failed) 0
 set stats(skipped) 0
+set extended_mode 0
 array set ui_options        [list ports_no_old_index_warning 1]
 array set global_options    [list]
 array set global_variations [list]
@@ -29,6 +30,7 @@ proc print_usage args {
     puts "-e:\tExit code indicates if ports failed to parse"
     puts "-o:\tOutput all files to specified directory"
     puts "-p:\tPretend to be on another platform"
+    puts "-x:\tInclude extra (optional) information in the PortIndex, like variant description and port notes."
 }
 
 proc _read_index {idx} {
@@ -236,6 +238,8 @@ for {set i 0} {$i < $argc} {incr i} {
                 lappend port_options os.platform $os_platform os.major $os_major os.arch $os_arch
             } elseif {$arg eq "-f"} { # Completely rebuild index
                 set full_reindex 1
+            } elseif {$arg eq "-x"} { # Build extended portindex (include extra information , eg.: notes, variant description, conflicts etc.)
+                set extended_mode 1
             } elseif {$arg eq "-e"} { # Non-zero exit code on errors
                 set permit_error 1
             } else {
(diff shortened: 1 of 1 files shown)


mojca:
@neverpanic, @jmroot: any objections?

ryandesign:
Is the intention for this to be a feature you would use locally on that linux server, or that the use of this flag would be incorporated on the server that runs mprsyncup?

mojca:
We would definitely like to use it locally on the server running the web app since we might potentially want to run `portindex` after each commit at some point, or at least more frequently / in more predictable time intervals than polling the server all the time, whether or not anything has changed or not.

That said, I would have preferred to publish the extended `.json` file if it wasn't a tiny bit annoying to have to generate the `PortIndex` twice.

ryandesign:
Anything done locally on that server is fine; you can of course apply this change to your local copy of MacPorts already and use it there.

If you can run it locally on that server, I'm not sure why you need it to be generated on a different server and distributed to you over the network. As you say there would be additional complications in needing to generate a separate extended copy of the portindex to then be converted to json for the web app to use.

I was concerned that this would increase the size of the portindex greatly but it's not that bad really:

```
$ ls -lh PortIndex* extended/PortIndex*
-rw-r--r--  1 root      wheel    14M Jul 31 13:08 PortIndex
-rw-r--r--  1 rschmidt  staff    23M Jul 31 13:28 PortIndex.json
-rw-r--r--  1 rschmidt  staff   545K Jul 31 13:08 PortIndex.quick
-rw-r--r--  1 root      wheel    16M Jul 31 13:08 extended/PortIndex
-rw-r--r--  1 rschmidt  staff    28M Jul 31 13:28 extended/PortIndex.json
-rw-r--r--  1 root      staff   548K Jul 31 13:26 extended/PortIndex.quick
```

I don't like the introduction of a flag called `-extended`. If we were to introduce a flag whose name is more than one letter, then it should be preceded by 2 dashes, not 1 (`--extended`). But the way that we handle flags in `port` is already weird, with single-letter single-dash flags having a different meaning and context from multi-letter two-dash flags. I would not like for `portindex` to introduce a different weird flag handling method. Since all the existing flags in `portindex` are single-letter single-dash, maybe that tradition should be continued and the new flag would be `-x`.

The new flag should be included in the `portindex` help output.

arjunsalyan:
> If you can run it locally on that server, I'm not sure why you need it to be generated on a different server and distributed to you over the network. As you say there would be additional complications in needing to generate a separate extended copy of the portindex to then be converted to json for the web app to use.

Once the new changes are deployed, webapp would not need the JSON from the rsync server. Repology uses the normal version of the `portindex.json`.
 
> I don't like the introduction of a flag called `-extended`. If we were to introduce a flag whose name is more than one letter, then it should be preceded by 2 dashes, not 1 (`--extended`). But the way that we handle flags in `port` is already weird, with single-letter single-dash flags having a different meaning and context from multi-letter two-dash flags. I would not like for `portindex` to introduce a different weird flag handling method. Since all the existing flags in `portindex` are single-letter single-dash, maybe that tradition should be continued and the new flag would be `-x`.
> 
> The new flag should be included in the `portindex` help output.

Thank you. `portindex -x` seems better, unless others have any objections/ different suggestions.

mojca:
> If you can run it locally on that server, I'm not sure why you need it to be generated on a different server and distributed to you over the network.

As Arjun already said, we don't really need the extended portindex generated on the server (all I was saying is that if we are already serving the `json`, we could just as well serve a slightly better `json`).

The reason why we would like to see this merged ASAP is not for the sake of being able to generate the extended `PortIndex` remotely, but for the sake of avoiding the need to fork and patch MacPorts sources to run the web app.

PR #202, opened 2020-08-17
portstartupitem: add startupitem options so that ports can specify the user and group to run a launchd service as

herbygillot:
This PR makes the following options available to Portfiles:
```
startupitem.user
startupitem.group
```

With these, ports can now specify the user and group that launchd will run the port's daemon process under.  These will create `<UserName>` and `<GroupName>` entries, respectively, in the launchd plist file. Per @ryandesign's [comment](https://github.com/macports/macports-ports/pull/7938#issuecomment-667800864), these options have been available in [launchd since 10.4](https://www.manpagez.com/man/5/launchd.plist/osx-10.4.php).

This was tested manually.  I wanted to check this in with proper tests, but was not having much success as I don't have enough of an understanding of the MacPorts internals.

### DIFF
--- a/src/port1.0/portstartupitem.tcl
+++ b/src/port1.0/portstartupitem.tcl
@@ -64,6 +64,12 @@
 #   startupitem.autostart   yes/no
 #       Automatically load the startupitem after activating. Defaults to no.
 #
+#   startupitem.user        <user>
+#       User to run the service/daemon as.
+#
+#   startupitem.group       <group>
+#       Group to run the service/daemon as.
+#
 #   startupitem.debug       yes/no
 #       Enable additional debug logging
 #       - for launchd, sets the Debug key to true
@@ -76,6 +82,7 @@ namespace eval portstartupitem {
 
 options startupitems startupitem.autostart startupitem.debug \
         startupitem.create startupitem.executable \
+        startupitem.user startupitem.group \
         startupitem.init startupitem.install startupitem.location \
         startupitem.logevents startupitem.logfile startupitem.name \
         startupitem.netchange startupitem.pidfile startupitem.plist \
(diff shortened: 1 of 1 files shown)


ryandesign:
Thanks, this should be useful. I tried it and the launchd plist it created looked correct.

Please limit the first line of the commit message to 50 characters. In base we don't enforce a component name prefix, so you have those extra characters available to you. If you need more than 50, leave a blank line and continue with paragraphs of up-to-72-character lines. Please add `See: https://trac.macports.org/ticket/13807` to the end of the commit message.

herbygillot:
OK, commit message fixed.

Also:

> https://trac.macports.org/ticket/13807

13 _years_! 😱😱😱😱

herbygillot:
I see these changes are approved, but I changed the commit message as requested.  Is this OK to merge?

raimue:
I agree this should indeed be possible to specify. How does this interact with startupitems using daemondo? It will attempt to write a pid file to `${prefix}/var/run`, which is owned by root. It cannot be written anymore when launchd already drops privileges. This is especially important for script startupitems (those with `.start`/`.stop` commands) as this is needed to kill/reload the daemon process.

neverpanic (approved):
I guess startupitems that use user and group will likely be unable to use daemondo. Since that's a choice the Portfile author has, and the default is not to set user or group, merging this will add new functionality but not break existing functionality. I'll thus merge this now – we can always improve user/group together with daemondo later.

herbygillot:
Thanks @neverpanic

mascguy:
Why was this approved and merged, without any updates to either the guide or manpage for `portfile`?

The lack of any documentation updates completely undermine the usefulness of new features, as they might as well not exist. And we shouldn't need to search the release notes, to see whether something exists.

Going forward, perhaps new feature additions - or at least, those that introduce new keywords - shouldn't be merged without both?

mascguy:
> I guess startupitems that use user and group will likely be unable to use daemondo. Since that's a choice the Portfile author has, and the default is not to set user or group, merging this will add new functionality but not break existing functionality. I'll thus merge this now – we can always improve user/group together with daemondo later.

Is this still accurate, as of the current MacPorts release? If so, what are the specific issues/limitations...?

PR #204, opened 2020-08-20
add small shell script wrapping `port` from local sources

danchr:
I wanted to test out some changes without upgrading my current installation. This minor script allowed me to do so.


port.sh
```diff
@@ -0,0 +1,19 @@
...
+port=$BASEDIR/src/port/port
+
+export TCLLIBPATH=$BASEDIR/src
+
+if [ ! -x $tclsh ]
+then
+    echo "please build MacPorts before running this script!" 1>&2
+    exit 1
+else
+    exec $tclsh $port "$@"
```
neverpanic, line 18:
```suggestion
    exec "$tclsh" "$port" "$@"
```

neverpanic (changes requested):
Strictly speaking, in autotools-based builds, the source directory is not always the build directory, so the tclsh and port executables may actually be located in a subfolder.

However, since out-of-tree building is broken in MacPorts anyway, I'm willing to ignore this.

danchr:
Ping?

port.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/sh
+#
+# Simple shell script to invoke MacPorts from the source directory
+#
+
+BASEDIR=$(cd $(dirname "$0"); pwd)
```
ryandesign, line 6: Need more quotes to properly handle paths that might contain spaces.
```suggestion
BASEDIR=$(cd "$(dirname "$0")"; pwd)
```

port.sh
```diff
@@ -0,0 +1,19 @@
...
+
+tclsh=$BASEDIR/vendor/tclsh
+port=$BASEDIR/src/port/port
```
ryandesign, line 9:
```suggestion
tclsh="$BASEDIR"/vendor/tclsh
port="$BASEDIR"/src/port/port
```

port.sh
```diff
@@ -0,0 +1,19 @@
...
+
+export TCLLIBPATH=$BASEDIR/src
```
ryandesign, line 11:
```suggestion
export TCLLIBPATH="$BASEDIR"/src
```

ryandesign (commented):
Being able to run port from the source directory without installing it would be useful for testing. Does this really work? Does it load all of its support files from the source directory, or are there still files it loads from the installation prefix?

danchr:
> Being able to run port from the source directory without installing it would be useful for testing. Does this really work? Does it load all of its support files from the source directory, or are there still files it loads from the installation prefix?

I'll try to run it in `opensnoop` and see if it accesses `${prefix}`, but AFAICT most things do work. Concerning spaces, I just tried to build it in another directory, and ran into this:

```
config.status: creating Makefile
config.status: creating pkgIndex.tcl
configure: === finished configuring in vendor/thread ([...]/macports-base-3 with space/vendor/thread)
configure: === configuring in vendor/tclx ([...]/macports-base-3 with space/vendor/tclx)
make: *** No rule to make target `distclean'.  Stop.
configure: running /bin/sh ./configure '--cache-file=/dev/null' '--disable-option-checking'  '--prefix=/opt/local/libexec/macports' '--srcdir=.' 'CC=clang' in vendor/tclx
checking for correct TEA configuration... ok (TEA 3.5)
checking for Tcl configuration... found [...]/macports-base-3 with space/vendor/tcl/unix/tclConfig.sh
checking for existence of [...]/macports-base-3 with space/vendor/tcl/unix/tclConfig.sh... loading
./configure: line 1603: [...]/macports-base-3: No such file or directory
configure: error: configure failed for vendor/tclx
``` 
I'll see if I can fix this easily.

danchr:
Okay, I've adjusted the script so that it's generated by `configure`; that should be more robust. I also added checks for out-of-source builds and build locations containing spaces, since both are broken. The latter is caused by Tcl, and it's non-trivial to fix — I managed to get most things building, but some oddity with `configure` invoking `make` still fails.

I've added a small sandbox profile that actively prevents accessing the installed sources, and `opensnoop` confirms that it works. The only thing accessed within `${prefix}` is the configuration.

port.sh
danchr: Simplied by using `autoconf`; fixed.

PR #206, opened 2020-08-26
notes: add UI_PREFIX to port name messages; hide them with -q

danchr:
notes: add UI_PREFIX to port name messages; hide them with -q

This makes 'port notes installed' much more useful since you can more
easily distinguish actual notes.

An example:

```
$ ./port.sh notes maven\* and installed
--->  maven3 has the following notes:
  To make maven 3.6.3 the default, please run
        sudo port select --set maven maven3
--->  maven_select has no notes.
```

### DIFF
--- a/src/port/port.tcl
+++ b/src/port/port.tcl
@@ -2377,6 +2377,8 @@ proc action_location { action portlist opts } {
 
 
 proc action_notes { action portlist opts } {
+    global UI_PREFIX
+
     if {[require_portlist portlist]} {
         return 1
     }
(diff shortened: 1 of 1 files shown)


src/port/port.tcl
```diff
@@ -2440,7 +2440,7 @@ proc action_notes { action portlist opts } {
                 puts [wrap $note 0 "  " 1]
             }
         } else {
-            puts "$portname has no notes."
+            ui_info "$portname has no notes."
```
neverpanic, line 2443: Ideally we'd hide this if run on a list of ports, but would still display this when run on a single port.

With this change, the current

```
$> port notes aquaterm
aquaterm has no notes.
```

would now become

```
$> port notes aquaterm
$>
```

which doesn't strike me as very user friendly :/

danchr: Yeah, I agree; I adjusted it so that the notes command simply adds the `UI_PREFIX` to the status message; this is the simplest solution I could think of without having to treat pseudoports or names specially.

neverpanic (approved):

PR #208, opened 2020-08-31
sandbox: handle /opt/local on separate mountpoint

danchr:
Some Xcode builds will fail when /opt is on a separate volume. The nature of the errors vary, for example `tiffutil` will frequently fail. They're all permission-related errors, though.

Looking at the system logs, I noticed this:

```
localhost kernel[0]: (Sandbox) Sandbox: tiffutil(99434) deny(1) file-write-create /opt/.TemporaryItems/folders.502/TemporaryItems/(A Document Being Saved By tiffutil)
```

To address this, I added some code that should detect the mountpoint for `${prefix}`, and add its `.TemporaryItems` to the allowed directories.

It's worth noting that I haven't tested this in the normal setup; the code _ought_ to work, but I'd appreciate the confirmation!

### DIFF
--- a/src/port1.0/portutil.tcl
+++ b/src/port1.0/portutil.tcl
@@ -3469,3 +3469,22 @@ proc _archive_available {} {
     set archive_available_result 0
     return 0
 }
+
+# get the mountpoint providing a given directory
+proc get_mountpoint {target_dir} {
+    file stat ${target_dir} target_stat
+
+    set parentdir ${target_dir}
+
+    while {$parentdir ne "/"} {
+        file stat [file dirname $parentdir] stat
+
+        if {$stat(dev) != $target_stat(dev)} {
+            return $parentdir
+        }
+
+        set parentdir [file dirname $parentdir]
+    }
+
+    return $parentdir
+}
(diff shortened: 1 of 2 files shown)


danchr:
See also https://github.com/macports/macports-ports/pull/8244 — that's where I ran into this, and it contains some relevant log messages.

src/port1.0/portutil.tcl
```diff
@@ -3469,3 +3469,22 @@ proc _archive_available {} {
...
     return 0
 }
+
+# get the mountpoint providing a given directory
+proc get_mountpoint {target_dir} {
+    file stat ${target_dir} target_stat
+
+    set parentdir ${target_dir}
+
+    while {$parentdir != "/"} {
```
neverpanic, line 3479:
```suggestion
    while {$parentdir ne "/"} {
```

In Tcl, "eq" and "ne" perform string comparisons, while "==" and "!=" attempt to perform integer comparisons first and then fall back to string comparisons.

neverpanic (changes requested):
A small nitpick, other than that this lg2m.

neverpanic (approved):

neverpanic:
Actually, could you please squash this so I can rebase the changes?

danchr:
> Actually, could you please squash this so I can rebase the changes?

Squashed and rebased!

neverpanic (approved):

PR #209, opened 2020-09-10
Improve deactivate/unmap failure error message

ryandesign:
When the registry fails to deactivate or unmap a file, say which file it is.

See: https://trac.macports.org/ticket/61154

### DIFF
--- a/src/cregistry/entry.c
+++ b/src/cregistry/entry.c
@@ -895,8 +895,8 @@ int reg_entry_unmap(reg_entry* entry, char** files, int file_count,
                     switch (r) {
                         case SQLITE_DONE:
                             if (sqlite3_changes(reg->db) == 0) {
-                                reg_throw(errPtr, REG_INVALID, "this entry "
-                                        "does not own the given file");
+                                reg_throw(errPtr, REG_INVALID, "%s is not "
+                                        "owned by this entry", files[i]);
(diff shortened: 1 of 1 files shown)


ryandesign:
There are several occurrences of this error message on our mailing list, in our tickets, and on Stack Exchange and other places from people who never learned how to resolve it, short of uninstalling MacPorts and starting over. I'd love to provide more info in the error message to help users resolve it.

Now that I know from 61154 that one reason why this message might be printed is if the index is corrupted, should I add `or the index is corrupted` to the end of the error message? Should I add additional suggestions like `run "pragma integrity_check;"`? The problem with that would be that users would not know that they have to run that within `sqlite3`, after having specified the registry file, using `sudo`, and also after using `.load macports.sqlext`, which they would also have to first build from source. Rather than explain all that in the error message, this might be a good opportunity to link to a web page with more information, which is something I'd like to do for many of our error messages, but we don't have a designated place in our documentation or wiki for pages that describe troubleshooting specific error messages. I suppose any old wiki page name would do for now.

neverpanic (approved):
I like the idea of having wiki pages with additional information. I guess you could just create a section in our wiki with additional information on error messages, i.e. https://trac.macports.org/wiki/diagnostics/<identifier> and assign a human readable identifier.

PR #212, opened 2020-10-19
Repoint manpage symlinks that point to symlinks

ryandesign:
MacPorts previously only fixed up manpage symlinks that pointed to real manpage files, but some ports install manpage symlinks that point to other symlinks that then point to real manpage files.

The old two-step method of first compressing all manpage files and then fixing up symlinks is replaced with a three-step process that first inventories the man directory, then compresses all manpages, then fixes up symlinks. This ensures that we have a complete and accurate inventory including what each symlink points to before we change anything.

Closes: https://trac.macports.org/ticket/61288

-----

This works for me to fix graphviz-devel's dot2gxl.1 symlink but I should do more testing on other types of ports, such as python modules or other ports that install manpages in a nonstandard man directory and then symlink those into the standard man directory.

### DIFF
--- a/src/port1.0/portdestroot.tcl
+++ b/src/port1.0/portdestroot.tcl
@@ -205,88 +205,73 @@ proc portdestroot::destroot_finish {args} {
     set bunzip2 "[findBinary bzip2 ${portutil::autoconf::bzip2_path}] -d"
     if {[file isdirectory ${manpath}] && [file type ${manpath}] eq "directory"} {
         ui_info "$UI_PREFIX [format [msgcat::mc "Compressing man pages for %s"] ${subport}]"
-        set found 0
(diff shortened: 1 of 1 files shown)


jmroot:
Do we actually need to do this at all on OS versions capable of hfscompression?

ryandesign:
If we could rely on hfscompression being available, then you're right, we wouldn't need this. However hfscompression requires the user to install libarchive to use its version of tar; the version provided by the OS hasn't been recent enough. We can't know at the time that we package up the archives on the build servers whether the user has hfscompression enabled. So I think we should continue to gzip-compress manpages as we've been doing.

Another reason not to change it now is that some ports may rely on the existing behavior, assuming that the manpages will be gzipped.

PR #213, opened 2020-11-08
Allow Universal on macOS 10.14 Mojave

Gcenx:
On MacOS Mojave require `macOS_SDK_headers_for_macOS_10.14.pkg` to be installed onto the system this allows 32Bit binaries to be compiled on the users system without needing to obtain a copy of MacOSX10.13.sdk and forcing a lower deployment target.

### DIFF
--- a/vendor/patches/tcl/configure-implicit-declarations.patch
+++ b/vendor/patches/tcl/configure-implicit-declarations.patch
@@ -88,3 +88,25 @@ index a7faae550..f482460ce 100644
  		[tcl_cv_]$1[_unbroken]=broken,[tcl_cv_]$1[_unbroken]=unknown))
  	if test ["$tcl_cv_]$1[_unbroken"] = "ok"; then
  	    tcl_ok=1
+diff --git a/vendor/tcl8.5.19/unix/configure b/vendor/tcl8.5.19/unix/configure
+index fcb64f18..96a26063 100755
+--- a/vendor/tcl8.5.19/unix/configure
++++ b/vendor/tcl8.5.19/unix/configure
+@@ -13665,7 +13665,7 @@ main ()
+ {
+ extern long timezone;
+ 	    timezone += 1;
+-	    exit (0);
++	    return (0);
+   ;
+   return 0;
+ }
+@@ -13731,7 +13731,7 @@ main ()
+ {
+ extern time_t timezone;
+ 		timezone += 1;
+-		exit (0);
++		return (0);
+   ;
+   return 0;
+ }
--- a/vendor/tcl8.5.19/unix/configure
+++ b/vendor/tcl8.5.19/unix/configure
@@ -13665,7 +13665,7 @@ main ()
 {
 extern long timezone;
 	    timezone += 1;
-	    exit (0);
+	    return (0);
   ;
   return 0;
 }
@@ -13731,7 +13731,7 @@ main ()
 {
 extern time_t timezone;
 		timezone += 1;
-		exit (0);
+		return (0);
   ;
   return 0;
 }


src/port1.0/portutil.tcl
```diff
@@ -3372,8 +3372,14 @@ proc _check_xcode_version {} {
             }
 
             # Check whether /usr/include and /usr/bin/make exist and tell users to install the command line tools, if they don't
-            if {${os.major} <= 17 && (![file isdirectory [file join $cltpath usr include]] || ![file executable  [file join $cltpath usr bin make]])} {
-                if {[vercmp $xcodeversion 10.0] >= 0} {
+            if {${os.major} <= 18 && (![file isdirectory [file join $cltpath usr include]] || ![file executable  [file join $cltpath usr bin make]])} {
+                if {${os.major} <= 18 && (![file isdirectory [file join $cltpath usr include]] || ![file executable  [file join $cltpath usr bin make]])} {
+                if {${os.major} == 18} {
```
ryandesign, line 3377: This doesn't look right, and the CI fails with:
```
warning: error while sourcing portutil.tcl: missing close-brace
```

ryandesign (commented):
If true, this would be great; thanks for figuring it out.

src/port1.0/portutil.tcl
Gcenx: Yeah a copy/paste mistake the repeated `if {${os.major} <= 18 && (![file isdirectory [file join $cltpath usr include]] || ![file executable  [file join $cltpath usr bin make]])} {` shouldn’t be there

I won’t be able to fix this until I’m home.

Gcenx:
> If true, this would be great; thanks for figuring it out.

Ignoring the copy/paste mistake it does indeed work, I should have verified this after copy/pasting from my testing in release-2.6 branch locally.

src/port1.0/portconfigure.tcl
```diff
@@ -404,12 +404,14 @@ proc portconfigure::configure_start {args} {
...
 # universal_archs, plus the SDK being used
 proc portconfigure::choose_supported_archs {archs} {
-    global supported_archs configure.sdk_version
+    global supported_archs configure.sdk_version os.major
 
     if {${configure.sdk_version} ne ""} {
         # Figure out which archs are supported by the SDK
         if {[vercmp ${configure.sdk_version} 11.0] >= 0} {
             set sdk_archs [list arm64 x86_64]
+        } elseif {${os.major} == 18 && [file exists /usr/include/sys/cdefs.h]} {
```
jmroot, line 413: This check is incomplete; i386 is only supported if `${configure.sdkroot}` is also empty, or equivalently if `${configure.sdk_version}` is the same as `$macosx_version`.

src/port1.0/portutil.tcl
```diff
@@ -3372,8 +3372,13 @@ proc _check_xcode_version {} {
...
 
             # Check whether /usr/include and /usr/bin/make exist and tell users to install the command line tools, if they don't
-            if {${os.major} <= 17 && (![file isdirectory [file join $cltpath usr include]] || ![file executable  [file join $cltpath usr bin make]])} {
-                if {[vercmp $xcodeversion 10.0] >= 0} {
+            if {${os.major} <= 18 && (![file isdirectory [file join $cltpath usr include]] || ![file executable  [file join $cltpath usr bin make]])} {
+                if {${os.major} == 18} {
+                    ui_warn "System headers do not appear to be installed"
+                    ui_warn "You can install it as part of the Xcode Command Line Tools package by running `xcode-select --install'."
+                    ui_warn "Next install 'macOS_SDK_headers_for_macOS_10.14.pkg' that's included with the Command Line Tools package"
+                    ui_warn "'installer -pkg /Library/Developer/CommandLineTools/Packages/macOS_SDK_headers_for_macOS_10.14.pkg -target /'"
```
jmroot, line 3380: I don't think it's useful to start complaining about this all the time. Only bring it up if a 32-bit build is actually requested and can't succeed.

jmroot (changes requested):

Gcenx:
I'll squash the commits again once any reviews are happy

kencu:
MacPorts 2.70 is imminent; this might squeak in if you're lucky!

kencu:
Lots of ports require Xcode to build, though...and the headers in Xcode won't support i386.

So won't some ports then build i386 against "/" and other ports that require_xcode fail?

Gcenx:
> MacPorts 2.70 is imminent; this might squeak in if you're lucky!

I guess if the current fail is fixed maybe it will make it, but I’m now stuck at work...

If anyone is able to tweak this so it passed and is acceptable by @jmroot I'm all for it

Gcenx:
> Lots of ports require Xcode to build, though...and the headers in Xcode won't support i386.
> 
> So won't some ports then build i386 against "/" and other ports that require_xcode fail?

As long as Macports is setting sdkroot this should be an issue

src/port1.0/portconfigure.tcl
Gcenx: I assume I resolved this in the way you explained

src/port1.0/portutil.tcl
Gcenx: This one I’m really not sure on

kencu:
MP has decided that if the SDKROOT is "/" then it won't be set to that, for various reasons. Some ports override that.

Gcenx:
> MP has decided that if the SDKROOT is "/" then it won't be set to that, for various reasons. Some ports override that.

You mean when the system has the needed /usr/include and /usr/lib SDKROOT=“” is set over the incorrect SDKROOT=“/“?

kencu:
> > MP has decided that if the SDKROOT is "/" then it won't be set to that, for various reasons. Some ports override that.
> 
> You mean when the system has the needed /usr/include and /usr/lib SDKROOT=“” is set over the incorrect SDKROOT=“/“?

I mean when the system has the needed /usr/include and /usr/lib SDKROOT=“” is set over the **correct** SDKROOT=“/“.

Exactly. We worked on this before as an issue, but there are various things about it. There's a PR in here I opened and closed several times that fixed that issue, but in the end, it is  NTBF.

Gcenx:
> I mean when the system has the needed /usr/include and /usr/lib SDKROOT=“” is set over the (what might be considered) **correct** SDKROOT=“/“.
> 
> Exactly. We worked on this before as an issue, but there are various things about it. There's a PR in here I opened and closed several times that fixed that issue, but in the end, it is NTBF.

The problem with setting SDKROOT=“/“ is then every other environment variable that uses SDKROOT ends up having an extra “/“ this can cause other issues.

I’m guessing your pull-request corrected the above issue.

ryandesign:
> every other environment variable that uses SDKROOT ends up having an extra “/“ this can cause other issues.

What issues does it cause? I've [asked (someone else) this before](https://trac.macports.org/ticket/59798#comment:3) and didn't get a response.

kencu:
I just want to triple-check with you that these headers really do have the i386 parts still in them.

The i386 parts were partially stripped out of the 10.14 SDK, and (I believe) are totally gone in the 10.15 SDK.

It seems very weird to me that there are headers that have the full i386 components still in place, in a 10.14 SDK.

It seems much more likely to me that these are just the 10.14 SDK headers, that install into "/", in which case they are no good to restore 32bit building comprehensively on Mojave.

Gcenx:
> What issues does it cause? I've [asked (someone else) this before](https://trac.macports.org/ticket/59798#comment:3) and didn't get a response.

It “shouldn’t” cause issues but some programs ended up not finding some components.


> I just want to triple-check with you that these headers really do have all the i386 parts still in them.
> 
> The i386 parts were partially stripped out of the 10.14 SDK, and (I believe) are totally gone in the 10.15 SDK.
> 
> It seems very weird to me that there are headers that have the full i386 components still in place, in a 10.14 SDK. Why? Where did they come from? Why would Apple do such a thing?
> 
> It seems much more likely to me that these are just the 10.14 SDK headers, that install into "/", in which case they are no good to restore 32bit building comprehensively on Mojave.

You can test this yourself by installing the headers then forcing SDKROOT=“” or “/“ then compile something as 32Bit.

If you would like a working example download one of my wine packages and load it on Mojave, now launch “wine explorer” and change from light to dark mode and watch the title bar change color. You can also check the “wine” binary and see what SDK versions is was compiled with.

The most recent packages wine-5.20/5.21 were building with SDKROOT=“” on Mojave

Gcenx:
To expand on this, the 10.14.SDK includes tbd files that only contain 64Bit components, however Mojave still ships with universal dylibs so only compatible headers are required. The SDK ends up being equivalent to 10.13 with a few differences

kencu:
The 10.14 SDK headers do have some of the i386 stuff still in them, but enough has been stripped out that sufficient builds fail to build that it's not worth the trouble. I tried a couple of years ago to just force building against them, and quickly ran into important i386 stuff that was gone, that was still in the 10.13 headers.

So some 32bit stuff will build against the 10.14 headers. But not enough to make it useful.

I thought you had come across secret headers that still had ALL the i386 stuff still in them, ie equivalent to the i386 support that is in the 10.13 headers.

I could be wrong -- I haven't gone back into trac looking for my old 10.14 tickets where I forced this and decided it was a dead end.

BUT -- if diff-ing the headers you are installing here, and the headers that are already in the 10.14 SDK, indicates they are just the same headers, then installing them into "/" would be a NO-GO and you're better off using the 10.13 SDK instead.

kencu:
I found the discussion <https://lists.macports.org/pipermail/macports-dev/2018-November/039681.html>

Gcenx:
The provided solution is something that’s legal, I have alternatives but that’s would be a huge no.

I’ll do a diff between the headers in the SDK and the installer headers if needed.

As for the link I’ve build FAudio +universal with this.

Gcenx:
And yes the headers package also installs the headers for frameworks

kencu:
See the link. 

Some things will build i386 against the 10.14 headers, but lots of the 32 bit things have been stripped out, so IMHO it is not a robust or worthwhile solution. It will just lead to headaches to try that approach. The 10.13 SDK is 100% reliable, if a bit of a PITA.

I thought you had come across some previously-unknown 10.14 headers that had not been stripped. It's true, the 10.14 headers are not FULLY stripped, but the are significantly stripped.

Gcenx:
I did read the contexts of the link and already said I’ve built FAudio +universal using this that means I also built SDL2 as +universal.

Once home I’ll give a comparison against some headers if you’d like.

To clarify the headers package is provided different headers then the MacOSX10.14.SDK, also as explain above I do have other solutions but at that point it’s the same are requiring the 10.13.SDK

kencu:
I know some things build. But I found the header are sufficiently stripped than not enough remains. I tried this two years ago, then ran into build errors with some ports, and abandoned it for the 10.13 SDK approach.

But feel free to press on. Perhaps I'm wrong. If the headers in you're installing in "/" are different than the headers in the SDK, and have all the i386 bits in them, OK.

Gcenx:
```
Deans-MBP:~ gcenx$ cd /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.14.sdk/System/Library/Frameworks/Carbon.framework/Versions/A/Frameworks/CarbonSound.framework/Versions/A/Headers
Deans-MBP:Headers gcenx$ ls -la
total 40
drwxr-xr-x  4 gcenx  staff    128 Sep 23 09:04 .
drwxr-xr-x  3 gcenx  staff     96 Sep 23 09:04 ..
-rw-r--r--  1 gcenx  staff    672 Sep 23 09:04 CarbonSound.r
-rw-r--r--  1 gcenx  staff  14278 Sep 23 09:04 Sound.r
```

And in the attached zip is log data of compiling wine-crossover-20.0.1 +universal from source, all requirments were compiled from source using commit 9d19560
[wine-crossover-20.0.1.zip](https://github.com/macports/macports-base/files/5531174/wine-crossover-20.0.1.zip)

Gcenx:
@jmroot this now functions in the manner you requested, x86_64 only builds are not affected but once i386 is requested and the headers package is not installed the user will be prompted to install them.

kencu:
Hey, that looks encouraging. 

Keep in mind that any software that forces the  macos SDK somehow, either the one in command line tools or in Xcode, will be suspect. (and there is such software out there -- qt5 comes to mind, off the bat, as a likely candidate, also llvm/clang and others). I suspect breakage -- but it's looking tempting.

Gcenx:
Are those not the same Ports that set SDKROOT=“/“ when SDKROOT=“” to workaround this kind of issue?

I know meson wasn’t affected as I happily produced gstreamer-1.18

kencu:
There are lots of shenanigans we don't pick up hidden in the ports.... remember that one fix last year we tried to get in where the port forced the SDK in the configure script, for example?  That one required patching the configure script to allow setting the SDK to be obeyed. clang does it's own thing in the cmake files -- sometimes it goes looking for an SDK it likes, and uses (for example) an iPhone SDK of all things.

So -- lots of places where issues might come up. But looks like you have @jmroot's ear hear, so you might get this in.

Gcenx:
Yeah I remember that one went in along with some others, removing that is considered correct Macports behavior thought?

It would be nice to have this in as these very issues would be found by anyone who decides to make use of this and can be fixed.

Gcenx:
This still shows “1 change requested” does that mean @jmroot needs to accept it’s resolved as it’s marked as resolved on my end. Or is GitHub just being stupid?

The last force push was to ensure Macports won’t keep showing warnings for a missing SDK when using “/“ as the SDK.

Too bad this didn’t make it into the new release

src/port1.0/portconfigure.tcl
```diff
@@ -404,12 +404,14 @@ proc portconfigure::configure_start {args} {
...
 # universal_archs, plus the SDK being used
 proc portconfigure::choose_supported_archs {archs} {
-    global supported_archs configure.sdk_version
+    global supported_archs configure.sdk_version macosx_version
 
     if {${configure.sdk_version} ne ""} {
         # Figure out which archs are supported by the SDK
         if {[vercmp ${configure.sdk_version} 11.0] >= 0} {
             set sdk_archs [list arm64 x86_64]
+        } elseif {[vercmp ${configure.sdk_version} 10.14] >= 0 && $macosx_version eq "10.14" && [file exists /usr/include/sys/cdefs.h]} {
```
jmroot, line 413: This still isn't quite correct; instead of `[vercmp ${configure.sdk_version} 10.14] >= 0` it should be `${configure.sdk_version} eq "10.14"`.

src/port1.0/portutil.tcl
```diff
@@ -3260,7 +3260,14 @@ proc check_supported_archs {} {
                     lappend unsupported $arch
                 }
             }
-            ui_error "$subport cannot be installed for the configured universal_archs '$universal_archs' because the arch(s) '$unsupported' are not supported."
+            if {$unsupported ne "" && ${os.major} == 18} {
```
jmroot, line 3263: This should check that `$unsupported` is only i386.

src/port1.0/portutil.tcl
```diff
@@ -3372,7 +3379,11 @@ proc _check_xcode_version {} {
             }
 
             # Check whether /usr/include and /usr/bin/make exist and tell users to install the command line tools, if they don't
-            if {${os.major} <= 17 && (![file isdirectory [file join $cltpath usr include]] || ![file executable  [file join $cltpath usr bin make]])} {
+            if {${os.major} <= 18 && (![file isdirectory [file join $cltpath usr include]] || ![file executable  [file join $cltpath usr bin make]])} {
+                if {${os.major} == 18 && [option configure.sdk_version] ne "" && [file tail [option configure.sdkroot]] ne "MacOSX[option configure.sdk_version].sdk"} {
+                    ui_warn "The macOS [option configure.sdk_version] SDK does not appear to be installed. Ports may not build correctly."
+                    ui_warn "You can install it as part of the Xcode Command Line Tools package by running `xcode-select --install'."
+                }
```
jmroot, line 3386: The else blocks of the two if statement below this are still going to execute and print incorrect warnings.

src/port1.0/portutil.tcl
Gcenx: Sure I'll change this to `i386`

src/port1.0/portutil.tcl
Gcenx: I can swap the xcode check to elseif or move this check into its own block that only deals with Mojave

src/port1.0/portconfigure.tcl
Gcenx: done

Gcenx:
@jmroot let me know once you have chance to check it over, but I believe since your last commit to master I’ll need to update this

PR #214, opened 2020-11-13
restrict compiler for C++14 and thread-local storage

chrstphrchvz:
…by revising approach from macports/macports-base#161
Fixes: https://trac.macports.org/ticket/61418

Untested; I currently am not using macports-base from git, nor an old-enough macOS version where this would be applicable.

### DIFF
--- a/src/port1.0/portconfigure.tcl
+++ b/src/port1.0/portconfigure.tcl
@@ -860,16 +860,15 @@ proc portconfigure::get_min_command_line {compiler} {
             } elseif {${compiler.cxx_standard} >= 2014} {
                 set min_value [max_version $min_value 602.0.49]
             } elseif {${compiler.cxx_standard} >= 2011} {
-                if {${compiler.thread_local_storage}} {
-                    # macOS has supported thread-local storage since Mac OS X Lion.
-                    # So __thread (GNU extension) and _Thread_local (C11) could be used.
-                    # However, the C++11 keyword was not supported until Xcode 8
-                    #    (https://developer.apple.com/library/archive/releasenotes/DeveloperTools/RN-Xcode/Chapters/Introduction.html#//apple_ref/doc/uid/TP40001051-CH1-SW127)
-                    #    (https://developer.apple.com/videos/play/wwdc2016-405/?time=354).
-                    set min_value [max_version $min_value 800.0.38]
-                } else {
-                    set min_value [max_version $min_value 500.2.75]
-                }
+                set min_value [max_version $min_value 500.2.75]
+            }
+            if {${compiler.cxx_standard} >= 2011 && ${compiler.thread_local_storage}} {
+                # macOS has supported thread-local storage since Mac OS X Lion.
+                # So __thread (GNU extension) and _Thread_local (C11) could be used.
+                # However, the C++11 keyword was not supported until Xcode 8
+                #    (https://developer.apple.com/library/archive/releasenotes/DeveloperTools/RN-Xcode/Chapters/Introduction.html#//apple_ref/doc/uid/TP40001051-CH1-SW127)
+                #    (https://developer.apple.com/videos/play/wwdc2016-405/?time=354).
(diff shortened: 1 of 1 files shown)


ryandesign:
Looks plausible but I haven't tested it yet either.

mascguy:
I created a suite of VMs specifically for MacPorts work, covering macOS releases 10.6 through 10.15. So if there is a specific release or two where this needs to me tested, please let me know. Would be glad to help!

ryandesign:
The issue affects ports that specify `compiler.cxx_standard` greater than 2011 on systems where Xcode clang's version is ≥ 500.2.75 < 800.0.38 (and the port does not already blacklist clang < 800.0.38), so you could test such ports on such systems.

mascguy:
I was tied up with endless local test rebuilds of py-pytorch today - for both python 3.8 and 3 9, across 10.12 through 10.14 - with build times that match or eclipse Mame for each iteration. (And enabling ccache for py-pytorch will require hacky gymnastics, exacerbating the situation. Grumble, grumble. Lol)

But now that that's out of the way, I'll run some validation tests for this PR tomorrow. Or make that later today, as it's 1:35am here.

Stay tuned...

chrstphrchvz:
Examples include poppler, gnss-sdr, gnss-sdr-devel. Remove `compiler.blacklist-append {clang < 800.0.38}` (if present) to test. I don't think completely building ports is necessary; only need to verify that at the beginning of the configure phase (`port -ds configure`) it does not output:
```
--->  Configuring (port)
…
DEBUG: Using compiler 'Xcode Clang'
```

mascguy:
Makes sense. Starting my testing now, I'll report back in a bit.

mascguy:
Okay, I've confirmed that the bug exists in macOS 10.10, using poppler.

In terms of patching my MacPorts installation, is `portconfigure.tcl` the only file I need from the master branch? (Obviously I'd prefer to not reinstall MacPorts from source, if possible...)

mascguy:
It looks like I might need everything from port1.0, to be safe. Is there anything else?

mascguy:
Okay, I've confirmed that the bug exists in macOS 10.10, using poppler.

In the interest of thoroughly testing the fix, I'll make a temporary clone of my 10.10 VM, and reinstall MacPorts from source (via master). Do you folks know whether master (for base) is in a good state at this point?

mascguy:
The testing took a bit longer than expected, due to other fires... but it's done. And looks good to me.

While I ran poppler's configure phase for both cases (and reviewed the output), you can immediately see a key change simply by running `port info poppler`: The differing build dependencies tell all.

MacPorts v2.6.4:
```
$ port info poppler
poppler @20.12.1_1 (graphics)
Sub-ports:            poppler-qt5
Variants:             boost, debug, universal

Description:          Poppler is a PDF rendering library based on the xpdf-3.0 code base.
Homepage:             https://poppler.freedesktop.org/

Extract Dependencies: xz
Build Dependencies:   cmake, pkgconfig
[...]
```

MacPorts Master:
```
$ port info poppler
poppler @20.12.1_1 (graphics)
Sub-ports:            poppler-qt5
Variants:             boost, debug, universal

Description:          Poppler is a PDF rendering library based on the xpdf-3.0 code base.
Homepage:             https://poppler.freedesktop.org/

Extract Dependencies: xz
Build Dependencies:   cmake, pkgconfig, clang-9.0
[...]
```

And kudos to whoever implemented the new progress bar. It's definitely a nice addition!

mascguy:
Folks, let me know if my testing was sufficient to provide this out. Or if there are any items related to this, I'm happy to assist with those too.

ryandesign:
Thanks, it's good to hear that the fix appears to work as intended.

FYI master is almost always in a good state, and many of us routinely run master.

Progress bar support was implemented by one of the MacPorts founders @landonf last year; see da3524e8a3006a9b9a2370b66fa91bb554dd768b. Thanks Landon!

PR #215, opened 2020-11-14
Support 11.0.1 of Latest Stable MacOS Version

koumaza:
The latest stable MacOS version is 11.0.1
Therefore, will update it.
If you take to merge, I recommend squash and merge.


ryandesign:
You've added "11.0.1" to the list of out of date OS versions. 11.0.1 isn't out of date; it's current.

koumaza:
@ryandesign understood. has come to list only outdates in configure, right?

ryandesign:
We print a warning if we know the user's OS version is out of date. Otherwise we don't print anything.

PR #216, opened 2020-11-18
configure: remove a bash-ism in daemondo check

danchr:
Using '==' with 'test' is not part of the POSIX shell specification, and 'dash' doesn't support it.

### DIFF
--- a/aclocal.m4
+++ b/aclocal.m4
@@ -843,10 +843,10 @@ AC_DEFUN([MP_PROG_DAEMONDO],[
 	result=no
 	case $host_os in
 		darwin*)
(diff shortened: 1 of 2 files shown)


ryandesign (approved):

danchr:
Should I merge this myself?

raimue:
Thanks, good catch! Maybe we should even some automated check using `shellcheck` for such things?

danchr:
> Thanks, good catch! Maybe we should even some automated check using `shellcheck` for such things?

Well, a much easier test might be simply running `dash ./configure` on systems where it's available 🙂

PR #217, opened 2020-11-19
port1.0: configure: Add check for implicit decls

neverpanic:
On macOS 11, implicit function declarations are now errors by default. This may affect configure checks and actually cause software to compile differently without us knowing.

Add a post-configure check that scans any `config.log` files in `${configure.dir}` for indicators of such problems and warn about them, so they will not go unnoticed.

Provide a way to disable this functionality in a Portfile, in case the maintainer did verify that those are false positives.

### DIFF
--- a/src/port1.0/portconfigure.tcl
+++ b/src/port1.0/portconfigure.tcl
@@ -33,11 +33,13 @@
 package provide portconfigure 1.0
 package require portutil 1.0
 package require portprogress 1.0
+package require struct::set
 
 set org.macports.configure [target_new org.macports.configure portconfigure::configure_main]
 target_provides ${org.macports.configure} configure
 target_requires ${org.macports.configure} main fetch checksum extract patch
 target_prerun ${org.macports.configure} portconfigure::configure_start
+target_postrun ${org.macports.configure} portconfigure::configure_end
 
 namespace eval portconfigure {
 }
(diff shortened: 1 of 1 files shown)


ryandesign:
Good idea. I'll test this.

ryandesign (approved):
Seems to work. And we do need to make it easier to see these types of errors. But it's going to point out soooooo many errors that we need to fix. ☹️☹️☹️

ryandesign:
And then there will be ports like freetds where it reports:
```
Warning:   conftest.c:152:16: error: implicit declaration of function 'gethostbyname_r' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
```
`gethostbyname_r` is not available on macOS, so it is fine that the configure script checks for it and doesn't find it.

I can add `configure.checks.implicit_function_declaration no` but won't that mean that I'll miss out on any new implicit declaration of function errors introduced by future versions of freetds?

neverpanic:
> And then there will be ports like freetds where it reports:
> 
> ```
> Warning:   conftest.c:152:16: error: implicit declaration of function 'gethostbyname_r' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
> ```
> 
> `gethostbyname_r` is not available on macOS, so it is fine that the configure script checks for it and doesn't find it.
> 
> I can add `configure.checks.implicit_function_declaration no` but won't that mean that I'll miss out on any new implicit declaration of function errors introduced by future versions of freetds?

Yes. We should probably extract the function names it checks for, and provide both a global and a local whitelist. Otherwise this may just cause way too many warnings to be useful?

neverpanic:
@ryandesign Would you please try again with this version? I believe that should be sufficiently tweak-able that we can quickly hide the false positives.

neverpanic:
@ryandesign I haven't upgraded to Big Sur yet, so I'd like to get feedback before merging this.

ryandesign:
I haven't upgraded to Big Sur either, but I fake it by modifying MacPorts base as follows:
```
--- src/port1.0/portconfigure.tcl.orig
+++ src/port1.0/portconfigure.tcl
         configure.classpath
 # compiler flags section
 default configure.optflags      -Os
-default configure.cflags        {${configure.optflags}}
-default configure.objcflags     {${configure.optflags}}
+default configure.cflags        {${configure.optflags} -Werror=implicit-function-declaration}
+default configure.objcflags     {${configure.optflags} -Werror=implicit-function-declaration}
 default configure.cppflags      {[portconfigure::configure_get_cppflags]}
 proc portconfigure::configure_get_cppflags {} {
     global prefix
```
This at least works for ports that honor MacPorts CFLAGS.

neverpanic:
@ryandesign Thanks, that's helpful. I've tested running configure on my ports with this and am satisfied with the result. I'll wait for CI to pass and merge this afterwards.

PR #222, opened 2020-12-18
darwintrace: Support arbitrary path lengths

neverpanic:
darwintrace.dylib did occasionally crash in `__darwintrace_is_in_sandbox` due to stack canary violations, because the reimplementation of realpath and splitting into path components assumed that paths will not be longer than `MAXPATHLEN`. In practice, this assumption was wrong, and many build systems will invoke file system operations with paths longer than that.

Switch to dynamically allocated strings and arrays and resize the buffers as necessary before using them. Since this could actually move the location of `normPath` around in memory, switch from the `char *dst` that pointed into this `normPath` buffer to a `size_t dstoffset` that can be used relative to a moved `normPath`.

With these changes, quassel builds successfully in trace mode – without them, the build abort(3)s.

### DIFF
--- a/src/darwintracelib1.0/darwintrace.c
+++ b/src/darwintracelib1.0/darwintrace.c
@@ -711,6 +711,76 @@ static inline bool __darwintrace_sandbox_check(const char *path, int flags) {
 }
 
 /**
+ * Check whether the given buffer size is large enough, and realloc(3) the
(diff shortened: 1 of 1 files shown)


neverpanic (commented):
Leaky leaky memory allocations leak memory. Do not merge this as-is.

neverpanic:
I'll fix this with a bigger PR that also adds test coverage later on.

PR #223, opened 2020-12-28
Allow Universal on macOS 10.14 Mojave

Gcenx:
On MacOS Mojave don't block `+universal` instead require `macOS_SDK_headers_for_macOS_10.14.pkg` to be installed when a 32Bit package or `+universal` is called.

This is an alternative to the current approach of obtaining a copy of `MacOSX10.13.sdk` and modify `macports.conf`

Redone version of #213 (it was auto closed when attempting to rebase), now in a separate branch instead of master.


Gcenx:
@jmroot I've rebased this to HEAD

kencu:
I looked over the pkg, and indeed it seems to contain all the headers needed to make "/" function as the SDK on Mojave, and as far as I can tell, has all the i386-required parts still in the headers. I have to 100% verify this -- I haven't actually installed this as I would like to keep my Mojave system matching MacPorts as much as possible.

If it does have all the i386 bits as it seems to, all users who need to / want to keep i386 building possible could move up to Mojave, which is great.

My only concern is that the MacOSX10.14 SDK will not have i386 support in it, so builds that force that to be used (qt5, libvpx, etc, etc) will all fail building universal, whereas they succeed on 10.13 where the SDK matches "/".

Gcenx:
@kencu If these ports already respect SDKROOT then the modification to portutils.tcl should handle this already.

kencu:
Some do, some don't. the ones I mentioned will ignore "/" always, and force an SDK.

So it is not handled already, and -- because we cannot (I think) force the i386 headers into the MacOSX10.14.sdk, it never will be handled.

So I believe there will be a permanent mismatch between the SDK and "/", and this situation exists in no other system.

Right now, there is no SDK in "/", so you have only the MacOSX10.X.sdk to contend with.

kencu:
If we had 100% of ports  that would always respect SDKROOT, we would have an easier time of things. But we don't as you and I well know, and although you and I in particular have worked to make more of them respect it, lots don't.

Gcenx:
Just adding the i386 heading into the SDK won’t help without also adding the i386 slice into the .tbd files. Would your really want an end user to apply a patch/modify an SDK?

From previous conversations with you and other Macports maintainers if a Port is trying to set sdkroot on its own it should be patched to respect Macports sdkroot configuration.

For libvpx patching out the relevant sections in configure.sh resolves the issue.

kencu:
> Just adding the i386 heading into the SDK won’t help without also adding the i386 slice into the .tbd files. Would your really want an end user to apply a patch/modify an SDK?
> 

Of course not -- will never happen -- which is why I think we should just continue to have users use the MacOSX10.13.sdk as we do now, if they want i386 on Mojave.



> From previous conversations with you and other Macports maintainers if a Port is trying to set sdkroot on its own it should be patched to respect Macports sdkroot configuration.
> 
> For libvpx patching out the relevant sections in configure.sh resolves the issue.

There are hundreds. Try the qt5 ports, if you have 2 years and nothing to do :>

kencu:
Anyway, I am not in charge of MacPorts, that's just my $0.02, and leave this to @ryandesign or @jmroot to decide. Either way, I know my Mojave system is going to be fine, so it's all good to me.

Gcenx:
Requiring a copy of MacOSX10.13.sdk and setting the deployment target to 10.13 will have the same ports fail as installing the headers package on macOS Mojave.

This provides an alternative to needing to download Xcode_9.4.1 just for the SDK then modifying Macports.

kencu:
No -- that is not true. With this package installed, some ports will build universal OK on Mojave, and others will not (the ones that go to the SDK) and there will be no rhyme or reason to the end-user as to what the heck is going on and why things sometimes work and sometimes don't.

kencu:
With the 10.13 SDK, a person needs to specifically install it, set the SDK and the deployment target in macports.conf, and then  they presumably will be in the know as to what they have done, and at least have a shot at recognizing why their build of qt5-qtwebkit will not build universal on Mojave.

IMHO.

Gcenx:
> because we cannot (I think) force the i386 headers into the MacOSX10.14.sdk, it never will be handled.

I'd thrown together a Portfile `MacOSX10.14u.sdk` that only works on macOS Mojave and requires the headers package to be installed, this creates a Stub SDK that supports i386 & x86_64.

Then tested `libvpx` see [main.log](https://github.com/macports/macports-base/files/6565413/main.log), `macports-test` install only has this commit applied, `MacOSX10.14u.sdk` was installed using my regular macports installation.

src/port1.0/portutil.tcl
```diff
@@ -3407,7 +3412,21 @@ proc _check_xcode_version {} {
...
-            if {${os.major} >= 18 && [option configure.sdk_version] ne "" &&
+            # macOS Mojave can support i386 if macOS_SDK_headers_for_macOS_10.14.pkg is installed
+            # if headers package was not installed handle like like usual
+            # https://developer.apple.com/documentation/xcode-release-notes/xcode-10-release-notes
+            if {${os.major} == 18 && (![file isdirectory [file join $cltpath usr include]] || ![file executable  [file join $cltpath usr bin make]])} {
+                if { [option configure.sdkroot] eq "" } {
+                    ui_warn "The macOS [option configure.sdk_version] SDK is requested but configure.sdkroot is set to"
+                    ui_warn "a NULL string. Ports may not build correctly with this configuration."
+                } else {
+                    ui_warn "The macOS [option configure.sdk_version] SDK does not appear to be match the configured"
```
ryandesign, line 3423:
```suggestion
                    ui_warn "The macOS [option configure.sdk_version] SDK does not appear to match the configured"
```
This typo should also be fixed in the `${os.major} >= 19` block below.

pmetzger:
Sorry about that, @Gcenx. I know it's a bit dispiriting. However, we generally are trying to do a better job about tracking such things.

PR #224, opened 2021-01-03
portfetch: require MacPorts svn on >= Catalina

kencu:
there is a /usr/bin/svn on Catalina, but it is
only a small forwarding binary and is non-functional

```
 % /usr/bin/svn
svn: error: Failed to locate 'svn'.
svn: error: The subversion command line tools are no longer provided by Xcode.
```

closes: https://trac.macports.org/ticket/61981

(Note: personally I would probably rather just require macports' subversion on all systems for simplicity)

### DIFF
--- a/src/port1.0/portfetch.tcl
+++ b/src/port1.0/portfetch.tcl
@@ -188,7 +188,8 @@ proc portfetch::set_fetch_type {option action args} {
             }
             svn {
                 # Sierra is the first macOS version whose svn supports modern TLS cipher suites.
-                if {${os.major} >= 16 || ${os.platform} ne "darwin"} {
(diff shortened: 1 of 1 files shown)


ryandesign:
Whether `svn` works on Catalina isn't based on the OS version; it's based on the Xcode version.

kencu:
We can’t use xcodeversion for any tests as it doesn’t work if have only the CLT’s installed.

kencu:
so we don't like this? I can close it -- almost nothing uses svn any more anyway.

kencu:
I still think @ryandesign or @jmroot will have to do something like this to fix base, otherwise svn fetching will remain broken, and what's the good in that?

But they will come up with their own fix for it, in their own good time.

PR #225, opened 2021-01-07
lint --nitpick: ignore lines python.versions, ruby.branch(es)?

Tatsh:
Resolves https://trac.macports.org/ticket/61980

@ryandesign


kurthindenburg:
This looks OK to me.

ryandesign:
What about perl or php?

Tatsh:
> What about perl or php?

This has been updated to cover both.

PR #228, opened 2021-02-12
Replicate make 'and' function for GNU Make 3.80

michaelld:
MacOS X 10.4 has this version of GNU Make; MacOS X 10.5 to current provides version 3.81, which has a working 'and' function.

I honestly don't know if this is the best solution, but it works robustly in my testing on 10.4, 10.5, 10.12, and 11.3-beta. The 3 latter have GNU Make version 3.81, while the first has 3.80 & thus this issue.


kencu:
cool workaround -- I certainly don't speak "Make-ese" like you guys do. I'll try it out and report back.

neverpanic (approved):

PR #229, opened 2021-02-26
Add vercmp as a port action that compares two versions

kurthindenburg:
Expose pextlib's vercmp as a port action to allow users to determine
how macports calculates versions.

closes: https://trac.macports.org/ticket/61050


cooljeanius:
it seems there are some merge conflicts?

PR #230, opened 2021-03-23
Adjust messages to clarify what is being fetched.

kurthindenburg:
Add 'binary' and 'source' to the fetching message so users will
understand what is being fetched.

See: https://trac.macports.org/ticket/60756


cooljeanius:
Check about those merge conflicts?

PR #231, opened 2021-03-23
Skip any inaccessible file:// source links

kurthindenburg:
This allows continuing to next source quicker and avoids outputting a
lot of debug lines.


cooljeanius:
Personally I'd word the info message as "Failed to" instead of "Could not" (due to a general desire to avoid the word "not"), but that's a pretty minor nit; I'd consider the merge conflicts here to be the only real blockers...

PR #232, opened 2021-03-23
Make the reinplace warning an error

kurthindenburg:
This changes the warning "reinplace didn't change anything in" to an error.  This will force maintainers to fix these issues.  This can be bypassed by using the quiet (-q) option.  Two tests are added to verify the code.

Closes: https://trac.macports.org/ticket/60844

### DIFF
--- a/src/port1.0/portutil.tcl
+++ b/src/port1.0/portutil.tcl
@@ -1062,7 +1062,7 @@ proc reinplace {args}  {
         close $tmpfd
 
         if {!$quiet && ![catch {exec -ignorestderr cmp -s $file $tmpfile}]} {
-            ui_warn "[format [msgcat::mc "reinplace %1\$s didn't change anything in %2\$s"] $pattern $file]"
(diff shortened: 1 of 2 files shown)


ryandesign:
It's an easy change to the code, certainly, and I do want to make this change, but it's going to break so many ports.

barracuda156:
I think this should _not_ be a default option. Totally agree with @mascguy – opt-in is fine.

ryandesign:
> Just an example of port where `reinplace` is used to adjust different files. Not all of them contains pattern: https://github.com/macports/macports-ports/blob/master/emulators/xhyve/Portfile#L31-L36

Use `reinplace`'s `-q` flag in this case.

> Another example of port which uses `reinplace` to clean up platform-specified traces to make universal variant: https://github.com/macports/macports-ports/blob/master/devel/icu/Portfile#L98-L116

You can use `-q` here too if there's no better solution. A better solution might be to do the `reinplace` only if you know it will replace something. I haven't tried to determine if that's workable in that specific situation.

catap:
@ryandesign thanks for `-q` option. It is that I'm looking for.

PR #233, opened 2021-04-07
cregistry: Avoid segfault on extended result codes

neverpanic:
When an SQLite call returned one of the extended result codes (see https://sqlite.org/rescode.html#extended_result_code_list), the existing switch/case did not handle that, `errPtr` remained uninitialized, and following code crashed with a segmentation fault.

Avoid this by handling the extended result codes just like we're handling `SQLITE_ERROR`. Additionally, default-initialize the `errPtr` structure to avoid crashes in the future.

This manifested itself when attempting to uninstall ports from a database that returned `SQLITE_CORRUPT_INDEX` (779).


jmroot (approved):
Thanks, LGTM. I guess this was an accidental consequence of switching to `sqlite3_prepare_v2`.

neverpanic:
Guess so, yeah.

PR #235, opened 2021-04-21
Add use_autoreconf option to portfile man page

gpanders:
`use_automake` and `use_autoconf` were already in there, but `use_autoreconf` got left out somehow.


jmroot (approved):
Thanks!

PR #236, opened 2021-04-24
Fix 'use_parallel_build no' builds with ninja which requires -j1 to disable parallelism

cjones051073:
Currently `portbuild::build_getjobsarg` returns an empty string if `use_parallel_build no` is set. 
Ninja though defaults to enabling parallel builds so to disable them you have to explicitly pass `-j1`. 
This PR implements this.

### DIFF
--- a/src/port1.0/portbuild.tcl
+++ b/src/port1.0/portbuild.tcl
@@ -130,8 +130,13 @@ proc portbuild::build_getmaketype {args} {
 }
 
 proc portbuild::build_getjobs {args} {
-    global buildmakejobs
+    global buildmakejobs use_parallel_build
     set jobs $buildmakejobs
+    # If parallel disabled disabled, return 1
+    if {![tbool use_parallel_build]} {
+        ui_debug "port disallows a parallel build, setting build jobs to 1"
+        set jobs 1
+    }
     # if set to '0', use the number of cores for the number of jobs
     if {$jobs == 0} {
         try -pass_signal {
@@ -167,21 +172,15 @@ proc portbuild::build_getargs {args} {
 }
 
 proc portbuild::build_getjobsarg {args} {
-    # check if port allows a parallel build
-    global use_parallel_build
-    if {![tbool use_parallel_build]} {
-        ui_debug "port disallows a parallel build"
-        return ""
-    }
-
     if {![exists build.jobs] || \
             !([string match "*make*" [option build.cmd]] || \
               [string match "*ninja*" [option build.cmd]] || \
               [string match "*scons*" [option build.cmd]])} {
         return ""
     }
+    
     set jobs [option build.jobs]
-    if {![string is integer -strict $jobs] || $jobs <= 1} {
+    if {![string is integer -strict $jobs] || $jobs < 1} {
         return ""
     }
     return " -j$jobs"


src/port1.0/portbuild.tcl
```diff
@@ -171,7 +171,13 @@ proc portbuild::build_getjobsarg {args} {
...
     if {![tbool use_parallel_build]} {
         ui_debug "port disallows a parallel build"
-        return ""
+        # Ninja defaults to enabling parallelism if -jX is omitted
+        # so need to explicitly pass -j1 to disable this
+        if { [string match "*ninja*" [option build.cmd]] } {
+            return " -j1"
+        } else {
+            return ""
+        }
```
ryandesign, line 180: Due to the below change, we'll already be using `-j1` for make/scons sometimes, so why not do it here as well?
```suggestion
        return " -j1"
```

src/port1.0/portbuild.tcl
```diff
@@ -181,7 +187,7 @@ proc portbuild::build_getjobsarg {args} {
         return ""
     }
     set jobs [option build.jobs]
-    if {![string is integer -strict $jobs] || $jobs <= 1} {
+    if {![string is integer -strict $jobs] || $jobs < 1} {
         return ""
     }
```
ryandesign, line 192: This will cause `-j1` to be used for make/ninja/scons if `build.jobs` is 1, which seems fine to me. But actually there should be no circumstance in which `build.jobs` could be < 1, so this could be simplified:
```suggestion
```

ryandesign: Well, looking a little further, a user can set `buildmakejobs` in macports.conf. If they set it to `0`, we set it to the number of active processors or `2` if we can't determine it. But this might've been the only code that handles if a user set `buildmakejobs` to a negative integer. There's also the remote possibility that a port or portgroup could have set `build.jobs` to `0` or a negative integer, and it's probably good to handle that here too.

There is still other invalid input that we probably don't handle, such as `buildmakejobs` or `build.jobs` being set to a non-integer number or string.

src/port1.0/portbuild.tcl
cjones051073: The final change is after the check

```
   if {![exists build.jobs] || \
            !([string match "*make*" [option build.cmd]] || \
              [string match "*ninja*" [option build.cmd]] || \
              [string match "*scons*" [option build.cmd]])} {
        return ""
    }
```

so we know the build system is one of three, that support -j. I didn't do it here as I presumed there might be other build systems not supporting -j.

cjones051073:
I've reworked things a little, to perform the parallel build check in `portbuild::build_getjobs` and to set `build.jobs` to 1 in this case. This then simplifies what `portbuild::build_getjobsarg` needs to do, and is also more consistent as ports then downstream will see `build.jobs=1` if `use_parallel_build=no`, and will avoid the need slightly peculiar workarounds like 

https://github.com/macports/macports-ports/blob/afff7bef94faaf3db8e6ff53276c1ae4295b13c2/_resources/port1.0/group/cmake-1.1.tcl#L165

cjones051073:
Any further comments ? Any objections to merging this ?

cjones051073:
Been testing this locally for a while now so i believe its good to go.

PR #237, opened 2021-05-01
Fix libmd linux

Ionic:
The standalone `libmd` implementation as used on Linux includes a header
called `sha2.h` which is supposed to be relative to its own include
directory.

Unfortunately, MacPorts ships its own SHA2 implementation in pextlib (in
case neither `CommonCrypto`, nor `libmd` nor `OpenSSL`-like implementations
are available), which also uses the `sha2.h` header file name. This
seems to take precedence and both header files are not compatible, so
the build fails.

Move the standalone SHA2 implementation to `src/compat/sha2` and update
the references.

Additionally, `SHA1_File` is almost never a proper function, but most
often defined as a macro set to the real function. In any case, checking
for it via `AC_CHECK_FUNCS()` is wrong, so switch to `AC_CHECK_DECLS()` and
use the updated checking macros.

Fixes: https://trac.macports.org/ticket/58060


harens:
Just rebased the PR. I setup an Ubuntu GitHub actions workflow to test whether these commits fix the issue.

- Before: The [Linux build fails](https://github.com/harens/macports-base/runs/5870024678?check_suite_focus=true), as indicated by the [referenced ticket](https://trac.macports.org/ticket/58060).
- With this PR: The [build succeeds](https://github.com/harens/macports-base/runs/5870923094?check_suite_focus=true). The tests fail, but that isn't to do with this PR.

Therefore, this PR seems good to me. Thank you for your work on this @Ionic.

harens (approved):

PR #240, opened 2021-05-16
add licensecheck target, copied from macports-infrastructure; updated work from defunct PR #211

mascguy:
This PR encompasses the excellent work by @danchr, from defunct [PR 211](https://github.com/macports/macports-base/pull/211). It includes one of the key requested changes, which is to source license data from `_resources`.

A separate PR will be submitted for `macports-ports`, encompassing both the aforementioned license data, as well as updated BASH completions.


src/port1.0/portlicensecheck.tcl
```diff
@@ -0,0 +1,276 @@
...
+    }
+}
+
+proc portlicensecheck::licensecheck_start {args} {
+    global UI_PREFIX subport
+    ui_notice "$UI_PREFIX [format [msgcat::mc "Checking license for %s"] ${subport}]"
+}
+
+proc portlicensecheck::licensecheck_main {args} {
+    # Obtain licenses from '_resources/port1.0/licenses/licenses.tcl'.
```
mascguy, line 113: NOTE: As a starting point, I named the new resource area `licenses`, to be more general. With the idea being that the data might be used elsewhere at some point.

But if folks prefer that it be specific to this target, it can be renamed to `licensecheck`.

jmroot:
Moving conversation from #211.

> In terms of sharing the data and logic, can the MacPorts infrastructure scripts utilize MacPorts base?

Sure. That has the (perhaps minor) drawback that it can then only be updated with a base release, of course.

> (For example, simplifying distributable_lib.tcl to delegate the license check to port licensecheck.)

That sounds backwards.

> That would alleviate all of your concerns, correct?

Having base ship `distributable_lib.tcl` (without removing features) and use it to implement the new action would go a long way.

mascguy:
> > In terms of sharing the data and logic, can the MacPorts infrastructure scripts utilize MacPorts base?
> 
> Sure. That has the (perhaps minor) drawback that it can then only be updated with a base release, of course.

Certainly, and we wouldn't want to be tied to a MacPorts base release. To eliminate that, most of the core logic can be migrated to `_resources`.

jmroot:
> Certainly, and we wouldn't want to be tied to a MacPorts base release. To eliminate that, could we simply migrate most of the core logic to `_resources`?

The drawback then is that you can't run it without a copy of the official ports tree. (Like when running tests, or when using only a custom ports tree.)

mascguy:
Do we really want to bake all of this into the MacPorts release, with no dependence on `_resources`? Virtually all other data for MacPorts base,  which we routinely want to change, resides in the latter.

Imagine, for example, if resource `fetch/archive_sites.tcl` were embedded in the release, rather than `_resources`. That would be horrible!

mascguy:
Reflecting on this for a few days, I had one other thought: What if there was a port command/option - either documented or undocumented, depending on your folks' preference - that only populates the `_resources` area?

Ideally we would publish/mirror an ultra-light-weight resources archive. But given the logistics and touch points involved in such a venture, we could keep it simple... and only extract resources from the existing ports tarball. That would limit the changes to base.

Assuming we don't have such functionality already (?), would that be a possible compromise...?

mascguy:
After still more reflection, I've come to a key realization: License info rarely changes. So while it would be nice to source that data from the `_resources` area, baking it into the release isn't a big deal.

So I'm willing to integrate into all into base.

@jmr Josh, can you provide an example, relative to how you'd like to be able to use this functionality from a utility script? Basically I just need the template of how I could test/validate. That said, you are also more than welcome to directly push changes to this branch. (My guess is that you could quickly tweak this stuff in 15 minutes or less, for your needs. And while I enjoy learning all of this, I'd also love to get this functionality integrated into base. So up to you!)

Is it essentially something like the following, per your "find port dependents by exact name/regex" example you sent to the mailing list?

```
#!/usr/bin/env port-tclsh

package require macports
mportinit

[...interaction with new licensecheck functionality...]
```

danchr:
> After still more reflection, I've come to a key realization: License info rarely changes. So while it would be nice to source that data from the `_resources` area, baking it into the release isn't a big deal.

Just to back this up; AFAICT the license lists have been unmodified since 2015, when commit https://github.com/macports/macports-infrastructure/commit/87d4d80eff4979cbb679859c04ee6743c25506ec added some Creative Commons licenses.

jmroot:
> @jmr Josh, can you provide an example, relative to how you'd like to be able to use this functionality from a utility script? 

This is the main existing client: https://github.com/macports/mpbb/blob/master/tools/sort-with-subports.tcl

There's another one I'm planning to add to the new buildbot config, which I can show you if you need it, though it uses `distributable_lib.tcl` in much the same way as the first one.

PR #241, opened 2021-05-22
show if the port is installed in the info and the search command

ra1nb0w:
This is a small enhancement that I found useful: it adds an `[installed]` tag in the info and search command.
In the info the tag is shown only with the heading print.
In the search the tag is shown only in multiline print.

Can be also useful to implement it in the other modes? Maybe, they are used only in scripting and therefore you can call `port installed`
Can be also useful to take in account the active status?

### DIFF
--- a/src/port/port.tcl
+++ b/src/port/port.tcl
@@ -2162,6 +2162,9 @@ proc action_info { action portlist opts } {
                 if {[info exists portinfo(categories)]} {
                     append inf " ([join $portinfo(categories) ", "])"
                 }
+                if {![catch {registry::installed $portinfo(name) {}}]} {
+                    append inf " \[installed\]"
+                }
             } elseif {$opt eq "fullname"} {
                 set inf "$portinfo(name) @"
                 append inf [composite_version $portinfo(version) $portinfo(active_variants)]
@@ -3838,6 +3841,9 @@ proc action_search { action portlist opts } {
                     }
                     if {[info exists portinfo(categories)]} {
                         puts -nonewline " ([join $portinfo(categories) ", "])"
+                    }
+                    if {![catch {registry::installed $portinfo(name) {}}]} {
+                        puts -nonewline " \[installed\]"
                     }
                     puts ""
                     puts [wrap [join $portinfo(description)] 0 [string repeat " " 4]]


raimue:
Hm, I think this could be slightly misleading because the version shown must not be the version that is installed.

ra1nb0w:
Yes @raimue. I thought about that but `[installed]` should be only an indication that the port is installed without referring to which version but you are right, can be slightly mislead. Use `[upgradable]` if a new version is available? What do you think?

ryandesign:
I mean we could go crazy and show a label for every pseudoport that this port belongs to. `[installed]` `[outdated]` `[active]` `[leaf]` `[obsolete]` `[requested]`

ra1nb0w:
@ryandesign sarcastic?
Since I mainly use `search` and `info` my starting idea was to show there if a package is installed or not without calling `installed` (just like `apt*`). Nothing too complicated (also to avoid slowdown).

raimue:
Just some brainstorming on this idea. I would suggest not to put even more stuff into the first line. How about adding another line for this information instead?

```
$ port info dummy
dummy @1.2 (security, net)
State: installed, active, requested, outdated (@1.1_3)
Variants: universal
...
```

But not sure whether `State: uninstalled` makes sense? Probably better to leave it out in this case?

In `port search` it could be shown in an abbreviated form to keep the format concise. Unfortunately, "outdated" and "obsolete" cause a collision on the first letter... But as of now, `port search` would not show obsolete ports anyway, so it should be clear enough.

```
$ port search dummy
dummy @1.2 (games) [iaro]
    Just a dummy port
```

mascguy:
I'm definitely supportive of the general idea!

ra1nb0w:
For `info` can be a good idea to move to a new line with full State (maintain the same UI) except `uninstalled`. In this case hide the line.
For `search`: I don't like abbreviations since you need to know it and since the command is used by anyone it should be auto-explanatory. Just one point: is not too much cpu intensive checks the full state in the search since it can be quite big?

eugenesvk:
Could you maybe use symbols similar to what Homebrew does (though only for dependencies)?
- green ✔ or ✓ for installed
- red ✘ or ✗ for 

They're much easier to notice that just the words

gulfcoastpaintingp (changes requested):
Every time someone has to take out money on me chat I want a fee

PR #243, opened 2021-05-30
base: add lower-level debug output, via new levels debug2/debug3

mascguy:
Per recent discussion on the dev mailing list, add lower-level debug logging support - via calls `ui_debug2` and `ui_debug3` - which are filtered out from our existing debug output level.

The goal is to reduce the verbosity during day-to-day work, whilst still being able to enable the additional output via levels `d2` and `d3`, when desired.

Proposed new arguments to `port`:
- `-d2`: Enable output through debug2; debug3 filtered out.
- `-d3`: Enable output through debug3.

Note: The choice to support two additional levels is arbitrary, and we could go further if we wish...supporting levels up to debug9, for example. But two additional levels provides plenty of granularity as-is.

Also note that this PR is very much a work-in-progress, and not complete. But it provides a place for collaborative discussion.


src/port1.0/port.tcl
```diff
@@ -54,9 +54,9 @@ namespace eval port {
 	proc run_callbacks {} {
 		variable _callback_list
 		foreach callback ${_callback_list} {
-			ui_debug "Running callback ${callback}"
+			ui_debug2 "Running callback ${callback}"
 			${callback}
-			ui_debug "Finished running callback ${callback}"
+			ui_debug2 "Finished running callback ${callback}"
 		}
```
mascguy, line 60: This is one suggested change, if we ultimately decide to merge this work. Though there are plenty of other opportunities in our portgroups too.

src/port1.0/portutil.tcl
```diff
@@ -2618,7 +2618,7 @@ proc PortGroup {group version} {
     if {[file exists $groupFile]} {
         lappend PortInfo(portgroups) [list $group $version $groupFile]
         uplevel "source $groupFile"
-        ui_debug "Sourcing PortGroup $group $version from $groupFile"
+        ui_debug2 "Sourcing PortGroup $group $version from $groupFile"
```
mascguy, line 2621: Hiding output for portgroup sourcing, is another potential.

src/port1.0/port.tcl
```diff
@@ -54,9 +54,9 @@ namespace eval port {
 	proc run_callbacks {} {
 		variable _callback_list
 		foreach callback ${_callback_list} {
-			ui_debug "Running callback ${callback}"
+			ui_debug2 "Running callback ${callback}"
 			${callback}
-			ui_debug "Finished running callback ${callback}"
+			ui_debug2 "Finished running callback ${callback}"
```
mascguy, line 59: Hiding output related to running callbacks, is one potential change.

src/port/port.tcl
```diff
@@ -4577,6 +4577,19 @@ proc parse_options { action ui_options_name global_options_name } {
                         # debug implies verbose
                         set ui_options(ports_verbose) yes
                     }
+                    d2 {
+                        set ui_options(ports_debug) yes
+                        set ui_options(ports_debug2) yes
+                        # debug implies verbose
+                        set ui_options(ports_verbose) yes
+                    }
+                    d3 {
```
mascguy, line 4586: Can potentially flatten the `dX` checks into one additional case, via a regex on argument string - something like `d[2-3]`?

Along with that, we could also set option `ports_debug_x` to the numeric level of what's enabled. That eliminates having to hard-code multiple option checks for the additional levels.

src/macports1.0/macports.tcl
```diff
@@ -339,14 +339,21 @@ proc macports::ui_prefix_default {priority} {
...
-    switch -- $priority {
+    switch -regexp -- $priority {
         debug {
             if {[ui_isset ports_debug]} {
                 return stderr
             } else {
                 return {}
             }
         }
+        debug[0-9] {
```
mascguy, line 350: Internally, levels 0-9 supported. But range 0-3 enforced via argument checks.

src/port/port.tcl
```diff
@@ -4577,6 +4577,15 @@ proc parse_options { action ui_options_name global_options_name } {
                         # debug implies verbose
                         set ui_options(ports_verbose) yes
                     }
+                    d[0-3] {
```
mascguy, line 4580: This won't work as-is, as argument processing occurs one character at a time. Need to pre-process the arguments string, and ensure the two-character `d<level>` combo is passed through.

mascguy:
@jmr Josh, I'm very close to having this working. There's just one key thing I haven't figured out: How are the new `ui_debugX` procedures made available in the TCL realm?

mascguy:
> Josh/Anyone, I'm very close to having this working. There's just one key thing I haven't figured out: How are the new `ui_debugX` procedures exported to the TCL realm?
> 
> I was expecting to see something related to the use of `Tcl_CreateObjCommand`. But while that's used for other calls, I couldn't find (or missed) anything related to the various `ui_xxx` logging calls. Despite grepping through every file in the base sources...

Never mind, finally figured it out. LOL

mascguy:
Update: This is working!

There's still some cleanup to do, but this is starting to approach a finished state:
- Argument Processing: Presently we deal with short args char-by-char, so need to make changes to support `-dX`.
- Regression Testing: Need to validate various touch points, such as `packageall.tc' and `portindex.tcl`.

But looking good overall.

Thoughts/comments welcome, and appreciated!

mascguy:
Example output, with and without this PR.

Current state, without:
```
$ port -d info --line --fullname --description mpich
DEBUG: Changing to port directory: /opt/local/var/macports/sources/rsync.macports.org/macports/release/tarballs/ports/science/mpich
DEBUG: OS darwin/17.7.0 (macOS 10.13.6) arch i386
DEBUG: Sourcing PortGroup active_variants 1.1 from /opt/local/var/macports/sources/rsync.macports.org/macports/release/tarballs/ports/_resources/port1.0/group/active_variants-1.1.tcl
DEBUG: GCC versions for Darwin 17 i386 - 5 6 7 8 9 10 11 devel
DEBUG: Clang versions for Darwin 17 i386 - 5.0 6.0 7.0 8.0 9.0 10 11 devel
DEBUG: Sourcing PortGroup compilers 1.0 from /opt/local/var/macports/sources/rsync.macports.org/macports/release/tarballs/ports/_resources/port1.0/group/compilers-1.0.tcl
DEBUG: Sourcing PortGroup mpiutil 1.0 from /opt/local/var/macports/sources/rsync.macports.org/macports/release/tarballs/ports/_resources/port1.0/group/mpiutil-1.0.tcl
DEBUG: universal_variant is false, so not adding the default universal variant
DEBUG: Running callback portconfigure::add_automatic_compiler_dependencies
DEBUG: Finished running callback portconfigure::add_automatic_compiler_dependencies
DEBUG: Running callback portbuild::add_automatic_buildsystem_dependencies
DEBUG: Finished running callback portbuild::add_automatic_buildsystem_dependencies
DEBUG: Running callback portstartupitem::add_notes
DEBUG: Finished running callback portstartupitem::add_notes
DEBUG: Running callback compilers::add_fortran_legacy_support
DEBUG: Finished running callback compilers::add_fortran_legacy_support
mpich @3.4.2	Message Passing Interface (MPI) Library
```

With the changes, at standard debug:
```
$ port -d info --line --fullname --description mpich
DEBUG: Changing to port directory: /opt/macports-debug3/var/macports/sources/rsync.macports.org/macports/release/tarballs/ports/science/mpich
DEBUG: OS darwin/17.7.0 (macOS 10.13.6) arch i386
DEBUG: GCC versions for Darwin 17 i386 - 5 6 7 8 9 10 11 devel
DEBUG: Clang versions for Darwin 17 i386 - 5.0 6.0 7.0 8.0 9.0 10 11 devel
DEBUG: universal_variant is false, so not adding the default universal variant
mpich @3.4.2	Message Passing Interface (MPI) Library
```

With the changes, at debug level 3:
```
$ port -3 info --line --fullname --description mpich
DEBUG: Changing to port directory: /opt/macports-debug3/var/macports/sources/rsync.macports.org/macports/release/tarballs/ports/science/mpich
DEBUG: OS darwin/17.7.0 (macOS 10.13.6) arch i386
DEBUG1: Sourcing PortGroup active_variants 1.1 from /opt/macports-debug3/var/macports/sources/rsync.macports.org/macports/release/tarballs/ports/_resources/port1.0/group/active_variants-1.1.tcl
DEBUG: GCC versions for Darwin 17 i386 - 5 6 7 8 9 10 11 devel
DEBUG: Clang versions for Darwin 17 i386 - 5.0 6.0 7.0 8.0 9.0 10 11 devel
DEBUG1: Sourcing PortGroup compilers 1.0 from /opt/macports-debug3/var/macports/sources/rsync.macports.org/macports/release/tarballs/ports/_resources/port1.0/group/compilers-1.0.tcl
DEBUG1: Sourcing PortGroup mpiutil 1.0 from /opt/macports-debug3/var/macports/sources/rsync.macports.org/macports/release/tarballs/ports/_resources/port1.0/group/mpiutil-1.0.tcl
DEBUG: universal_variant is false, so not adding the default universal variant
DEBUG1: Running callback portconfigure::add_automatic_compiler_dependencies
DEBUG1: Finished running callback portconfigure::add_automatic_compiler_dependencies
DEBUG1: Running callback portbuild::add_automatic_buildsystem_dependencies
DEBUG1: Finished running callback portbuild::add_automatic_buildsystem_dependencies
DEBUG1: Running callback portstartupitem::add_notes
DEBUG1: Finished running callback portstartupitem::add_notes
DEBUG1: Running callback compilers::add_fortran_legacy_support
DEBUG1: Finished running callback compilers::add_fortran_legacy_support
mpich @3.4.2	Message Passing Interface (MPI) Library
```

src/pextlib1.0/Pextlib.c
```diff
@@ -135,6 +135,14 @@ static void ui_message(Tcl_Interp *interp, const char *severity, const char *for
     free(tclcmd);
 }
 
+__attribute__((format(printf, 3, 0)))
+static void ui_debug_x(Tcl_Interp *interp, unsigned int level, const char *format, va_list va) {
+    char cLevel[20];  // longer than necessary, but also safer
+
+    sprintf(cLevel, "debug%u", level);
```
raimue, line 142: Please use `asprintf` as in the other functions or at least `snprintf`. Actually, I would just use a lookup table to map the level to the three possible strings.

src/port/port.tcl
```diff
@@ -4577,6 +4577,23 @@ proc parse_options { action ui_options_name global_options_name } {
                         # debug implies verbose
                         set ui_options(ports_verbose) yes
                     }
+                    [0-3] {
+                        #d[0-3]
+                        # TEMP HACK: Since args processed char-by-char, just use single-digit
+                        # numeric arg for now. To fix, we either need to pre-process the args,
+                        # and ensure 'dX' is combined before this switch. Or use non-destructive
+                        # lookahead.
```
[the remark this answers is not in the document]
mascguy: Well, the problem with processing args one character at a time, without any context/state, is that they can be repeated ad infinitum. As it stands today, without this patch, the following veriations are valid. Or at least, we don't validate whether they make sense or not:

```
$ port -dddddddddddddddddddddddddddddd info my_port
$ port -d -d -d -d -d -d -d -d -d -d info my_port
```

You get the picture.

I'd prefer that we not allow that, actually. As well as preprocess the args for cases like the debug level. (And combine multi-character args - like `dX` - before passing them through the switch statement.)

src/pextlib1.0/Pextlib.c
mascguy: Fair enough, either would be preferable.

src/port/port.tcl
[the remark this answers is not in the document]
mascguy: As for repeating `-d` multiple times, to specify the level, hmmm.

Yes, that paradigm is sometimes used/allowed in command-line utilities. But it seems clearer to simply specify -dX, with the numeric level. This would be even more important if we allowed levels beyond debug3, as it gets a bit ridiculous after a while.

I mean, we'd have to specify `-dddd` for debug3? Yeah, no thanks. It's difficult to read, and you have to carefully count the characters. Perhaps it's usable when maximizing out at three levels, but definitely wouldn't be my choice.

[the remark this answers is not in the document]
raimue: Sorry, I wasn't clear. Yes, I meant the `port -ddd` repetition as a way to specify the debug level. I would prefer this over an option with a digit argument.

Having a `-d[X]` that can be used with or without an argument is a bad choice. If an option does not take an argument, then it can be combined freely with any other option of that kind into a single `-abc...`. If an option takes an argument, then that follows in the next argument, not immediately after the letter. These rules are simple, but "`-d` can be combined with other options, but if the next letter is a digit, then it is treated differently" complicates both parsing and documentation. For example, `-dyp` is equivalent to `-d -y -p`, but `-d3yp` would be `-d3 -y -p`.

For the proposed `-dX` you would have to look at the argument that follows and decide whether this is valid or not and then treat it differently. This mixes the syntax parsing of options with their semantics and I do not like that. It will only work for this one special case where `-d` takes a single digit. I do not see how allowing this would help with any other option in the future as it cannot be generalized. It also forbids `-0` to `-9` from ever being used as an option, as it would be ambiguous.

Therefore the only choice would be `-d X` and make it *always* take this argument. However, as I wouldn't want to change the existing `-d`, we would even have to settle on a different letter to set the debug log level... And then we end up with both `-d` and `-L 3` (or whatever).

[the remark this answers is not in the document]
mascguy: Rainer, I'm very open to whatever you folks want to use, in terms of arguments. If that's what it takes to get this merged, I'll implement whatever argument style you folks wish!

Repetition is certainly fine. The single-digit approach we're using now works well too.

Tell me what you folks prefer, and I'll do it!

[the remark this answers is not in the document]
mascguy: In other words, "What can I do to get you to buy this new car TODAY?"  :-)

src/pextlib1.0/Pextlib.c
mascguy: Done, changed to use `snprintf`.

cjones051073:
Having 4 levels of debug seems a bit excessive to me. I would have thought that say 3 would easily be enough granularity? I also find having level ‘0’ called debug, whilst 1-3 called debugN a little confusing. I would personally suggest making debug and debug1 the same level (so just make debug an alias to debug1 for compatibility) and then have debug2 and debug3 for two increasingly more detailed levels.

mascguy:
> Having 4 levels of debug seems a bit excessive to me. I would have thought that say 3 would easily be enough granularity? I also find having level ‘0’ called debug, whilst 1-3 called debugN a little confusing. I would personally suggest making debug and debug1 the same level (so just make debug an alias to debug1 for compatibility) and then have debug2 and debug3 for two increasingly more detailed levels.

Sure, and three levels isn't set in stone; that was merely an arbitrary choice to start with.

Overall, this effort is still evolving, and the goal is to support logging based on namespaces/packages - similar to Python and Java - to allow more control over output and levels. (Per recommendations by folks on the mailing list.)

The latter hasn't been implemented yet, but watch this space...

mascguy:
Relative to namespace/package logging, I'm currently looking at the TCL `logger` package, to avoid reinventing the wheel. Based on a cursory look, it may provide a great foundation for what we're trying to accomplish.

If anyone has any experience with that package - and/or knows of a better option - please let me know!

src/pextlib1.0/tests/curl.tcl
```diff
@@ -32,8 +32,9 @@ proc main {pextlibname} {
     #curl fetch -u "I accept www.opensource.org/licenses/cpl:." http://www.research.att.com/~gsf/download/tgz/sfio.2005-02-01.tgz $tempfile
     #test {[md5 file $tempfile] == "48f45c7c77c23ab0ccca48c22b3870de"}
 
-    curl fetch --enable-compression https://www.whatsmyip.org/http-compression-test/ $tempfile
-    grepper $tempfile {gz_yes}
+    # CNielsen: Temporarily disable compressed fetch, as failing after 30-second timeout
+    #curl fetch --enable-compression https://www.whatsmyip.org/http-compression-test/ $tempfile
+    #grepper $tempfile {gz_yes}
```
cjones051073, line 37: This doesn’t seem directly related to the debug level changes, or am I missing something ?

cjones051073 (approved):
Seems OK to me, thanks for this.

src/pextlib1.0/tests/curl.tcl
mascguy: This was necessary back a month or so ago, as the test consistently failed due to timeout. (It wasn't intended to be kept long-term though.)

Ultimately it will be reverted in this PR, or changed to use a different testing endpoint if necessary. (The latter would be based on guidance from Josh and/or Ryan though.)

cooljeanius:
> Having 4 levels of debug seems a bit excessive to me. I would have thought that say 3 would easily be enough granularity?

4 levels is how many Fink has, IIRC...

cooljeanius:
> Relative to namespace-scoped logging - something not currently included - I'm wondering if we should shelve that at this point?

Yeah, probably worth saving that for a separate PR.

src/pextlib1.0/tests/curl.tcl
cooljeanius: hm, @jmroot @ryandesign any guidance?

PR #245, opened 2021-08-22
Add opt-in Altivec (ppc7400) architecture

evanmiller:
Address https://github.com/macports/macports-ports/pull/11625 with a modest proposal to add a new `ppc7400` architecture. The idea is that this Altivec-enabled architecture will be preferred on G4/G5 machines but fall back to regular `ppc` if not available. When properly implemented in Portfiles, G3 users should no longer see the dreaded `incompatible cpu subtype` run-time error message.

Support for the `ppc7400` architecture is completely controlled by the individual Portfiles, which need to declare support in `supported_archs` and then pass appropriate configure and compiler flags for Altivec (`ppc7400`) and non-Altivec (`ppc`) builds. Existing Portfiles do not require modification (it's OK if `ppc` builds are secretly Altivec builds), but this PR will provide the tools for maintainers to replace run-time hardware detection with branching on `${configure.build_arch}`, or to declare that a port absolutely, positively requires Altivec on PowerPC (e.g. `mpg123`). See the commit message for a more complete explanation and rationale.

Rudimentary testing performed but I do not have access to a G4/G5, and I also have no idea if universal builds will work here. Earlier, someone suggested supporting ppc + ppc7400 universal builds, which would be a neat trick for binary distributions.

Opening this PR as a draft to discuss the concept and field suggestions. Note that this general strategy might also be used for advanced Intel ISAs (e.g. AVX2 and AVX-512) though that's outside the current scope here.

### DIFF
--- a/doc/macports.conf.5
+++ b/doc/macports.conf.5
@@ -343,7 +343,7 @@ lt lt.
 T{
 \fBRegular architectures include:\fR
 T}:T{
-ppc, i386, ppc64, x86_64, arm64
+ppc, ppc7400, ppc64, i386, x86_64, arm64
 T}
 T{
 \fBDefault (11 and later):\fR
@@ -376,7 +376,7 @@ lt lt.
 T{
 \fBRegular architectures include:\fR
 T}:T{
-ppc, i386, ppc64, x86_64, arm64
+ppc, ppc7400, ppc64, i386, x86_64, arm64
 T}
 T{
 \fBDefault (11 and later):\fR
--- a/doc/macports.conf.5.txt
+++ b/doc/macports.conf.5.txt
@@ -124,7 +124,7 @@ keeplogs::
 
 build_arch::
     The machine architecture to try to build for in normal use.
-    *Regular architectures include:*;; ppc, i386, ppc64, x86_64, arm64
+    *Regular architectures include:*;; ppc, ppc7400, ppc64, i386, x86_64, arm64
     *Default (11 and later):*;; arm64 or x86_64 depending on hardware
     *Default (10.6-10.15):*;; x86_64 or i386 depending on hardware
     *Default (10.5 and earlier):*;; i386 or ppc depending on hardware
@@ -133,7 +133,7 @@ universal_archs::
     The machine architectures to use for +universal variant (multiple
     architecture entries should be space separated). Should contain at
     least two entries, or be empty to disable universal building.
-    *Regular architectures include:*;; ppc, i386, ppc64, x86_64, arm64
+    *Regular architectures include:*;; ppc, ppc7400, ppc64, i386, x86_64, arm64
     *Default (11 and later):*;; arm64 x86_64
     *Default (10.6-10.13):*;; x86_64 i386
     *Default (10.5 and earlier):*;; i386 ppc
--- a/doc/portfile.7
+++ b/doc/portfile.7
@@ -230,6 +230,7 @@ Archs currently supported by macOS are:
 .Em i386 ,
 .Em ppc ,
 .Em ppc64 ,
+.Em ppc7400 ,
 and
 .Em x86_64 .
 If this option is not set, it is assumed that the port can build for all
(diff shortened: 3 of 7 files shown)


kencu:
The problem I see with this is there are hundreds of ports that run the list of supported arches, branching on the build_arch or configure.build_arch.

ppc7400 is not recognized by any of these Portfiles, and I doubt they want to add specs for this for the 0.0001% coverage it would add.

IMO the simplest thing to do in the relatively few ports that build differently with or without altivec is something like this in the PPC block:

```
if (altivec available) {
  add_altivec_enabling_option whatever_it_is
} else {
  # clear archive sites forcing a local build should a buildbot someday exist
  archives_sites
}
```

This presumes only that any PPC buildbot someone sets up would be at least an altivec-capable machine, and we can put that as a requirement someplace if we like. I don't think is unreasonable at all.

It gets around this theoretical problem very clearly, with minimal disruption, and no complicated variants.

evanmiller:
Just to be clear, `ppc7400` is not included in the new default value of `supported_archs` (see change to src/port1.0/portmain.tcl). So ports that iterate over supported_archs will not see it - or am I missing something? (If ports are iterating over `sdk_archs` that is a different matter...)

kencu:
so let's say you set up your G3 system with a build_arch of ppc7400 specified in macports.conf (I assume that is what you mean users to do with this).

And you get to one of the many many ports that iterates over the known build_archs (ppc, ppc64, i386, x86_64, arm64 at present). There is no entry in any of those Portfiles for ppc7400, so the branching logic that branches on the build_arch will fail.

Modifying them all is not reasonable to ask.

The solution I suggest is incredibly simple and unobtrusive, and solves the problem fairly elegantly (with the proviso that a PPC buildbot would have to be a G4 or better). It seems so much simpler to me to do that.

kencu:
Sorry I guess I had it backwards. G3 users you propose would leave their build_arch as the default ppc, and anyone with a G4 or G5 be asked to set their system up with ppc7400 in macports.conf to get altivec? Perhaps I am not fully following how this might work transparently without any changes in any portfiles.

I still think clearing archive_sites on G3s when needed is a very simple solution to the issue though.

evanmiller:
ppc vs ppc7400 is now auto-selected (via `[sysctl hw.vectorunit]`) in src/macports1.0/macports.tcl – no changes are needed to anyone's macports.conf.

The idea is that ppc is a fall-back architecture for ppc7400, just as i386 is a fallback architecture for x86_64 (and ppc is a fallback for ppc64). Ports iterating over build_archs should not see ppc7400 unless they explicitly put ppc7400 into their own `supported_archs`.

If you can point me to an example of a Portfile that you think will need modifying I will be able to understand your concerns better.

kencu:
Reading it over some more, I see what you mean. 

I think it could be made to work -- like arm64 was made to work -- as it is, I'm not sure about things like the muniversal PortGroup and ports that iterate over the folders in there of various archs, or meson with it's universal enhancements I added, etc, etc. None of those consider ppc7400 as an arch to work with. So it could be made to work, but I think pitfalls still seem likely.

It's clever. I continue to push for the practical simplicity of clearing archive_sites on G3s for the few ports that build differently with altivec, as -- to be honest -- there is likely not much enthusiasm for taking on such a project to support G3s when a simpler alternative is readily available.

PPC usage in MacPorts as a whole is a very tiny portion of the user base -- you may well be the only active G3 user we have. There is only so much effort to be expected to support that base.

I am just now in the process of switching several of my G5s over to Debian 10, which has some future as an OS.

evanmiller:
Yeah I was afraid that the universal stuff may prove the downfall of this approach.

Hardware checks in a Portfile make me wince a little bit – but only a little, and I'm fine with your solution if I/we can sell it to MacPorts. The other discussants didn't seem too keen on it.

The separate architecture outlined here seems "right" to me – but really I just want to be able to `port install libjpeg-turbo` on a G3 on not have it barf on me at run-time. I'm fine with either approach, but mainly just want a clear policy in place so that I know how I should be fixing ports to run on G3s.

kencu:
>  mainly just want a clear policy in place so that I know how I should be fixing ports to run on G3s.

At this moment in time, I would think that you and I could decide right here what the clear policy about that is going to be.

hamishmb:
I have an iBook G4 that I'm willing to use to test stuff for you if it is still needed.

PR #246, opened 2021-10-02
Fix Tcl bug 98af80f133: unsafe buffer lifetime

chrstphrchvz:
An old bug in Tcl was exposed by an optimization in LLVM Clang 13: see https://core.tcl-lang.org/tcl/info/24b9181478 and https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=258392

I have not attempted to observe the issue or verify this patch myself, but would suggest it be applied before this eventually becomes a more widespread issue when MacPorts base is compiled with LLVM Clang 13 and future versions of Xcode Clang derived from it.

### DIFF
--- a/vendor/patches/tcl/fix-unsafe-buffer-lifetime.patch
+++ b/vendor/patches/tcl/fix-unsafe-buffer-lifetime.patch
@@ -1,0 +1,33 @@
+"Fix unsafe buffer lifetime" (Tcl bug exposed by LLVM Clang 13 optimization)
+https://core.tcl-lang.org/tcl/info/98af80f133
+
+Index: generic/tclIO.c
+==================================================================
+--- a/vendor/tcl8.5.19/generic/tclIO.c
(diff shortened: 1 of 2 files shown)


chrstphrchvz:
10.14 CI test failure appears to be unrelated:


```
/opt/local/libexec/macports/bin/tclsh8.5 ./tests/curl.tcl ./Pextlib.dylib
SSL certificate problem: certificate has expired
    while executing
"curl fetch --enable-compression https://www.whatsmyip.org/http-compression-test/ $tempfile"
    (procedure "main" line 29)
    invoked from within
"main $argv"
    (file "./tests/curl.tcl" line 63)
```

mascguy:
The certificate issues are being tracked by issue:

https://trac.macports.org/ticket/63563

chrstphrchvz:
My guess is whatsmyip.org is affected by the Let’s Encrypt expiration on 10.14 and earlier: https://lists.macports.org/pipermail/macports-users/2021-October/050298.html

PR #247, opened 2021-11-22
Allow selecting Xcode clang for OpenMP

chrstphrchvz:
As mentioned on macports-dev most recently in https://lists.macports.org/pipermail/macports-dev/2021-November/043902.html. There is a good chance that ports currently using OpenMP will not build with Xcode clang without specifying `-Xpreprocessor -fopenmp` or other additional flags; unless there is a good way to add OpenMP flags to ports automatically, this probably something to mention in https://trac.macports.org/wiki/CompilerSelection#Parallelism, since the MacPorts guide currently doesn’t discuss OpenMP support.

Pinging @eborisch @kencu in case of interest


src/port1.0/portconfigure.tcl
```diff
@@ -1552,6 +1554,15 @@ proc portconfigure::add_automatic_compiler_dependencies {} {
...
         portconfigure::add_compiler_port_dependencies [portconfigure::configure_get_fortran_compiler]
     }
+
+    # Clang (LLVM or Xcode) requires OpenMP runtime from libomp
+    # (GCC will use OpenMP runtime from libgcc)
+    if {[option compiler.openmp_version] ne "" && [string match *clang* ${configure.compiler}]} {
+        ui_debug "Adding depends_lib port:libomp"
+        depends_lib-delete "port:libomp"
+        depends_lib-append "port:libomp"
+    }
```
chrstphrchvz, line 1564: The automatic libomp dependency currently only applies to MacPorts clang, so I have tried relocating it to also apply to Xcode clang, but I am not certain I have done so correctly.

src/port1.0/portconfigure.tcl
```diff
@@ -857,12 +857,12 @@ proc portconfigure::max_version {verA verB} {
...
-#|      3.1       |   3.8   |    Future?    | Future? |   4.7   |
-#|      4.0       | Partial |    Future?    | Future? |   4.9   |
-#|      4.5       | Partial |    Future?    | Future? |   ???   |
+#|--------------------------------------------------------------|
+#|      2.5       |   3.8   |   703.0.29    |   7.3   |   4.2   |
+#|      3.0       |   3.8   |   703.0.29    |   7.3   |   4.4   |
+#|      3.1       |   3.8   |   703.0.29    |   7.3   |   4.7   |
+#|      4.0       |   6.0   | 1000.11.45.2  |  10.0   |   4.9   |
+#|      4.5       |   6.0   | 1000.11.45.2  |  10.0   |    8    |
 #----------------------------------------------------------------
```
chrstphrchvz, line 866: The Xcode clang versions are inferred from the corresponding upstream LLVM versions as listed at https://en.wikipedia.org/wiki/Xcode#Toolchain_versions. I may want to ask @mascguy or someone else with access to older macOS/Xcode to confirm these cutoffs (which should only require compiling a sample OpenMP program).

Also, I notice that Clang 6 is allowed for OpenMP > 3.1; was doing so provisional? Only Clang 7 and later claim to fully support OpenMP 4.5, so would it be better to raise the requirement to Clang 7 (Xcode 10.2)?

I presume this table should be kept synchronized with https://trac.macports.org/wiki/CompilerSelection#Parallelism

kencu:
This is a complex issue to get right that I don't plan to take on myself. The main issue we have noted in the past is that sometimes libraries built with OpenMP enabled have OpenMP flags embedded in their pkgconfig files, and we had no mechanism to tell applications using those libraries that they needed to use OpenMP-capable compilers and pass suitable flags,etc, so the builds would fail.

In general, I have only a high-level understanding of OpenMP and parallelism in general, so I do not claim to have any expertise here beyond recognizing build failures when the flags are wrong. I defer to to @eborisch and will leave it all to him and others to sort out how they wish to do so.

I do know that Apple's preferred parallelism approach for many years now is not OpenMP but Grand Central Dispatch, and they are heavily invested in that parallelism technology instead, and perhaps for good reason:

http://www.idryman.org/blog/2012/08/05/grand-central-dispatch-vs-openmp/

So the reason Xcode doesn't support it (easily) is probably because Apple does not really want you to use it. 

I think.

kencu (commented):
good work!

jasonliu--:
> I do know that Apple's preferred parallelism approach for many years now is not OpenMP but Grand Central Dispatch, and they are heavily invested in that parallelism technology instead, and perhaps for good reason:
> 
> http://www.idryman.org/blog/2012/08/05/grand-central-dispatch-vs-openmp/
> 
> So the reason Xcode doesn't support it (easily) is probably because Apple does not really want you to use it.
> 
> I think.

This behavior would be 100% consistent with how Apple has dealt with other topics that might have implications regarding interoperability. Some other examples include:

* Instead of adopting OpenXR for augmented/virtual reality, invent ARKit.
* Instead of adopting micro-USB, invent the Lightning connector.
* Instead of adopting Vulkan, invent Metal.

eborisch:
I'm going to call the current design option A: clang-mp-X + libomp, and this proposal option B: Xcode's clang + libomp.

I'm not wedded (as I allude to in https://github.com/macports/macports-ports/pull/12933) to B for a few of reasons:

 * It's not officially supported.
 * I don't know which OS/Xcode versions it does/doesn't work on; would need investigation, and may stop working in the future without warning.
 * I'm concerned that configure scripts trying to test for OpenMP will get confused; if they add a bare `-fopenmp` rather than first test to see if OpenMP is working, it all falls apart.

But one things that **is nice** about **B** is it enables -- at least at the application level, see @kencu's caveats about libraries above -- being able to compile OpenMP applications without requiring an *additional* (on top of Xcode) compiler infrastructure. (Not to mention the never-ending clang-9, clang-10, clang-11, ....) That said, even with option **A**, if the package in question is distributable, all that complexity rests on the buildbots, and only the (small) libomp port will need to be installed.

At a high level, for OpenMP to work: (I'm sure I'm missing some things)
 1. The compiler needs to understand the `#pragma omp ...` directives and emit code with the appropriate segments and calls into OpenMP runtime functions
 2. The runtime library needs to be available at link/run time, which handles the actual getting tasks/threads expressed and scheduled in accordance with the OS's threading interface. (`-lomp`)
 3. `omp.h` is required to be available if any user-facing OpenMP functions are called (omp_get_max_threads(), for example)

The libomp port provides the library/headers for _2_  and  _3_ for either **A** or **B**; passing `-Xpreprocessor -fopenmp -I${prefix}/include/libomp` in CPPFLAGS appears (**for some versions of Xcode**) to get Apple's clang to perform _1_ and satisfies _3_ (for B; for A just `-fopenmp`), while adding `-L${prefix}/lib/libomp -lomp` to LDFLAGS competes _2_ (for **B**; for **A** again just `-fopenmp`).

You can see how we similarly patch MacPorts'-provided clang to achieve this automatically (baking in port:libomp's path when passed `-fopenmp`) here: https://github.com/macports/macports-ports/blob/master/lang/llvm-13/files/0001-MacPorts-only-clang-use-MP-omp-locations.patch

I'm at the "this is somewhat interesting" stage; deciding if we (as a group) are interested (how much actual value is there in avoiding a llvm/clang-X install?) is likely the next step, followed by mapping out what versions of Xcode it actually works on.

chrstphrchvz:
I notice that GCC+libgomp (from libgcc) appears to already be supported by base. However many ports using OpenMP already have libomp dependency and paths hardcoded in them, and so assume MacPorts clang will be used without specifically requesting it.

I think this indicates the need for base to provide some abstraction (variables? commands?) with the include directories and other preprocessor flags, linker flags, etc. which ports/portgroups then use as needed, based on the compiler selected by the port. Anything similar in base that could be adapted for this purpose?

chrstphrchvz:
I do not expect to make any progress on this anytime soon.

PR #249, opened 2021-12-04
Do not pass "-arch ppc" to GCC 4.7 - 6

evanmiller:
`-arch ppc` is not a valid option to pass prior to GCC 7. As a consequence, configuring any port using GCC 4.7 - 6 on PPC results in a failed build. Update the `arch_flag_supported` logic to reflect this reality and split off `multiple_arch_flags_supported` (used in `universal_setup`) into a separate function for clarity.

See: https://trac.macports.org/ticket/63723


mascguy:
For some reason, GitHub thinks you're a first-time contributor, and required approval for the CI jobs.

Just took care of that.

mascguy:
> For some reason, GitHub thinks you're a first-time contributor, and required approval for the CI jobs.
> 
> Just took care of that.

Actually this makes sense, as it's your first contribution to macports-base.

evanmiller:
Any updates on this? Would be great to get it reviewed / merged.

ryandesign:
I haven't tried this but I am confused about the PowerPC-specific parts of the change. Are you saying there are compilers that accept Intel -arch flags but not PowerPC -arch flags?

evanmiller:
> I haven't tried this but I am confused about the PowerPC-specific parts of the change. Are you saying there are compilers that accept Intel -arch flags but not PowerPC -arch flags?

Correct. GCC 4.7 - 6 accepts `-arch i386` and `-arch x86_64` but not `-arch ppc`.

mascguy:
Rebased against master

hamishmb:
Good idea, hopefully this will get some more ports working on old PowerPC macs :)

evanmiller:
> Good idea, hopefully this will get some more ports working on old PowerPC macs :)

Glad to have another PPC soldier in the good fight :-)

hamishmb:
Yeah, it's just something completely different and interesting.

I need to fix GRUB for Void Linux too as it pretty much combusted when I upgraded the disk drive, but yeah, always nice to see other people interested in these quirky machines.

ryandesign:
> GCC 4.7 - 6 accepts `-arch i386` and `-arch x86_64` but not `-arch ppc`.

So you're saying that on a PowerPC Mac I can use `gcc-mp-4.7` to compile for PowerPC and `gcc-mp-4.7 -arch i386` to compile for Intel, but on an Intel Mac I can use `gcc-mp-4.7` or `gcc-mp-4.7 -arch i386` to compile for Intel and there is no way to compile for PowerPC? I'm just very confused that the developers of GCC would have done this very strange thing. Or maybe that should not surprise me.

kencu:
no gcc version can compile for a different arch. 

gcc’s idea of cross compiling is to call a whole different compiler to do it, eg on Intel, you build a specific powerpc-apple-darwin compiler (Intel code) that builds powerpc code. You cannot build powerpc code with the x86_64-apple-darwin compiler. 

Apple’s driver was simply a front end to multiple installed gccs that made this more seamless. They last made it available for gcc-4.2. It supported multiple arch flags and lipo’d the results together like clang does now.

a given gcc can compile 32 and 64 bit for the same processor though, if so configured during the build. It uses -m32 and -m64 to drive this. The -arch xyz flags given to gcc are just translated to -mxy flags during parsing.

Support for the flags -arch ppc and -arch ppc64 were added by Iain later; the flags -arch i386 and -arch x86_64 were supported earlier indeed, added by someone else. That’s just how it worked out.

Cross-arch flags are just ignored. The gcc x86_64-apple-darwin compiler at present will just ignore the -arch ppc and -arch ppc64 flags, for example.

Updating Apple’s driver to front end an installation of multiple gccs to again make it work like clang and gcc-4.2 work would be nice. People have talked about doibg that work, but nobody has done it to the point of upstreaming it to gcc.

Changing gcc to have multiple back ends like llvm has is unlikely to happen, but you never know.

hamishmb:
@evanmiller Any chance you know your way around GRUB and OpenFirmware a bit more than I do? This is off topic, but I have a GRUB issue with VOID Linux PPC that I worked around, details here: https://github.com/void-ppc/void-ppc-docs/issues/15

If you could shed any light on why this might be happening it'd be super useful - I'd like to fix the bug(s) rather than stick with my hacky workaround forever.

PR #253, opened 2022-01-28
pextlib1.0: Fix mtime restore in lazy_copy()

neverpanic:
On newer versions of macOS, Apple's implementation of `copyfile(3)` no longer restores the file's modification time when copying a SIP-protected file with `COPYFILE_STAT` (or `COPYFILE_ALL`, which should include `COPYFILE_STAT`). I have not tested whether this is a general problem with `copyfile(3)`, or just a problem when copying SIP-protected files.

Because `lazy_copy()` uses the destination file's mtime to determine whether a new copy of the file is needed, this meant that recently every single invocation of `lazy_copy()` did in fact make a new copy of the file. This caused trace mode to jump to much higher overhead between 6 and 9 times of the normal runtime.

Switch `copyfile(3)` to explicitly not copying the file's stat information and invoke `fchmod(2)` and `futimes(2)` after the copy has been made to ensure that the file modification times are copied as expected.

### DIFF
--- a/src/pextlib1.0/sip_copy_proc.c
+++ b/src/pextlib1.0/sip_copy_proc.c
@@ -17,7 +17,7 @@
  * 3. Neither the name of The MacPorts Project nor the names of its contributors
  *    may be used to endorse or promote products derived from this software
  *    without specific prior written permission.
- * 
+ *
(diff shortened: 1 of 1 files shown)


neverpanic:
See #254 for the build failure.

PR #257, opened 2022-02-13
filemap: replace duplicate code with function call

l2dy:


neverpanic (approved):
Wow that some old code.

PR #258, opened 2022-02-14
cregistry: add build instruction to README.sqlext

l2dy:
See also commit 4664a4a6c680ec55387f7c3ecd94102588a8f299.


neverpanic (approved):

neverpanic:
We also have a port for this, I think.

l2dy:
For the record, the port is `macports.sqlext`.

PR #260, opened 2022-02-15
cregistry: replace nested statements with a loop

l2dy:
with the necessary evil of `goto`

### DIFF
--- a/src/cregistry/entry.c
+++ b/src/cregistry/entry.c
@@ -257,72 +257,52 @@ int reg_entry_delete(reg_entry* entry, reg_error* errPtr) {
     errPtr->free = NULL;
 
     sqlite3_stmt* ports = NULL;
-    sqlite3_stmt* files = NULL;
-    sqlite3_stmt* dependencies = NULL;
-    sqlite3_stmt* portgroups = NULL;
+    sqlite3_stmt* followups[] = { NULL, NULL, NULL };
+    sqlite3_stmt** pFiles = &followups[0];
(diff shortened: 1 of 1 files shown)


src/cregistry/entry.c
```diff
@@ -257,72 +257,53 @@ int reg_entry_delete(reg_entry* entry, reg_error* errPtr) {
...
-                                    break;
-                                case SQLITE_BUSY:
-                                    break;
-                                case SQLITE_ERROR:
-                                default:
-                                    reg_sqlite_error(reg->db, errPtr, NULL);
-                                    break;
-                            }
-                        } while (r == SQLITE_BUSY);
+                        size_t i;
```
neverpanic, line 286: I believe we're using C99, so you can move the declaration into the initialization statement of the for loop.

src/cregistry/entry.c
```diff
@@ -344,14 +325,14 @@ int reg_entry_delete(reg_entry* entry, reg_error* errPtr) {
...
     }
-    if (dependencies) {
-        sqlite3_finalize(dependencies);
+    if (*pDependencies) {
+        sqlite3_finalize(*pDependencies);
     }
-    if (portgroups) {
-        sqlite3_finalize(portgroups);
+    if (*pPortgroups) {
+        sqlite3_finalize(*pPortgroups);
```
neverpanic, line 335: I'd replace those with a loop, too – less chance for missing the free if we ever wanted to add an additional query.

PR #265, opened 2022-02-20
port1.0: set a timeout for ping time checks

l2dy:
Ping time checks should not take longer than 10 seconds without giving users an indication of progress.

### DIFF
--- a/src/port1.0/fetch_common.tcl
+++ b/src/port1.0/fetch_common.tcl
@@ -268,6 +268,7 @@ proc portfetch::sortsites {urls default_listvar} {
         }
 
         set max_hosts_to_ping 50
+        set ping_phase_timeout 10
         set len [llength $hosts]
         if {$len > $max_hosts_to_ping} {
             # randomize them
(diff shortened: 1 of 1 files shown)


jmroot:
I'm not sure how this helps. Does it ever take 10 seconds to start all the pings? Note that you're not timing how long they take to finish, and there is a 3-second timeout specified in the ping command.

l2dy:
> I'm not sure how this helps. Does it ever take 10 seconds to start all the pings? Note that you're not timing how long they take to finish, and there is a 3-second timeout specified in the ping command.

Thanks for the review! You are right that starting the pings would not take that long so this change is not effective, so I'm closing this PR for now.

I am experiencing slow start-up of `port` with no output at all if the system DNS server is unresponsive, and in this situation `Ctrl+C` cannot interrupt the command. I though it could be the pings, but now that I've read code in ping.c, it probably did not block here for too long. I'll continue to investigate this problem.

jmroot:
I've also seen very slow startup sometimes lately, but according to the Terminal title bar it's while `xcodebuild` is being run. Not sure if that's related to what you're seeing, but worth mentioning.

l2dy:
The `xcodebuild` problem is more general and sometimes even blocks zsh from starting up in time, but what I'm seeing here is a different problem, for which I have identified the cause and created a ticket for it. https://trac.macports.org/ticket/64702

PR #268, opened 2022-03-21
Set correct `os_arch` on additional archs

harens:
Whilst running MacPorts on a Raspberry Pi with an armv7l processor, I noticed that MacPorts set the incorrect `os_arch`.

This PR aims to improve MacPorts support on some more architectures by setting the correct `os_arch`. For a given architecture, the new functionality is:

- `Power Macintosh` and anything that begins with `ppc` is set as `powerpc`.
- ` i386`-`i786` and `x86_64` is mapped to `i386`.
- Any architecture that begins with `arm` or `aarch` will be mapped to `arm`.

Closes: https://trac.macports.org/ticket/64863

### DIFF
--- a/src/macports1.0/macports.tcl
+++ b/src/macports1.0/macports.tcl
@@ -699,12 +699,19 @@ proc mportinit {{up_ui_options {}} {up_options {}} {up_variations {}}} {
     # set up platform info variables
     set os_arch $tcl_platform(machine)
     # Set os_arch to match `uname -p`
-    if {$os_arch eq "Power Macintosh"} {
-        set os_arch "powerpc"
-    } elseif {$os_arch eq "i586" || $os_arch eq "i686" || $os_arch eq "x86_64"} {
-        set os_arch "i386"
-    } elseif {$os_arch eq "arm64"} {
-        set os_arch "arm"
+    switch -glob $os_arch {
+       "Power Macintosh" -
+       ppc* {
+           set os_arch powerpc
+       }
+       i[3-7]86 -
+       x86_64 {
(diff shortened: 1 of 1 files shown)


src/macports1.0/macports.tcl
```diff
@@ -699,12 +699,16 @@ proc mportinit {{up_ui_options {}} {up_options {}} {up_variations {}}} {
     # set up platform info variables
     set os_arch $tcl_platform(machine)
     # Set os_arch to match `uname -p`
-    if {$os_arch eq "Power Macintosh"} {
-        set os_arch "powerpc"
-    } elseif {$os_arch eq "i586" || $os_arch eq "i686" || $os_arch eq "x86_64"} {
-        set os_arch "i386"
-    } elseif {$os_arch eq "arm64"} {
-        set os_arch "arm"
+    switch -regexp -- $os_arch {
```
jmroot, line 702: There's really no reason to involve regexes here, `-glob` and some fallthrough cases would be preferable.

harens: Thanks @jmroot. Done.

jmroot:
The `os_arch` part looks fine, but I still don't understand why you would even want to set `build_arch` on non-Darwin platforms. There's no ability to build dependencies universal, so building things for anything other than the toolchain's default arch is very unlikely to succeed. Ignoring that for the moment, it looks like the code will give incorrect results on 64-bit PowerPC and 32-bit ARM.

jmroot:
Well, I guess if it's broken, it's only broken on non-Darwin.

PR #269, opened 2022-03-25
Set `configure.cxx_stdlib` on Linux

harens:
See: https://trac.macports.org/ticket/48957
See: https://github.com/macports/macports-ports/pull/14323#discussion_r835618657

This also means one less thing is required in @kencu's [MacPorts Ubuntu guide](https://trac.macports.org/wiki/InstallingMacPortsOnUbuntuLinux):

```
cxx_stdlib             libstdc++        # configure.cxx_stdlib is not yet working on Ubuntu, this fills in appropriate default
```

`libstdc++` was chosen over`libc++` since it is [more complete on Linux](https://stackoverflow.com/a/16788372/10763533).

### DIFF
--- a/src/port/portindex.tcl
+++ b/src/port/portindex.tcl
@@ -228,12 +228,16 @@ for {set i 0} {$i < $argc} {incr i} {
                         } else {
                             set cxx_stdlib libc++
                         }
+                    } elseif {$os_platform eq "linux"} {
+                        set cxx_stdlib libstdc++
                     }
                     set os_arch [lindex $platlist 2]
                 }
                 if {$os_platform eq "macosx"} {
                     lappend port_options os.subplatform $os_platform os.universal_supported yes cxx_stdlib $cxx_stdlib
                     set os_platform darwin
+                } elseif {$os_platform eq "linux"} {
+                    lappend cxx_stdlib $cxx_stdlib
                 }
                 lappend port_options os.platform $os_platform os.major $os_major os.arch $os_arch
             } elseif {$arg eq "-f"} { # Completely rebuild index


src/port/portindex.tcl
```diff
@@ -228,12 +228,16 @@ for {set i 0} {$i < $argc} {incr i} {
...
+                    } elseif {$os_platform eq "linux"} {
+                        set cxx_stdlib libstdc++
                     }
                     set os_arch [lindex $platlist 2]
                 }
                 if {$os_platform eq "macosx"} {
                     lappend port_options os.subplatform $os_platform os.universal_supported yes cxx_stdlib $cxx_stdlib
                     set os_platform darwin
+                } elseif {$os_platform eq "linux"} {
+                    lappend cxx_stdlib $cxx_stdlib
```
ryandesign, line 240: Why only on Linux do you need to append `$cxx_stdlib` to itself here?

jmroot:
It was intentional to default to not using a `-stdlib` option at all on non-darwin platforms, because it's only on macOS that being able to use a stdlib other than the system default is important.

ryandesign:
Ok, then I guess batch-replace that to add a `${os.platform} eq "darwin"` to the conditional in all the ports as originally proposed in https://github.com/macports/macports-ports/pull/14323.

harens:
> It was intentional to default to not using a -stdlib option at all on non-darwin platforms, because it's only on macOS that being able to use a stdlib other than the system default is important.

Shouldn't a default `stdlib` still be set on Linux? Since it's always going to be `libstdc++`, wouldn't it be easier to set that here?

Otherwise, I'm happy to try batch replacing all the statements. 👍

jmroot:
I'm not quite sure what Ryan is getting at, but if there are ports that don't handle an empty cxx_stdlib correctly, they should probably be fixed regardless.

ryandesign:
What I'm getting it is what was mentioned in https://github.com/macports/macports-ports/pull/14323: that blocks like
```
if {[string match *clang* ${configure.cxx}]} {
     configure.env-append    CXXLIBS=-stdlib=${configure.cxx_stdlib}
}
```
(or at least blocks with that condition) appear in many ports, so if we're not going to fix base so that is guarantees that `${configure.cxx_stdlib}` is set to the right value, then all those ports will need to be changed so those changes happen only on darwin, or perhaps as you suggest only if `${configure.cxx_stdlib}` is nonempty.

kencu:
But -- why not make sure that ```${configure.cxx_stdlib}``` is set to the right value instead of adding workarounds to every usage of this in a Portfile? 

On all platforms, clang has the ability to use either libstdc++ or libc++, and for all we know, someone might want to use libc++ somewhere other than Darwin (like FushiaOS, which I believe does default to libc++ https://fuchsia.dev/fuchsia-src/development/build/toolchain).

Just my $0.02.

jmroot:
Empty is a valid value. Portfiles need to be able to handle it either way.

> On all platforms, clang has the ability to use either libstdc++ or libc++, and for all we know, someone might want to use libc++ somewhere other than Darwin (like FushiaOS, which I believe does default to libc++)

Using no `-stdlib` flag and going with the default still works fine in that case. If someone in the increasingly hypothetical and untenable position of wanting MacPorts to work on some non-Darwin platform *and* use a different C++ stdlib than that platform defaults to, they can set it in macports.conf.

jmroot:
> instead of adding workarounds to every usage of this in a Portfile?

Are there many ports looking at this variable? If so, a better question would be why they need to do that. What ncurses does with it is already a workaround and not at all intended to be something that Portfiles routinely have to do.

ryandesign:
> Are there many ports looking at this variable?

Quite a few.

> If so, a better question would be why they need to do that.

It's fairly common for Portfiles to say:
```
if {[string match *clang* ${configure.cxx}]} {
    configure.ldflags-append -stdlib=${configure.cxx_stdlib}
}
```
This is needed in ports that have C++ code and use $(CXX) as the linker but do not use $(CXXFLAGS) when linking. The alternative would be to patch the Makefiles to use $(CXXFLAGS) when linking but that seems like more work.

jmroot:
Ports that do that must be using the C++ compiler to link, presumably. Does this simpler approach work?
```
configure.ldflags-append {*}${configure.cxxflags}
```

ryandesign:
Yes I think so.

harens:
I've just patched 70 instances in https://github.com/macports/macports-ports/commit/bf574293e5d62f9007c7201acdda21fc50e318e2 by checking whether stdlib is empty (based on https://github.com/macports/macports-ports/commit/ee5188c0921e48a93c1f2ae10a206941fcc8f4fd). 

Given that fixing the issue within the ports is the way forward, I've closed this PR. Thanks all for your help.

PR #270, opened 2022-03-26
Fix sysctl implementation on non-Darwin

harens:
I'm pretty sure that I shouldn't use exec here, so any suggestions are appreciated :)

`sysctl` works slightly different on Linux, and so [`nproc`](https://man7.org/linux/man-pages/man1/nproc.1.html) was chosen to determine the number of processing units. Since `sysctl` might work on some BSDs as well as Darwin, it was left in a default `else` statment.

```
> /sbin/sysctl -n hw.activecpu 
sysctl: cannot stat /proc/sys/hw/activecpu: No such file or directory
```

```
> nproc --all
4
```

### DIFF
--- a/src/macports1.0/macports.tcl
+++ b/src/macports1.0/macports.tcl
@@ -1148,7 +1148,12 @@ match macports.conf.default."
                 }
             }
         } else {
-            set macports::build_arch {}
+            # List of currently supported CPU architectures
+            if {$tcl_platform(machine) in [list arm64 x86_64 i386 ppc]} {
+                set macports::build_arch $tcl_platform(machine)
+            } else {
+                set macports::build_arch {}
+            }
         }
     } else {
         set macports::build_arch [lindex $macports::build_arch 0]
--- a/src/macports1.0/sysctl.c
+++ b/src/macports1.0/sysctl.c
@@ -39,8 +39,11 @@
 #include <errno.h>
 #include <sys/types.h>
 
-#if HAVE_SYS_SYSCTL_H
+/* sys/sysctl.h is deprecated on Linux and behaves unexpectedly */
+#if HAVE_SYS_SYSCTL_H && !defined __linux__
 #include <sys/sysctl.h>
+#else
+#include <unistd.h>
 #endif
 
 #include "sysctl.h"
@@ -48,13 +51,9 @@
 /*
  * Read-only wrapper for sysctlbyname(3). Only works for values of type CTLTYPE_INT and CTLTYPE_QUAD.
  */
-#ifdef HAVE_SYSCTLBYNAME
 int SysctlCmd(ClientData clientData UNUSED, Tcl_Interp *interp, int objc, Tcl_Obj *CONST objv[])
-#else
-int SysctlCmd(ClientData clientData UNUSED, Tcl_Interp *interp, int objc UNUSED, Tcl_Obj *CONST objv[] UNUSED)
-#endif
 {
-#ifdef HAVE_SYSCTLBYNAME
+#if defined HAVE_SYSCTLBYNAME && !defined __linux__
     const char error_message[] = "sysctl failed: ";
     Tcl_Obj *tcl_result;
     int res;
@@ -92,7 +91,37 @@ int SysctlCmd(ClientData clientData UNUSED, Tcl_Interp *interp, int objc UNUSED,
     Tcl_SetObjResult(interp, tcl_result);
     return TCL_OK;
 #else
-    Tcl_SetObjResult(interp, Tcl_NewStringObj("sysctl not available", -1));
-    return TCL_ERROR;
+    char *name;
+    if (objc != 2) {
+        Tcl_WrongNumArgs(interp, 1, objv, "name");
+        return TCL_ERROR;
+    }
+
+    name = Tcl_GetString(objv[1]);
+
+    if (strcmp(name, "hw.activecpu") == 0) {
+        int res = sysconf(_SC_NPROCESSORS_ONLN);
+        if (res < 1)
+        {
+            Tcl_SetObjResult(interp, Tcl_NewStringObj("sysconf not available", -1));
+            return TCL_ERROR;
+        }
+        Tcl_SetObjResult(interp, Tcl_NewIntObj(res));
+        return TCL_OK;
+    } else if (strcmp(name, "hw.memsize") == 0) {
+        long pages = sysconf(_SC_PHYS_PAGES);
+        long page_size = sysconf(_SC_PAGE_SIZE);
+        unsigned long long res = pages * page_size;
+        if (pages < 0 || page_size < 0)
+        {
+            Tcl_SetObjResult(interp, Tcl_NewStringObj("sysconf not available", -1));
+            return TCL_ERROR;
+        }
+        Tcl_SetObjResult(interp, Tcl_NewIntObj(res));
+        return TCL_OK;
+    } else {
+        Tcl_SetObjResult(interp, Tcl_NewStringObj("sysctl option not defined", -1));
+        return TCL_ERROR;
+    }     
 #endif
 }


ryandesign:
> I'm pretty sure that I shouldn't use exec here, so any suggestions are appreciated :)

I'm not sure how bad using `exec` is here, but if it is considered to be bad, as perhaps it was for `sysctl`, then one suggestion would be that you do for `nproc` what base already does for `sysctl`: Provide a Tcl command of that name that does what you want.

harens:
> Provide a Tcl command of that name that does what you want.

Great idea. I wrote an `nproc` clone based around [`sysconf(_SC_NPROCESSORS_ONLN)`](https://www.man7.org/linux/man-pages/man3/sysconf.3.html), which returns the number of processors currently online.

I've tested the new function on macOS 12 with success, and it [should work](https://stackoverflow.com/a/150971/10763533) on Linux, Solaris, AIX and macOS >=10.4 (Tiger). I therefore replaced the original `sysctl` call for greater compatibility.

n.b. On Darwin, build jobs is set based on the no. of processors and memory size. This PR improves support on Linux/Solaris/etc. by setting build jobs based on the no. of processors, rather than not setting it at all.

As an improvement, a future PR might also take into account memory size, which is currently determined using `sysctl`.

src/macports1.0/macports.tcl
```diff
@@ -1135,7 +1135,7 @@ match macports.conf.default."
                     set macports::build_arch x86_64
                 }
             } elseif {$os_major >= 10} {
-                if {[sysctl hw.cpu64bit_capable] == 1} {
+                if {$tcl_platform(machine) eq "x86_64"} {
```
ryandesign, line 1138: This is not correct on 64-bit Macs that use the 32-bit kernel.
```
$ sw_vers -productVersion; echo 'puts $tcl_platform(machine)' | tclsh; sysctl hw.cpu64bit_capable
10.6.8
i386
hw.cpu64bit_capable: 1
```

harens: Ah good catch. That's pretty cool, but also makes things a bit tricky. Let me have a think about this.

Would you mind running ` getconf LONG_BIT`? Also, what is the processor?

ryandesign: The processor is irrelevant. Apple decides which 64-bit Macs use the 32-bit kernel by default and which use the 64-bit kernel by default, and it can vary by OS version, and the user can [override the default](https://web.archive.org/web/20150323044856/https://support.apple.com/en-us/HT3773).

On this 64-bit 10.6 VM, `getconf LONG_BIT` says `64`, and on a 32-bit 10.6 VM, it says `32`, so that may indeed be a useful replacement for `sysctl hw.cpu64bit_capable`.

(I also checked on a PowerBook G4 (which is 32-bit) and under both 10.4 and 10.5 `getconf LONG_BIT` says `32`. I did not test on a Power Mac G5 (which is 64-bit); presumably it says `64` there. The current code was only running `sysctl hw.cpu64bit_capable` for 10.6 and later anyway so I guess this is irrelevant.)

harens: > On this 64-bit 10.6 VM, getconf LONG_BIT says 64, and on a 32-bit 10.6 VM, it says 32, so that may indeed be a useful replacement for sysctl hw.cpu64bit_capable.

Great. I've updated it to use getconf. Thanks for testing @ryandesign.

kencu:
some of these functions are provided to older darwin versions only via legacysupport

https://github.com/macports/macports-legacy-support/blob/master/src/macports_legacy_sysconf.c

harens:
> some of these functions are provided to older darwin versions only via legacysupport
> 
> https://github.com/macports/macports-legacy-support/blob/master/src/macports_legacy_sysconf.c

Ah ok I didn't realise this. [The place](https://stackoverflow.com/a/150971) where I got the idea of _SC_NPROCESSORS_ONLN from suggested that it would work on Tiger, so thanks for correcting me on that. Either way, _SC_PHYS_PAGES definitely won't work.

I'll try and think of a different approach, which I'll cross-check with legacysupport before pushing. Thanks @kencu.

kencu:
or can maybe just copy them over? Not sure how easy that is.

Having MP run nicely on linux would be great!

harens:
> or can maybe just copy them over? Not sure how easy that is.

I had a go at this. It seemed a bit too interlinked with sys/sysctl.h, which is what we're trying to avoid on Linux. Thank you for the suggestion though.

I realised that this can be massively simplified if we conditionally define the Tcl sysctl command, rather than creating lots of new commands. This then doesn't change any of the functions used on Darwin, whilst also allowing MacPorts to correctly determine the number of CPUs and memory size on Linux.

```tcl
haren@raspberrypi: ~ » port-tclsh                                                                                                                                                [14:24:17]
% package require macports
1.0
% mportinit
Warning: port definitions are more than two weeks old, consider updating them by running 'port selfupdate'.
% sysctl hw.activecpu
4
% sysctl hw.memsize
1020002304
% sysctl dont.exist
sysctl option not defined
```

src/macports1.0/sysctl.c
```diff
@@ -92,7 +91,36 @@ int SysctlCmd(ClientData clientData UNUSED, Tcl_Interp *interp, int objc UNUSED,
...
     return TCL_OK;
 #else
-    Tcl_SetObjResult(interp, Tcl_NewStringObj("sysctl not available", -1));
-    return TCL_ERROR;
+    char *name;
+    name = Tcl_GetString(objv[1]);
+    if (objc != 2) {
+        Tcl_WrongNumArgs(interp, 1, objv, "name");
+        return TCL_ERROR;
+    }
```
neverpanic, line 99: Shouldn't you check `objc` before accessing `objv[1]`?

```suggestion
    char *name;
    if (objc != 2) {
        Tcl_WrongNumArgs(interp, 1, objv, "name");
        return TCL_ERROR;
    }

    name = Tcl_GetString(objv[1]);
```

src/macports1.0/sysctl.c
```diff
@@ -92,7 +91,36 @@ int SysctlCmd(ClientData clientData UNUSED, Tcl_Interp *interp, int objc UNUSED,
...
+            Tcl_SetObjResult(interp, Tcl_NewStringObj("sysconf not available", -1));
+            return TCL_ERROR;
+        }
+        Tcl_SetObjResult(interp, Tcl_NewIntObj(res));
+        return TCL_OK;
+    } else if (strcmp(name, "hw.memsize") == 0) {
+        long pages = sysconf(_SC_PHYS_PAGES);
+        long page_size = sysconf(_SC_PAGE_SIZE);
+        unsigned long long res = pages * page_size;
+        if (res < 1)
```
neverpanic, line 114: Wouldn't this incorrectly return `1` if both calls to `sysconf` fail?

Why not
```suggestion
        if (pages < 0 || page_size < 0)
```
or with 1, although strictly speaking that's not documented as an error in the manpage.

neverpanic (changes requested):
Minor nitpicks, other than that this looks good to go.

harens:
Two very good spots. Thanks @neverpanic.

neverpanic (approved):

src/macports1.0/macports.tcl
```diff
@@ -1148,7 +1148,12 @@ match macports.conf.default."
                 }
             }
         } else {
-            set macports::build_arch {}
+            # List of currently supported CPU architectures
+            if {$tcl_platform(machine) in [list arm64 x86_64 i386 ppc]} {
```
jmroot, line 1152: Are we sure these are the names that will be used on non-Darwin platforms? Tcl gets this value by calling `uname(3)`. I'm pretty sure I remember `i686` being used on Linux sometimes, for example.

src/macports1.0/sysctl.c
```diff
@@ -92,7 +91,37 @@ int SysctlCmd(ClientData clientData UNUSED, Tcl_Interp *interp, int objc UNUSED,
     Tcl_SetObjResult(interp, tcl_result);
     return TCL_OK;
 #else
```
jmroot, line 93: We're assuming that all the other platforms we care about have `sysctlbyname`, then?

src/macports1.0/macports.tcl
harens: Yep this can be improved by pretty much copying https://github.com/macports/macports-base/pull/268. I will extend these checks further in that PR.

src/macports1.0/sysctl.c
harens: Darwin (and maybe BSD?) should have `sysctlbyname`. The `else` case, at least in my testing, works on Linux which was my main focus. It [might](https://stackoverflow.com/a/150971) work on Solaris and AIX, although this hasn't been tested.

(Technically Linux does have sysctl, it's just [deprecated](https://www.phoronix.com/news/Linux-5.5-Kills-SYSCTL-SYSCALL) and behaves weirdly.)

jmroot: It's just that the else case is used on all platforms that don't have `sysctlbyname`. The Tcl code can handle sysctl failing, which is what used to happen on platforms we don't know how to implement sysctl on. Seems like now it would fail to compile in those cases.

harens: I'm not quite sure I understand. If sysctl fails, the else statement falls back to sysconf.  I tested this with an [ubuntu workflow](https://github.com/harens/macports-base/runs/7498881727), where it compiles successfully  (tests fail for an unrelated reason).

If sysconf and sysctl both fail, it should still compile and error successfully, just like before the PR. Rather than `Tcl_NewStringObj("sysctl not available", -1)`, it should produce `Tcl_NewStringObj("sysconf not available", -1)`. Please let me know if I've misunderstood :)

jmroot: I mean if there's a platform that doesn't have `sysctlbyname` and also doesn't have either `sysconf` or the `_SC_*` constants that are being used, the code will fail to compile, where previously it would compile and sysctl would just always fail.

PR #272, opened 2022-05-12
portlint.tcl: Error if version contains + character

ryandesign:
See: https://trac.macports.org/ticket/65073


raimue (approved):
Looks good to me. :+1:

jmroot:
I'm not totally sold on the wording and it being an error, since it doesn't actually break anything, it just makes it inconvenient to specify the version. I certainly agree that lint should say something about it though.

PR #274, opened 2022-06-29
Set SOURCE_DATE_EPOCH

harens:
Closes: https://trac.macports.org/ticket/59672

This PR allows base to define the [`SOURCE_DATE_EPOCH`](https://reproducible-builds.org/docs/source-date-epoch/) environment variable to assist with reproducible builds. The value can be manually set in a portfile via the `source_date_epoch` variable.

If the variable isn't assigned, its value is set to the greatest Unix timestamp of the files present in the source code (credit to Ryan for [implementation details](https://trac.macports.org/ticket/59672#comment:3)).

### DIFF
--- a/src/port1.0/portextract.tcl
+++ b/src/port1.0/portextract.tcl
@@ -165,6 +165,7 @@ proc portextract::extract_main {args} {
         if {[llength $worksubdirs] == 1} {
             set origpath [lindex $worksubdirs 0]
             set newpath [file join $workpath $distname]
+            set_source_date_epoch [lindex $worksubdirs 0]
             if {$newpath ne $origpath} {
                 ui_debug [format [msgcat::mc "extract.rename: Renaming %s -> %s"] [file tail $origpath] $distname]
                 move $origpath $newpath
@@ -174,6 +175,9 @@ proc portextract::extract_main {args} {
         } else {
             return -code error "extract.rename: multiple directories exist in ${workpath}: $worksubdirs"
         }
+    } else {
+        global worksrcpath
+        set_source_date_epoch $worksrcpath
     }
 
     return 0
--- a/src/port1.0/portutil.tcl
+++ b/src/port1.0/portutil.tcl
@@ -429,7 +429,7 @@ proc command_exec {args} {
         file mkdir ${dir}
     }
 
-    global ${varprefix}.env_array ${varprefix}.nice env
+    global ${varprefix}.env_array ${varprefix}.nice env target_state_fd
 
     # Set the environment.
     # If the array doesn't exist, we create it with the value
@@ -455,6 +455,9 @@ proc command_exec {args} {
     set ${varprefix}.env_array(DEVELOPER_DIR) [option configure.developer_dir]
     if {[option configure.sdkroot] ne ""} {
         set ${varprefix}.env_array(SDKROOT) [option configure.sdkroot]
+    }
+    if {[get_statefile_value "SOURCE_DATE_EPOCH" $target_state_fd source_date_epoch] != 0} {
+        set ${varprefix}.env_array(SOURCE_DATE_EPOCH) $source_date_epoch
     }
 
     # Debug that.
@@ -3215,6 +3218,24 @@ proc exec_as_uid {uid code} {
     }
 
     return -code $retcode $result
+}
+
+# See https://trac.macports.org/ticket/59672
+proc set_source_date_epoch {dir} {
+    global target_state_fd
+    set source_date_epoch 0
+    if { [file exists $dir] } {
+        fs-traverse i [list ${dir}] {
+            if {[file isfile ${i}]} {
+                set t [file mtime ${i}]
+                if {${t} > ${source_date_epoch}} {
+                    set source_date_epoch ${t}
+                }
+            }
+        }
+        ui_debug "Setting SOURCE_DATE_EPOCH to ${source_date_epoch}"
+        write_statefile "SOURCE_DATE_EPOCH" ${source_date_epoch} $target_state_fd
+    }
 }
 
 # dependency analysis helpers


harens:
Just to show how powerful `SOURCE_DATE_EPOCH` is, I've pushed a commit that standardises the file modification timestamps produced by MacPorts archives.

I tested this with a port I maintain, [ioping](https://ports.macports.org/port/ioping/). I ran `sudo port -ds archive ioping` twice, and then compared the results.

Before this PR: https://try.diffoscope.org/kbxsdthrrxqf.html - File timestamps are different, so the archives are as well.
Afterwards: https://try.diffoscope.org/cugyrxrbcbus - The archives are completely identical.

I plan to replicate this in the future for other archive/binary package formats produced by MacPorts

src/port1.0/portextract.tcl
```diff
@@ -165,7 +165,10 @@ proc portextract::extract_main {args} {
             set dir [file tail [lindex $workdirs 0]]
             ui_debug [format [msgcat::mc "Symlink: %s -> %s"] $worksrcpath $dir]
             symlink $dir $worksrcpath
+            set_source_date_epoch [lindex $workdirs 0]
         }
+    } else {
+        set_source_date_epoch $worksrcpath
```
neverpanic, line 171: Does this work correctly if the `extract` and `destroot` steps are run separately, i.e. when I do

```
sudo port -v extract
sudo port -v destroot
```

It seems to me this would lead to `source_date_epoch` being left at 0. I guess we should record the value of `source_date_epoch` in the statefile, and read it from there in the other phases that need it.

src/port1.0/portinstall.tcl
```diff
@@ -314,6 +314,14 @@ proc portinstall::create_archive {location archive.type} {
     puts $fd "@cxx_stdlib_overridden ${portinstall::cxx_stdlib_overridden}"
     close $fd
 
+    # Set standardised file modification times
+    fs-traverse file $destpath {
+        # Keep the original times for files that have not been modified during the build
+        if {[file mtime $file] > $source_date_epoch} {
+            file mtime $file $source_date_epoch
+        } 
+    }
```
neverpanic, line 323: I'm not sure that will work for all ports as we expect it to. As far as I recall, Python scripts check that the timestamps of their associated .pyc files is strictly larger than the timestamp of the script itself.

For python scripts generated during the build (e.g., because of some autotools replacement magic), this would mean that we might end up shipping a pre-compiled .pyc file that ends up being ignored by all users.

Maybe we should provide an option to turn this behavior off for ports where it would have adverse effects, or add a method to override the time for certain files?

neverpanic (commented):
Thanks for working on this, this is and important step and a good thing to have. I have a few concerns, though.

src/port1.0/portinstall.tcl
harens: Sorry it's taken me this long to get to this @neverpanic. Appreciate the feedback. 

> As far as I recall, Python scripts check that the timestamps of their associated .pyc files is strictly larger than the timestamp of the script itself.

Just looking it up, it seems like it should be fine for all supported python versions (see [PEP 552](https://peps.python.org/pep-0552/)). Setting SOURCE_DATE_EPOCH triggers using hash-based pyc files (https://github.com/python/cpython/commit/ccbe5818af20f8c12043f5c30c277a74714405e0).

However, I completely agree that there might be other scenarios where this has adverse effects.

> Maybe we should provide an option to turn this behavior off for ports where it would have adverse effects, or add a method to override the time for certain files?

I've added a `reproducible_timestamps` option. It's default is set to true, but this can be changed if that's the general consensus.

harens:
Seem to be a lot of "failed to extract" errors:

```
error: failed to extract dependencies-b: /private/tmp/macports-tests/opt/local/var/macports/build/_users_runner_work_macports-base_macports-base_tests_test_dependencies-b/dependencies-b/work/dependencies-b-1: no such file or directory
...
Error: Failed to extract envvariables: /private/tmp/macports-tests/opt/local/var/macports/build/_Users_runner_work_macports-base_macports-base_tests_test_envvariables/envvariables/work/envvariables-1: no such file or directory
...
etc.
```

Any help debugging appreciated.

src/port1.0/portmain.tcl
```diff
@@ -113,6 +114,7 @@ default destroot {${destpath}}
 default filesdir files
 default revision 0
 default epoch 0
+default reproducible_timestamps 1
```
ryandesign, line 117: Although using `1` probably works, MacPorts convention is to use `yes` and `no` for Booleans.
```suggestion
default reproducible_timestamps yes
```

src/port1.0/portinstall.tcl
```diff
@@ -308,6 +308,19 @@ proc portinstall::create_archive {location archive.type} {
     puts $fd "@cxx_stdlib_overridden ${portinstall::cxx_stdlib_overridden}"
     close $fd
 
+    if {[get_statefile_value "SOURCE_DATE_EPOCH" $target_state_fd source_date_epoch] != 0 && $reproducible_timestamps == 1} {
```
ryandesign, line 311: This will need a corresponding change.

src/port1.0/portinstall.tcl
```diff
@@ -308,6 +308,19 @@ proc portinstall::create_archive {location archive.type} {
...
+        # Set standardised file modification times
+        fs-traverse file $destpath {
+            if {[file type $file] == "link"} {
+                # file mtime doesn't work on symbolic links
+                exec touch -ht $source_date_epoch $file
```
ryandesign, line 316: I would rather see the MacPorts [`touch` procedure](https://github.com/macports/macports-base/blob/82b064968d0ae5561911b850249939a7f77ae158/src/port1.0/portutil.tcl#L1168) be used than using the external `touch` command. The MacPorts `touch` procedure doesn't support `-h` yet but perhaps that could be added.

ryandesign:
Took me a minute of reading the code, but I see now that you implemented:

1. setting `SOURCE_DATE_EPOCH` during all phases, and
2. changing the timestamp of all files in the archive to that value

In [the ticket](https://trac.macports.org/ticket/59672) I was only proposing doing (1) and I don't know what the consequences are of doing (2).

harens:
> Took me a minute of reading the code, but I see now that you implemented:
> 
>     1. setting `SOURCE_DATE_EPOCH` during all phases, and
> 
>     2. changing the timestamp of all files in the archive to that value
> 
> 
> In [the ticket](https://trac.macports.org/ticket/59672) I was only proposing doing (1) and I don't know what the consequences are of doing (2).

Thank you for the review Ryan, you make a very valid point. I think the easiest thing for now is to simply add SOURCE_DATE_EPOCH to base as was discussed in the original ticket. Standardising the timestamps can be discussed at a later date.

PR #283, opened 2022-10-28
Add 2020 as known c++ standard

cjones051073:
Compiler support has I think reached a level it make sense to add as a known standard.

As always, support is a gradual process with compilers. The versions here are a preliminary estimate of when the compilers first support the *primary* c++20 features, which of course is a bit subjective.


jmroot:
This is https://trac.macports.org/ticket/62677 BTW.

cjones051073:
Thanks, I had not noticed that trac ticket...

PR #284, opened 2022-11-13
ci: Add ubuntu runner

harens:
To test that MacPorts builds and installs on non-Darwin, I thought that it would be useful to introduce an Ubuntu GitHub actions runner.

This is especially important since the ports.macports.org machine runs MacPorts on a linux distro.

Some of the tests fail on non-Darwin, so they are disabled for now. The most important part is checking that the build itself succeeds.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [macos-11, macos-12]
+        os: [macos-11, macos-12, ubuntu-latest]
 
     steps:
       - name: Checkout repository
@@ -19,9 +19,16 @@ jobs:
         with:
           fetch-depth: 64
       - name: Cleanup /usr/local
+        if: startsWith(matrix.os,'macos')
         run: |
           sudo mkdir /opt/local.old
           sudo mv /usr/local/* /opt/local.old
+      - name: Install Linux Dependencies
+        if: startsWith(matrix.os,'ubuntu')
+        run: >
+          sudo apt update &&
+          sudo apt install mtree-netbsd
+          libcurl4-openssl-dev
       - name: Configure MacPorts Base
         run: |
           set -eu
@@ -29,12 +36,19 @@ jobs:
       - name: Build MacPorts Base
         run: |
           set -eu
-          make -j$(sysctl -n hw.activecpu)
+          platform=$(uname -s)
+          case ${platform} in
+            Darwin) make -j$(sysctl -n hw.activecpu)
+            ;;
+            Linux) make -j$(nproc)
+            ;;
+          esac
       - name: Install MacPorts Base
         run: |
           set -eu
           sudo make install
       - name: Test MacPorts Base
+        if: startsWith(matrix.os,'macos')
         run: |
           set -eu
           make test


herbygillot:
> This is especially important since the ports.macports.org machine runs MacPorts on a linux distro.

Interesting, I wasn't aware of that.

harens:
> Interesting, I wasn't aware of that.

Yep, we've been discussing it over at https://github.com/macports/macports-ports/pull/16657. ([here's the dockerfile](https://github.com/arjunsalyan/macports-ubuntu/blob/master/Dockerfile))

.github/workflows/main.yml
```diff
@@ -22,19 +22,32 @@ jobs:
         run: |
           sudo mkdir /opt/local.old
           sudo mv /usr/local/* /opt/local.old
```
ryandesign, line 24: Is removing everything from /usr/local going to be correct on a Linux system? We do it here because the macOS runners GitHub Actions provides come pre-installed with Homebrew in /usr/local and we don't want that.

.github/workflows/main.yml
```diff
@@ -22,19 +22,32 @@ jobs:
...
+      - name: Install Linux Dependencies
+        if: startsWith(matrix.os,'ubuntu')
+        run: >
+          sudo apt update &&
+          sudo apt install mtree-netbsd
+          gnustep libcurl4-gnutls-dev
```
ryandesign, line 30: Is it correct for this to be split up across three lines?

ryandesign (commented):
Seems reasonable to have some Linux runner. We used to have an Oracle Linux worker for MacPorts base on our Buildbot system back when it was being administered by macOS forge. Most non-building tasks, including the portindexing, were handled on Oracle Linux VMs then. When I took over hosting our Buildbot in late 2016 I switched to macOS VMs because I was more familiar administering macOS.

.github/workflows/main.yml
harens: Yes you're right, it's not required to remove /usr/local/ on the ubuntu runner. Just having a quick flick through the [pre-installed software](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#package-management), Homebrew is installed. However, it's in `/home/linuxbrew` and not in PATH by default.

I've changed it so that it's only removed on macOS.

.github/workflows/main.yml
harens: This should be fine. Instead of the usual `run: |`, I used `run: >`. This means that newlines are replaced with spaces (based off [here](https://stackoverflow.com/a/65808412)). Reviewing the logs, all the packages seem to be installed.

.github/workflows/main.yml
```diff
@@ -11,30 +11,44 @@ jobs:
...
+        if: startsWith(matrix.os,'macos')
         run: |
           sudo mkdir /opt/local.old
           sudo mv /usr/local/* /opt/local.old
+      - name: Install Linux Dependencies
+        if: startsWith(matrix.os,'ubuntu')
+        run: >
+          sudo apt update &&
+          sudo apt install mtree-netbsd
+          gnustep libcurl4-gnutls-dev
```
jmroot, line 31: The gnustep package isn't needed. Also it might make sense to use libcurl4-openssl-dev, since openssl is already installed in the image and gnutls is not?

harens: Thanks @jmroot. I've made the suggested changes.

jmroot (approved):

.github/workflows/main.yml
```diff
@@ -11,30 +11,44 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [macos-11, macos-12]
+        os: [macos-11, macos-12, ubuntu-latest]
```
DIEGOHJJACOBO, line 14: COPY --from=builder /opt/local /opt/local
COPY --from=builder /MacPorts-2.8.1 /MacPorts-2.8.1

ARG CONF_PATH=/opt/local/etc/macports/macports.conf
# Write some required defaults.
# Unless, macports-base is modified so that it is able to detect these, 
# we need to hardcode them

# 2 because this image is mostly intended to run on VMs
RUN echo "buildmakejobs 2" >> $CONF_PATH

RUN echo "cxx_stdlib libstdc++" >> $CONF_PATH
RUN echo "build_arch x86_64" >> $CONF_PATH

# add symlink for gnumake
RUN ln -s /usr/bin/make /usr/bin/gnumake

ENV PATH="/opt/local/bin:$PATH"

DIEGOHJJACOBO: > 
> COPY --from=builder /opt/local /opt/local
> COPY --from=builder /MacPorts-2.8.1 /MacPorts-2.8.1
> 
> ARG CONF_PATH=/opt/local/etc/macports/macports.conf
> # Write some required defaults.
> # Unless, macports-base is modified so that it is able to detect these, 
> # we need to hardcode them
> 
> # 2 because this image is mostly intended to run on VMs
> RUN echo "buildmakejobs 2" >> $CONF_PATH
> 
> RUN echo "cxx_stdlib libstdc++" >> $CONF_PATH
> RUN echo "build_arch x86_64" >> $CONF_PATH
> 
> # add symlink for gnumake
> RUN ln -s /usr/bin/make /usr/bin/gnumake
> 
> ENV PATH="/opt/local/bin:$PATH"

[192.168.1.1](diego-duy1@hotmail.com)

DIEGOHJJACOBO (commented):
(Diego-duy1@hotmail.com)

PR #285, opened 2022-11-14
Do not pass -stdlib with gcc 10+ in PPC builds

cjones051073:
https://github.com/macports/macports-base/commit/ebd226139d5ba8389367606761658ad0e3bb5cce

Added support for -stdlib with gcc10+, but neglected the fact this is not available on PPC.

See also https://trac.macports.org/ticket/66246

### DIFF
--- a/src/port1.0/portconfigure.tcl
+++ b/src/port1.0/portconfigure.tcl
@@ -80,10 +80,13 @@ proc portconfigure::should_add_stdlib {} {
     # GCC also supports -stdlib starting with GCC 10 (and devel)
     set is_gcc_with_stdlib no
     if { [string match *g*-mp-* [option configure.cxx]] } {
-        # Extract gcc version from value after last -
-        set gcc_ver [lindex [split [option configure.cxx] "-"] end]
-        if { ${gcc_ver} eq "devel" || ${gcc_ver} >= 10 } {
-            set is_gcc_with_stdlib yes
+        # -stdlib is not available with PPC builds
+        if { [option configure.build_arch] ni [list ppc ppc64] } {
+            # Extract gcc version from value after last -
+            set gcc_ver [lindex [split [option configure.cxx] "-"] end]
+            if { ${gcc_ver} eq "devel" || ${gcc_ver} >= 10 } {
+                set is_gcc_with_stdlib yes
+            }
         }
     }
     return [expr {$has_stdlib && ($is_clang || $is_gcc_with_stdlib)}]


kencu:
sometimes, macports sets the stdlib to macports-libstdc++…gcc cannot handle this setting, but macports-clang uses it.

FYI, Iain tells us none of the existing c++ header sets work quite right with gcc/libc++ and he is working on a curated set of headers to use for this.

https://github.com/iains/darwin-toolchains-start-here/discussions/31

cjones051073:
> sometimes, macports sets the stdlib to macports-libstdc++…gcc cannot handle this setting, but macports-clang uses it.

Any port that is doing this, so requesting macports-libstdc++ but then also using macports gcc to build, should probably be considered mis-configured.

jmroot:
> Any port that is doing this, so requesting macports-libstdc++ but then also using macports gcc to build, should probably be considered mis-configured.

But that is exactly the stdlib that ports building with (older) macports-gcc are often using. I would suggest that the code that translates a cxx_stdlib setting into a flag to pass to the compiler needs to handle this case.

cjones051073:
> > Any port that is doing this, so requesting macports-libstdc++ but then also using macports gcc to build, should probably be considered mis-configured.
> 
> But that is exactly the stdlib that ports building with (older) macports-gcc are often using. I would suggest that the code that translates a cxx_stdlib setting into a flag to pass to the compiler needs to handle this case.

I am confused now, as this seems to be in conflict with what Ken says above, that gcc cannot handle this.

in any case the change here fixes things for PPC so in itself is correct. I don’t run any systems using macports-libstdc++ so will have to defer testing this with that to those that are using it.

jmroot:
> I am confused now, as this seems to be in conflict with what Ken says above, that gcc cannot handle this.

GCC versions that don't support the `-stdlib` flag still link to a macports-provided libstdc++.

cjones051073:
> > I am confused now, as this seems to be in conflict with what Ken says above, that gcc cannot handle this.
> 
> GCC versions that don't support the `-stdlib` flag still link to a macports-provided libstdc++.

My changes here then do nothing to impact that.

kencu:
it does get a little hard to follow.

probably the simplest pure fix is to only pass the stdlib to gcc if it is libc++ and gcc >= 10.

everything else is a mess, I believe.

cjones051073:
> it does get a little hard to follow.
> 
> probably the simplest pure fix is to only pass the stdlib to gcc if it is libc++ and gcc >= 10.
> 
> everything else is a mess, I believe.

Thats precisely what is already done.

cjones051073:
> > it does get a little hard to follow.
> > probably the simplest pure fix is to only pass the stdlib to gcc if it is libc++ and gcc >= 10.
> > everything else is a mess, I believe.
> 
> Thats precisely what is already done.

Sorry, correction. the stdlib is always passed. That is perfectly fine for both libc++ and libstdc++. For macports-stdlibc++ I cannot say, but if a port is configuring to use gcc10+ and use this, then thats a weird combination anyway.

cjones051073:
> and even then, Iain says the header situation is not ready for prime time yet, but I haven’t kicked the tires hard yet. I guess we’ll find out soon enough if there are or are not issues…

In my testing most ports using gcc built fine configured to use libc++. But indeed some might have issues, and if they do its trivial to explicitly state in those ports they wish to stick with libstdc++ by just adding the appropriate configuration. Making libc++ the default is I think still the best approach right now.

kencu:
ergo… only pass libc++ to gcc, otherwise, a mess

kencu:
anyway, just trying to be helpful a bit here, but clearly, I’m not being :) 

I’ll leave you to it

kencu:
stdlib=libc++, to macports, means linking to /usr/lib/libstdc++.dylib

cjones051073:
> stdlib=libstdc++, to macports, means linking to /usr/lib/libstdc++.dylib

thats not what I see happening.

kencu:
stdlib=macports-libstdc++ means linking to to the libstdc++.dylib from libgcc

kencu:
you are not seeing the whole picture yet

kencu:
I know… unless you use these systems, it’s hard

cjones051073:
> I know… unless you use these systems, it’s hard

I mean, if you genuinely think the -stdlib flag should only be passed when libc++ is requested, then fine this could be added. I would just like to see the explicit example which shows this is necessary.

kencu:
use any macports clang compiler, build a c++ “hello.cpp”, and pass it the three different stdlibs.

assuming there is a system /usr/lib/libstdc++.dylib, you should see what I mean

cjones051073:
> use any macports clang compiler, build a c++ “hello.cpp”, and pass it the three different stdlibs.
> 
> assuming there is a system /usr/lib/libstdc++.dylib, you should see what I mean

I cannot do that test myself as I do not have a system with /usr/lib/libstdc++.dylib - Can you do it please and post what you see here ?

kencu:
sure. It’s 1:41 am though, do in the morning ok?

cjones051073:
> > sure. It’s 1:41 am though, do in the morning ok?
> 
> its 9.40 am here for me ;) But yes it can obviously wait.

b.t.w. my system does have /usr/lib/libstdc++.6.dylib and building with stdlib = libc++ or libstdc++ does not do what you describe ? See my post above.

kencu:
with macports clang? then something has broken recently

cjones051073:
> with macports clang? then something has broken recently

Not sure what you mean. the commands above are running macports gcc.

I think its best you check this yourself...

cjones051073:
I am on macOS12 b.t.w. in case that matters.

cjones051073:
Hi Ken.

I still do not understand the point you are trying to make. The examples you post above all use `clang++-mp-14` to build. Nothing I have done will have any bearing at all on using any macports clang to build anything. Only if you use macports gcc.

kencu:
Have you given any consideration to setting up a VM for one of these older systems? It does not take more than a few minutes, the OS is free to download, and it just takes up a few GB.

cjones051073:
Also note, my changes have not altered what happens if you request stdlibc++ with gcc. Please try your test above, but use macports gcc12 instead of clang14c ?

cjones051073:
> Also note, my changes have not altered what happens if you request stdlibc++ with gcc. Please try your test above, but use macports gcc12 instead of clang14 ?

cjones051073:
> Have you given any consideration to setting up a VM for one of these older systems? It does not take more than a few minutes, the OS is free to download, and it just takes up a few GB.

I have a VM, just only on a now very old mac mini that is currently having problems. I don't have time to try and fix it right now, nor the space on my MacBook to host VMs there.

kencu:
Well, that was my go at explaining it. Maybe Josh can help better.

cjones051073:
> Macports base will change `-stdlib=libstdc++` to `-stdlib=macports-libstdc++` if compiler.cxx_standard is >= 2011.
> 
> passing `-stdlib=macports-libstdc++` to gcc is never going to work. So that will break.
> 
> SO we should not do that.
> 
> Only pass `-stdlib=libc++` if needed, and all will be well.

can you please run two tests yourself, suing gcc12

1. with -stdlib=libstdc++ 
2. without -stdlib=libstdc++ 

and post what you currently get ?

kencu:
they will work, but that is not our issue.

and gcc12 will not link to /usr/lib/libstdc++.6.dylib with those settings, which is what macports expects it to do.

kencu:
what gcc does with the stdlib setting has to be the same as what clang does with the stdlib setting. At present, the only one that works right with gcc is -stdlib=libc++.

kencu:
And if YOU have so much trouble getting this, I can only imagine how much trouble Joe Blow out there is having !!!!!

cjones051073:
> Well, that was my go at explaining it. Maybe Josh can help better.

I'll happily make the change, I can see at least avoiding macports-libstdc++ is a good idea. Its just I have yet to see what you describe as using -stdlib=libstdc++ being a problem. As far as I can see that option is exactly equivalent to not passing any stdlib at all to gcc.

cjones051073:
> they will work, but that is not our issue.
> 
> and gcc12 will not link to /usr/lib/libstdc++.6.dylib with those settings, which is what macports expects it to do.

But, nor does it do if if you don't bass any -stdlib at all. My point is either passing -stdlib=stdlibc++ or not passing anything at all gives exactly the same results.

kencu:
Just want to say, thank you Chris for the time and talent you spend on this, and MacPorts in general. I am sorry my explaining skills don't seem to be up to the task, but all your efforts are much appreciated, and I know we're just trying to get it right.

Esp for all these folks who seem to just love these older systems and so flock to MacPorts for it!

cjones051073:
OK Ken, if you don't mind try this.

```
Oberon ~/cernbox/MacPorts/c++17 > g++-mp-12 -O3 -std=c++17 -o test1 ./test.cpp
Oberon ~/cernbox/MacPorts/c++17 > g++-mp-12 -O3 -std=c++17 -stdlib=libstdc++ -o test2 ./test.cpp
Oberon ~/cernbox/MacPorts/c++17 > otool -L test1 test2
test1:
	@rpath/libstdc++.6.dylib (compatibility version 7.0.0, current version 7.30.0)
	@rpath/libgcc_s.1.1.dylib (compatibility version 1.0.0, current version 1.1.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1319.0.0)
test2:
	@rpath/libstdc++.6.dylib (compatibility version 7.0.0, current version 7.30.0)
	@rpath/libgcc_s.1.1.dylib (compatibility version 1.0.0, current version 1.1.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1319.0.0)
```

Do you agree the result of the two compilations above is exactly the same ?

If so, where is the flaw in my logic that *if* not passing stdlib to gcc is OK, then passing stdlib=libstdc++ has to also be OK ?

cjones051073:
> right here is the crux of the issue, I believe:
> 
> https://github.com/macports/macports-base/blob/44ce17dd0dfa7b7d4ef347fd05d1f09f3f379c41/src/port1.0/portconfigure.tcl#L132
> 
> macports changes -stdlib=libstdc++ to -stdlib=macports-libstdc++ to indicate that only the newer libgcc runtime will work.
> 
> macports-clang just handles that right, as above. For gcc, a correct gcc that defaults to the new libstdc++ is chosen.
> 
> but you can't pass -stdlib=macports-libstdc++ to gcc, it won't understand it.

Right. That explains why passing macports-libstdc++ should not be done. I've already accepted that, and will veto it such that nothing is passed, which means gcc will fallback to its default with is to use its own libstdc++, which is what is wanted.

What I still do not see, and what I have been discussing, is any reason to veto passing -stdlib=libstdc++ to gcc, when base has *not* changed it to macports-libstdc++.

I will post a PR that vetos passing macports-libstdc++

jmroot:
Maybe this helps.
| `configure.cxx_stdlib` value | Flag for clang | Flag for gcc |
| --- | --- | --- |
|`libc++` | `-stdlib=libc++` | `-stdlib=libc++` |
| `macports-libstdc++` | `-stdlib=macports-libstdc++` | `-stdlib=libstdc++` (or none) |
| `libstdc++` | `-stdlib=libstdc++` | (unsupported configuration) |

cjones051073:
> Maybe this helps.
> 
> `configure.cxx_stdlib` value	Flag for clang	Flag for gcc
> `libc++`	`-stdlib=libc++`	`-stdlib=libc++`
> `macports-libstdc++`	`-stdlib=macports-libstdc++`	`-stdlib=libstdc++` (or none)
> `libstdc++`	`-stdlib=libstdc++`	(unsupported configuration)

I do not see why the bottom right box is unsupported ? Can you explain that ? Passing -stdlib=libstdc++ to gcc10+ works just fine, and is exactly equivalent to passing nothing.

cjones051073:
Is what you have been trying to say that the setting `configure.cxx_stdlib=libstdc++` means use *only* the system libstdc++, and *nothing* else ? If thats the case then that is the point Ken wasn't making clear to me.

If it is the case, I agree there is current no `-stdlib=XYZ` you can use to get gcc10+ to link against the system libstdc++. But, this was the case before I made my changes, so whilst broken, my change does not make things any more or less broken.

jmroot:
macports-gcc will never use the system libstdc++, which is what setting `configure.cxx_stdlib` to `libstdc++` is meant to do.

cjones051073:
> macports-gcc will never use the system libstdc++, which is what setting `configure.cxx_stdlib` to `libstdc++` is meant to do.

Right, I get it now. 

But as I said, my changes have not made things any worse or better.

I see two fixes. 

1. Patch our gcc builds to support some new `-stdlib=XYZ` that tells gcc to link against the system libstdc++, and then patch base to pass this whenever gcc is used tighter with `configure.cxx_stdlib=libstdc++`

2. basically ban the combination of requesting gcc(10+) to build c++ together with requesting `configure.cxx_stdlib=libstdc++`. Any build that requests gcc10+ to build should always make sure it then requests `configure.cxx_stdlib=macports-libstdc++`

Personally I would prefer 2., as patching gcc to support the system runtime sounds messy.

cjones051073:
> Yes. I'm certainly not saying you broke that case. This whole discussion was started by this statement:
> 
> > Any port that is doing this, so requesting macports-libstdc++ but then also using macports gcc to build, should probably be considered mis-configured.
> 
> If we agree that that is in fact a valid (and in some cases, the only valid) setting to use with macports-gcc, then I think we're done. :)

Yes, that statement I made was based on a mis-understanding of what `configure.cxx_stdlib=libstdc++` and `configure.cxx_stdlib=macports-libstdc++`

I think we are now in agreement.

kencu:
```configure.cxx_stdlib=libstdc++```

means the software is linked against ```/usr/lib/libstdc++.dylib`` and no other stdlib.

```configure.cxx_stdlib=macports-libstdc++```

means the software is linked against ```${prefix}/lib/libstdc++.dylib`` and no other.

kencu:
if you pass ```-stdlib=libstdc++``` to gcc12 and it links against ```${prefix}/lib/libstdc++.dylib```, as far as MacPorts is concerned, it is misconfigured. However, nobody might care.

if you pass ```-stdlib=macports-libstdc++``` to gcc12 expecting to link against ```${prefix}/lib/libstdc++.dylib```, you will error out. Which is what we are trying to prevent here.

cjones051073:
> if you pass `-stdlib=libstdc++` to gcc12 and it links against `${prefix}/lib/libstdc++.dylib`, as far as MacPorts is concerned, it is misconfigured. However, nobody might care.
> 
> if you pass `-stdlib=macports-libstdc++` to gcc12 expecting to link against `${prefix}/lib/libstdc++.dylib`, you will error out. Which is what we are trying to prevent here.

The second point there is addressed with https://github.com/macports/macports-base/commit/887b2fb974c8363413cdbdee65a4962e0cf02dd6

The first would require someone to implement this in the gcc ports, to define a new -stdlib=XYZ setting that would force the compiler to use the system libstdc++. I have my doubts this is worth the effort. Better at this port, if a users wishes to use a system provided c++ runtime to just use libc++ instead now that this is a (approximately) supported option, and hopefully will only improve with time.

PR #288, opened 2023-01-05
portconfigure: ensure SDK directory is not empty

MarcusCalhoun-Lopez:
On older versions of macOS (e.g. OS X El Capitan), even though "macosxfake" is not a valid SDK,
`env DEVELOPER_DIR="/Library/Developer/CommandLineTools" /usr/bin/xcrun --show-sdk-path --sdk macosxfake` has zero exit value but outputs an empty string.


jmroot:
When does this make a difference? Normally on these OS versions with the CLTs installed we would already have returned on line 560 or 567.

MarcusCalhoun-Lopez:
Instead of tracking down when the behavior of `xcrun` changed or worrying about the possibility of a Portfile changing `configure.sdk_version`, this seemed safer.

PR #296, opened 2023-08-20
startupitem: add new option startupitem.logfile.stderr, for capturing stderr

mascguy:
Presently, when `startupitem.logfile` is specified, we add a plist entry for stdout logging (via `StandardOutPath`). But when errors and such occur - particularly due to a fatal startup issue - that output often goes to stderr.

Not capturing the latter, makes troubleshooting more difficult for users. Potentially requiring them to manually edit the plist file, to add an entry for `StandardErrorPath` themselves.

To ensure the user doesn't have to deal with that, add new option `startupitem.logfile.stderr`.

By default, it is set to `startupitem.logfile`. But the user can specify a different logfile, clear the value (to disable stderr logging), or set it independently of `startupitem.logfile` (if only stderr should be logged).

Fixes: https://trac.macports.org/ticket/67955

### DIFF
--- a/doc/portfile.7
+++ b/doc/portfile.7
@@ -1996,6 +1996,20 @@ standard output from the daemon may also be directed to the logfile.
 .br
 .Sy Example:
 .Dl startupitem.logfile ${prefix}/var/log/mydaemon.log
+.It Ic startupitem.logfile.stderr
+Path to a logfile for capturing standard error output from the daemon.
+.br
+.Sy Type:
+.Em optional
+.br
+.Sy Default:
+.Em startupitem.logfile
+.br
+.Sy Values:
+.Em path
+.br
+.Sy Example:
+.Dl startupitem.logfile.stderr ${prefix}/var/log/mydaemon.log
 .It Ic startupitem.logevents
 Control whether or not to log events to the log file.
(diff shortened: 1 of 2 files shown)


jmroot:
I would like a separate option to control where stderr is sent. That can default to the same log as stdout.

mascguy:
> I would like a separate option to control where stderr is sent. That can default to the same log as stdout.

Agreed, as it provides more flexibility. Any thoughts as to the option name?

mascguy:
> I would like a separate option to control where stderr is sent. That can default to the same log as stdout.

How about this:
* Add new option `startupitem.logfile.stderr`
* Defaults to `startupitem.logfile`
* If cleared, disables stderr logging
* Can be set independently of `startupitem.logfile`, if user only wants stderr logging

mascguy:
PR updated to support new option `startupitem.logfile.stderr`, along with update to the manpage.

mascguy:
> PR updated to support new option `startupitem.logfile.stderr`, along with update to the manpage.

Note: Once we have consensus for all of the details, a corresponding PR will be created to update the MacPorts Guide.

herbygillot (approved):

PR #300, opened 2023-10-01
src: Add strl{cat,cpy},getline compat includes

neverpanic:
pextlib1.0/tracelib.c additionally had its own implementation, which is now no longer required since we have a centrally provided compatibility implementation.

machista1.0 did not yet use the compat implementation and thus probably did not work on operating systems that do not have strlcpy.


jmroot (approved):
Looks good, thanks.

PR #302, opened 2023-10-28
Specify missing `void` parameter in function declarators

chrstphrchvz:
Fixes: https://trac.macports.org/ticket/68320

### DIFF
--- a/src/darwintracelib1.0/darwintrace.c
+++ b/src/darwintracelib1.0/darwintrace.c
@@ -130,12 +130,12 @@ volatile bool __darwintrace_initialized = false;
  * interposed functions might end up being called first) we'll run them manually
  * before we interpose anything.
  */
-static void (*constructors[])() = {
+static void (*constructors[])(void) = {
 	__darwintrace_setup_tls,
 	__darwintrace_store_env,
 };
 
-void __darwintrace_run_constructors() {
+void __darwintrace_run_constructors(void) {
 	for (size_t i = 0; i < sizeof(constructors) / sizeof(*constructors); ++i) {
 		constructors[i]();
 	}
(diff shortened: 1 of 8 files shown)


chrstphrchvz:
See https://github.com/macports/macports-base/pull/301#issuecomment-1783645161 regarding CI failure.

raimue (approved):
Looks good to me. But I did not try to confirm whether this caught all warnings of this kind.

chrstphrchvz:
> Looks good to me. But I did not try to confirm whether this caught all warnings of this kind.

Currently, this PR does not address all `-Wstrict-prototypes` warnings. It only tries to deal with any which MacPorts itself would be responsible for. There are plenty more in vendor (Tcl, etc.) and autoconf tests; should this PR instead deal with those as well?

Also, this would only resolves warnings seen in `make all` and the macOS 13 build log the ticket referred to; I have not combed through the code to see if that missed any others (e.g. not compiled or hidden from the compiler).

chrstphrchvz:
Added fix for the one -Wstrict-prototypes warning seen during the Tcl 8.6.13 build. It will probably be accepted by upstream for Tcl 8.6.14.

raimue:
> Currently, this PR does not address all `-Wstrict-prototypes` warnings. It only tries to deal with any which MacPorts itself would be responsible for. There are plenty more in vendor (Tcl, etc.) and autoconf tests; should this PR instead deal with those as well?

I agree if we need it, we should try to submit this upstream if they did not already address this. Then sync it with a new upstream version. Just as you did now. :smiley: 

Our patches to the vendored sources should be kept minimal and only apply build fixes we absolutely need. And vendored code throws other warnings as well.

PR #303, opened 2023-10-30
daemondo/main.c: use static

chrstphrchvz:
As recommended by e.g. https://wiki.sei.cmu.edu/confluence/display/c/DCL15-C.+Declare+file-scope+objects+or+functions+that+do+not+need+external+linkage+as+static

### DIFF
--- a/src/programs/daemondo/main.c
+++ b/src/programs/daemondo/main.c
@@ -70,8 +70,8 @@
 #include <IOKit/IOMessage.h>
 
 // Constants
-const CFTimeInterval kChildDeathTimeout = 20.0;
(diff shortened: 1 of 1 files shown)


pmetzger:
Looks reasonable. Any idea why one of the CI runners failed?

jmroot:
> Looks reasonable. Any idea why one of the CI runners failed?

"SSL connection timeout"

jmroot (commented):
This really doesn't change anything since this is a main executable (it would fail to build if there were any name collisions). But I guess it doesn't do any harm.

PR #304, opened 2023-10-30
Migrate `CONST` usage to `const`

chrstphrchvz:
Currently, `CONST` (from tcl.h) is to be deprecated in Tcl 8.7, and removed in Tcl 9.0.


chrstphrchvz:
There is more `CONST` usage in vendor; this PR currently focuses on any usage which MacPorts alone would be responsible for.

pmetzger:
Thanks so much for this cleanup!

PR #305, opened 2023-10-31
Remove usage of deprecated `VOID` macro

chrstphrchvz:
The `VOID` macro (from tcl.h) was for compatibility with systems lacking `void` (where it would instead be defined to `char`). It is to be deprecated in Tcl 8.7 and removed in Tcl 9.0. Usage of `VOID *` in Tcl, such as for the `.internalRep.otherValuePtr` field of a `Tcl_Obj`, has been replaced with `void *`; casts are no longer needed when assigning to/from `.internalRep.otherValuePtr`.


pmetzger:
I only argue a bit with the commit message; not everything here is about VOID from what I can see in the actual changes.

chrstphrchvz:
Is the commit message okay now?

PR #307, opened 2023-11-15
tclx: update to 8.6.2

chrstphrchvz:
See https://github.com/flightaware/tclx/compare/v8.4.4...v8.6.2 for changes. These mainly seem to avoid compiler warnings/future errors, however I spot few if any functional changes, so I do not think there are any incompatibilities. I have not tested this beyond `make all` with clang-mp-17 on macOS 12.7.1.


neverpanic (approved):
This looks good to me.

raimue:
Sorry I am so late to this, but I think we also need to update the postflight script to get rid of the old files when the upgrade is installed with the .dmg installer: https://github.com/macports/macports-base/blob/3ac0da62a9c648f5963c329e9f049f63af7b2369/portmgr/dmg/postflight.in#L55-L68

jmroot:
I think that was already done in f78d33f. TclX was actually previous updated to 8.6 without changing the directory name, which is probably a source of confusion.

PR #308, opened 2023-11-15
pextlib: return 2-element list from mkstemp

chrstphrchvz:
…which when combined with the Tcl 8.5 `lassign` command should be simpler to use. The two uses updated by this PR are the only ones I am aware of.

Hopefully this makes `mkstemp` a slightly more attractive replacement for `mktemp` which relies on the `mktemp()` C function, which is deprecated since macOS 13 due to possible TOCTOU errors.

### DIFF
--- a/src/pextlib1.0/mktemp.c
+++ b/src/pextlib1.0/mktemp.c
@@ -43,6 +43,7 @@
 
 #include <errno.h>
 #include <stdlib.h>
+#include <stdio.h>
 #include <string.h>
 #include <unistd.h>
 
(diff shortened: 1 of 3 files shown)


chrstphrchvz:
Or should usage be replaced by Tcl 8.6 `file tempfile`?

neverpanic:
Both look good to me. #309 has the advantage of being less code, which is always good. Less code is less code that Apple might break in the future. I quickly checked the ports tree, but didn't see any (Tcl) use of mktemp or mkstemp, so I think I'd go for #309.

Feel free to merge either of them and close the other one.

neverpanic (approved):

PR #312, opened 2023-12-12
Document fetch --no-mirrors

kurthindenburg:


mascguy (approved):
LGTM, and this would be a great addition for the pending 2.9 release!

Haven't had a chance to test locally with it though.

Anyone else?

neverpanic (approved):
This is a clear improvement over the status quo, so I'm going to merge this now.

Ideally, we should however start creating a separate manpage for port-fetch in `port-fetch.1.txt` that describes in more detail what port-fetch does, move the catch-all content to the next file (maybe `port-checksum.1.txt`, since that's next in line) and adjust all the other `port-{extract,patch,configure,build,destroot,test}.soelim` links to point to that.

PR #313, opened 2023-12-30
Changelog typo fix

contextnerror:


ChangeLog
```diff
@@ -23,7 +23,7 @@ Release 2.9.0 (unreleased)
     - Dependencies installed for the first time when upgrading a port
       no longer inherit all variants from the dependent port. Arch
       checking is now done in this case so +universal is added to the
-      dependency iff needed. (#67637, jmr in 7be167c)
+      dependency if needed. (#67637, jmr in 7be167c)
```
raimue, line 26: `iff` is a common shorthand for `if and only if`. https://en.wikipedia.org/wiki/If_and_only_if

PR #314, opened 2024-01-02
selfupdate: Prints number of outdated ports after updating ports tree

thelastlin:
Hi,

Here is a tiny suggestion about user interface.
People may want to know how many ports can be upgraded after the ports tree has been updated by `port selfupdate`.

This patch prints about numbers of outdated ports like this:

```
# port selfupdate
--->  Updating MacPorts base sources using rsync
MacPorts base version 2.8.1 installed,
MacPorts base version 2.8.1 downloaded.
--->  Updating the ports tree
--->  MacPorts base is already the latest version

The ports tree has been updated. To upgrade your installed ports, you should run
  port upgrade outdated

2 ports are outdated. Run 'port outdated' to see them.
```

and when no installed ports are outdated will look like this:
```
# port selfupdate
--->  Updating MacPorts base sources using rsync
MacPorts base version 2.8.1 installed,
MacPorts base version 2.8.1 downloaded.
--->  Updating the ports tree
--->  MacPorts base is already the latest version

The ports tree has been updated. To upgrade your installed ports, you should run
  port upgrade outdated

All installed ports are up to date.
```

### DIFF
--- a/src/port/port.tcl
+++ b/src/port/port.tcl
@@ -2418,13 +2418,16 @@ proc action_selfupdate { action portlist opts } {
     if {[prefix_unwritable]} {
         return 1
     }
-    global global_options
+    global global_options ui_options
     set options [array get global_options]
     if {[dict exists $options ports_${action}_nosync] && [dict get $options ports_${action}_nosync] eq "yes"} {
         ui_warn "port selfupdate --nosync is deprecated, use --no-sync instead"
         dict set options ports_${action}_no-sync [dict get $options ports_${action}_nosync]
     }
-    if { [catch {macports::selfupdate $options base_updated} result ] } {
+    if {[info exists ui_options(ports_commandfiles)]} {
+        dict set options ports_${action}_presync 1
+    }
+    if { [catch {macports::selfupdate $options selfupdate_status} result] } {
         ui_debug $::errorInfo
         ui_error $result
         if {![macports::ui_isset ports_verbose]} {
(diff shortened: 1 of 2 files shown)


src/port/port.tcl
```diff
@@ -2798,6 +2798,20 @@ proc action_selfupdate { action portlist opts } {
         fatal "port selfupdate failed: $result"
     }
 
+    if {![info exists options(ports_${action}_no-sync)] || !$options(ports_${action}_no-sync)} {
+        if {[dict get $selfupdate_status needed_portindex]} {
```
thelastlin, line 2802: Does `needed_portindex` actually required if the port base has been updated with no `--no-sync` given?

Seems that we can check `base_updated` at here to print messages for rerun `port selfupdate`.

raimue: Moving this message to port.tcl looks good to me, as it is up to the client to restart the selfupdate (or ask the user to do so).

The check for --no-sync is the same as inside the macports1.0 API and that seems redundant. The `needed_portindex` would not be set otherwise.

But your change lost the condition on `base_updated` here.

As a side note, the macports1.0 API change is not backwards compatible, so we should only ship this with the next 2.x.0 release.

thelastlin: > The check for --no-sync is the same as inside the macports1.0 API and that seems redundant. The `needed_portindex` would not be set otherwise.

Checking for `--no-sync` might be required for messages about restarting the selfupdate at here.
It can be returned by `macports::selfupdate`, but I'm not sure if it's necessary since there are some similar checks in `port.tcl`.


> But your change lost the condition on `base_updated` here.

The check for `base_updated` might not be required at here, as`portindex` may not running in some cases:

https://github.com/macports/macports-base/blob/87a8c78fe9f4eb5894c4a46134d5db49f14d22ed/src/macports1.0/macports.tcl#L3034

raimue: This is when the ports tree was downloaded, but the Portfiles use new features of base, so they cannot be parsed with the old version. The message being printed only makes sense when base was updated.

KJshuhj:
Thanks for the contribution

jmroot:
Thanks for the update. I'm currently merging some other work involving selfupdate, so I'll come back to this once that's done.

jmroot (approved):

PR #318, opened 2024-03-19
registry2.0: Enable progress bar for activation & deactivation

neverpanic:
See the commit messages for more detailed descriptions:

```
d4461665c registry2.0: Improve extract progress bar
818d5d019 registry2.0: Enable progress bar for deactivation
63a832960 port: Tolerate progress bars where the 'start' action wasn't called
5ee0e4518 registry2.0: Enable progress bar for activation
```

### DIFF
--- a/src/registry2.0/portimage.tcl
+++ b/src/registry2.0/portimage.tcl
@@ -537,7 +537,16 @@ proc _extract_progress {event} {
             _progress start
         }
         stdin {
-            # we don't really care about the line, just that it exists
+            set line [string trimright [dict get $event line]]
+
+            # We only want to count files, not directories. Additionally,
+            # filter MacPorts metadata files that start with "+".
+
+            #   directories                        bsdtar output for metadata             gnutar output for metadata
+            if {[string index $line end] == "/" || [string range $line 0 4] eq "x ./+" || [string range $line 0 2] eq "./+"} {
+                return
+            }
+
             incr progress_step
             _progress update $progress_step $progress_total_steps
         }
--- a/src/registry2.0/portimage.tcl
+++ b/src/registry2.0/portimage.tcl
@@ -516,7 +516,7 @@ proc extract_archive_to_tmpdir {location} {
         } else {
             set cmdstring "${unarchive.pipe_cmd} ( ${unarchive.cmd} ${unarchive.pre_args} ${unarchive.args} )"
         }
-        system $cmdstring
+        system -callback portimage::_extract_progress $cmdstring
     } on error {_ eOptions} {
         ::file delete -force $extractdir
         throw [dict get $eOptions -errorcode] [dict get $eOptions -errorinfo]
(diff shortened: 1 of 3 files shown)


src/registry2.0/portimage.tcl
```diff
@@ -527,15 +527,59 @@ proc extract_archive_to_tmpdir {location} {
...
+            # tar output.
+            set line [string trimright [dict get $event line]]
+
+            # See port1.0/portinstall.tcl for the metadata filenames
+            if {[regexp {^x \./\+(STATE|PORTFILE|COMMENT|DESC|CONTENTS)$} $line]} {
+                return
+            }
+            if {[regexp {^x .*/$} $line]} {
+                return
+            }
```
jmroot, line 551: I'm not a fan of running two regexps on each line. How much does this affect the runtime? The second one at least could be a `string match` which is a lot faster, but is there no better way to estimate the total number of lines in advance?

Also, I believe different versions of tar, including the gnutar on older macOS versions, produce slightly different output.

jmroot (commented):
Looks pretty good overall.

src/registry2.0/portimage.tcl
neverpanic: Switched to using `[string index …]` and `[string range …]` with comparisons, which should be much faster. Not that I think it matters much, because the system calls to write to the file during extraction probably need a lot more processing power than the regexes, but you're right that we shouldn't be needlessly wasting CPU cycles either.

There's no good way to get the number of lines estimated in advance, because MacPorts does not keep track of the directories or metadata files in an archive. I could theoretically query the list of files from the registry, then walk over them to determine the number of directories, but we do that in the second pass already and it scales in the number of files, so I'm not convinced it would be faster. We also don't know in advance whether an archive will contain metadata files, and if so which ones, although we could probably get away just ignoring them since they're currently only 5.

The `+CONTENTS` file also does not list directories, and there's no guarantee that it will be at the start of the tar file, so extracting it to get the list of files doesn't seem like a good idea, either. Running `tar tvf` to get a list of files before extracting requires a full pass over the archive due to the tar format, and thus also incurs the entire decompression and read overhead, so that's out as well.

Good catch on the gnutar output, I have adjusted the range expressions to deal with that. If there are other output formats on older versions of macOS, directories will just be counted and the progress bar might jump from 47% back to 33.3% (it does that if the filtering isn't correct with the `google-cloud-sdk` port for me), which isn't great, but not a catastrophic failure.

jmroot: > Switched to using `[string index …]` and `[string range …]` with comparisons, which should be much faster. Not that I think it matters much, because the system calls to write to the file during extraction probably need a lot more processing power than the regexes, but you're right that we shouldn't be needlessly wasting CPU cycles either.

You're right of course, which is why I asked how much it actually affects the runtime. I've just been slightly horrified at seeing how much slower `regexp` is than an equivalent `string match` (in cases where both will work of course.)

jmroot: Out of idle curiosity I benchmarked it, and the difference is around 2 ms across the output of extracting texlive-fonts-recommended. :)

PR #319, opened 2024-03-25
port1.0: Fix lint on portfiles with universal variant

neverpanic:
tcp_wrappers is one of the Portfiles that failed because of this, since get_variant_description expects three arguments, and only two were passed.


jmroot (approved):
Oops. Thanks!

PR #320, opened 2024-04-01
Do not change stdlib of mpi wrapped gcc

numaryu:
This is to apply the same cxx_stdlib to mpi wrapped gcc compiler version higher than 10


jmroot:
Sorry this was forgotten. Unfortunately I'm not familiar with the MPI wrappers and can't confirm if this change is correct. I'm also not sure who would be the right person to ask.

pmetzger:
You could have asked an LLM.

PR #321, opened 2024-04-02
portpatch: pass --forward to avoid silent reverse patch

vallon:
On macOS ~~15~~ 13, if a patch is already applied, /usr/bin/patch asks the user if it is reversed.  If running interactively, the build will wait for user input (hang); if no stdin, it assumes reverse.  --forward prevents the "Reversed?" prompt and triggers failure (*.rej file and non-zero exit)

Root cause of macports/macports-ports#23318

I do not have any older macOS systems to test whether the '--forward' flag is available (or works).  Here is a script that demonstrates the behavior:

```
  rm -f targetA targetB targetC
  yes | head -n 20 | cat -n > file0
  sed -e '5s/y/n/' < file0 > file1
  diff -u file0 file1 > patch01
  cp file1 targetA
  patch --forward targetA < patch01 ; echo A $?
  sed -e '15s/y/n/' < file1 > file2
  diff -u file0 file2 > patch02
  cp file1 targetB
  patch --forward targetB < patch02 ; echo B $?
  cp file0 targetC
  patch --forward targetC < patch02 ; echo C $?
  diff targetC file2 ; echo diff $?
```

The --forward option should not generate any warnings.  Expected output (with noise removed):

```
A 1
B 1
C 0
diff 0
```

### DIFF
--- a/src/port1.0/portpatch.tcl
+++ b/src/port1.0/portpatch.tcl
@@ -50,7 +50,7 @@ options patch.asroot
 default patch.asroot no
 default patch.dir {${worksrcpath}}
 default patch.cmd {[portpatch::build_getpatchtype]}
-default patch.pre_args -p0
+default patch.pre_args "--forward -p0"
 
 proc portpatch::build_getpatchtype {args} {
     if {![exists patch.type]} {


vallon:
Tested by manually changing /opt/local/libexec/macports/lib/port1.0/portpatch.tcl, and tried using the old borgbackup Portfile, which previously gave trouble:

```
# port -v patch borgbackup
--->  Applying patches to borgbackup
--->  Applying patch-allow-msgpack-1.0.8.diff
Executing:  cd "/opt/local/var/macports/build/_opt_local_var_macports_sources_macports-ports_sysutils_borgbackup/borgbackup/work/borgbackup-1.2.8" && /usr/bin/patch --forward -p0 < '/opt/local/var/macports/sources/macports-ports/sysutils/borgbackup/files/patch-allow-msgpack-1.0.8.diff'
patching file setup.py
Ignoring previously applied (or reversed) patch.
1 out of 1 hunks ignored--saving rejects to setup.py.rej
patching file 'src/borg/helpers/msgpack.py'
Ignoring previously applied (or reversed) patch.
1 out of 1 hunks ignored--saving rejects to 'src/borg/helpers/msgpack.py.rej'
Command failed:  cd "/opt/local/var/macports/build/_opt_local_var_macports_sources_macports-ports_sysutils_borgbackup/borgbackup/work/borgbackup-1.2.8" && /usr/bin/patch --forward -p0 < '/opt/local/var/macports/sources/macports-ports/sysutils/borgbackup/files/patch-allow-msgpack-1.0.8.diff'
Exit code: 1
Error: Failed to patch borgbackup: command execution failed
Error: See /opt/local/var/macports/logs/_opt_local_var_macports_sources_macports-ports_sysutils_borgbackup/borgbackup/main.log for details.
Error: Follow https://guide.macports.org/#project.tickets if you believe there is a bug.
Error: Processing of port borgbackup failed
```

src/port1.0/portpatch.tcl
```diff
@@ -50,7 +50,7 @@ options patch.asroot
 default patch.asroot no
 default patch.dir {${worksrcpath}}
 default patch.cmd {[portpatch::build_getpatchtype]}
-default patch.pre_args -p0
+default patch.pre_args "--forward -p0"
```
ryandesign, line 53: Since `patch.pre_args` has always been used only for the `-p0` argument, it has not been uncommon for ports to override `patch.pre_args` when they need a different patch level such as `-p1`; there are currently 233 portfiles that do this, and three portfiles that erroneously use `patch.post_args` for this. So we either have to fix those 236 portfiles to use `patch.pre_args-replace -p0 -p1` (which perhaps arguably they should have been using all along) or we need to stash this flag somewhere else.

ryandesign (commented):
This PR would lessen the need for https://trac.macports.org/ticket/69649.

> On macOS 15, if a patch is already applied, /usr/bin/patch asks the user if it is reversed.

Since there isn't a macOS 15 yet, I assume you mean macOS 13 and 14, as previously discussed.

> I do not have any older macOS systems to test whether the '--forward' flag is available (or works).

It seems to be available at least as far back as Mac OS X 10.6; I didn't check earlier systems yet.

jmroot:
Can this be combined with `-t/--batch` to give us other more desirable behaviour? And what happens to the exit code in each case? The manual uses terms like "skip" and "ignore" but what we really want is "error out".

vallon:
> The manual uses terms like "skip" and "ignore" but what we really want is "error out".

I noticed that as well.  It seems like a documentation bug.  The --forward option seems to disable the "Reversed?" login (prompt and reverse-patch).  If the diff block cannot be applied as-is, it is written to \*.rej, and causes non-zero exit.

--batch says: "assume that patches are reversed if they look like they are".

--force does not allow reversed patches, but it also has some different behavior with respect to "Prereq:", though I don't know what that is or whether it would be used.

jmroot:
> --batch says: "assume that patches are reversed if they look like they are".

Right, that's why I wondered if `-N/--forward` would override that part.

vallon:
The snippet from the first update gives the same behavior for "patch --batch --forward" as "patch --forward".  "patch --force" also returns non-zero exit, but with slightly different output:

Using ```patch --batch --forward``` or ```patch --forward```:

```
patching file targetA
Ignoring previously applied (or reversed) patch.
1 out of 1 hunks ignored--saving rejects to targetA.rej
A 1
patching file targetB
Ignoring previously applied (or reversed) patch.
2 out of 2 hunks ignored--saving rejects to targetB.rej
B 1
patching file targetC
C 0
diff 0
```

Using ```patch --force```:

```
patching file targetA
1 out of 1 hunks failed--saving rejects to targetA.rej
A 1
patching file targetB
1 out of 2 hunks failed--saving rejects to targetB.rej
B 1
patching file targetC
C 0
diff 0
```

Both show "A1 B1 C0 diff0".

For completeness, just ```patch --batch``` shows "A 0" indicating that it silently reversed:

```
patching file targetA
A 0
# targetA == file0 here
patching file targetB
1 out of 2 hunks failed--saving rejects to targetB.rej
B 1
patching file targetC
C 0
diff 0
```

src/port1.0/portpatch.tcl
jmroot: Maybe we could use `patch.args` instead. Very few ports use it, and almost all of those that do are setting it to `-p1` and so should really be modifying `patch.pre_args` instead.

jmroot:
It looks as though these options are supported all the way back to Tiger. So I think `--batch --forward` is the winning combination, since it will also error out if the file to be patched can't be found, rather than hanging.

ryandesign:
You had previously changed MacPorts to use `-t` (a.k.a. `--batch`) last year n 7b1be05a1271e3af7872fb4ac0e46f13fcade9d7 but had to revert it in 82b064968d0ae5561911b850249939a7f77ae158 due to "unwanted side effects, like assuming that reversed patches should be applied in reverse". Sounds like using `--forward` will mitigate that. Was that the only unwanted side effect or were there others?

PR #322, opened 2024-04-04
Fix standard_configure.sh for out-of-source builds

ryandesign:
See: https://trac.macports.org/ticket/28001


neverpanic (approved):
Do out of source builds work now?

ryandesign:
No, but I will send more PRs if I can fix the other issues.

PR #324, opened 2024-04-07
Fixed typo

polluks:


doc/port.1
```diff
@@ -785,7 +785,7 @@ If any of the ports tree(s) uses a file: URL that points to a local subversion w
 .PP
 outdated
 .RS 4
-Lists the installed ports which need a
+Lists the installed ports which need an
```
neverpanic, line 788: This is a generated file, which is checked in to source control so that users without the required tools installed can still install MacPorts from source. Please fix this in `port.1.txt`, then re-generate the `port.1` file.

See https://github.com/macports/macports-base/blob/master/doc/Makefile.in#L190-L196 for instructions on how to re-generate the files.

neverpanic (changes requested):

doc/port.1
polluks: I see.

PR #325, opened 2024-04-07
Fixed typo

polluks:


doc/port.1.txt
```diff
@@ -366,7 +366,7 @@ copy with the user set to the owner of the working copy.
 +
 
 outdated::
-    Lists the installed ports which need a 'upgrade'.
+    Lists the installed ports which need an 'upgrade'.
```
raimue, line 369: Thank you for your attention to detail!

You could have continued this in the previous PR #324. You can force-push to your branch to replace the proposed commit with a new version. There was no need to open a new PR losing the previous discussion. As a general note, you should also use a specific branch name for your PR and not use your own "master" branch, as it is easy to confuse these or accidentally sync the branch from upstream.

You modified the right file now, but as explained in the previous PR, the `port.1` man page needs to be regenerated with asciidoc. That has to be done as part of any proposed update to the man pages. We have to keep both the source file and the rendered man page under version control. This resolves the bootstrap problem that there would be no way to generate the man page without asciidoc, libxslt, and docbook-xsl-nons which you (usually) cannot obtain without installing MacPorts first.

polluks: Aha, so I don't need to regenerate the runoff file myself?
By the way the synopsis `[[portname | pseudo-portname | port-expressions | port-url]]` looks strange. How do you optional twice like this?

raimue: The `port.1` as roff needs to be generated with asciidoc. This PR must update both `port.1.txt` and `port.1` together.

jmroot:
We want to avoid merge commits by rebasing, but I'll squash the multiple commits rather than make you do any more work on such a simple change.

PR #327, opened 2024-04-21
Create codeql.yml

shopdrop1:


ryandesign:
Could you explain what this is and why we would want it?

jmroot:
Looks like a spam account TBH. Reported.

PR #328, opened 2024-04-24
port lint: warn for "platforms darwin" as this is the default

pguyot:


pmetzger:
Seems like a good idea. Would like more opinions on the implementation. @jmroot and @ryandesign ?

src/port1.0/portlint.tcl
```diff
@@ -249,6 +249,30 @@ proc portlint::lint_checksum {checksum_string} {
...
+    global lint_platforms ports_lint_nitpick
+
+    foreach platform $platforms {
+        set platname [lindex $platform 0]
+        if {$platname ni $lint_platforms} {
+            lappend issues "Unknown platform: $platname"
+        }
+    }
+
+    if {[tbool ports_lint_nitpick] && [llength $platforms] == 1 && "darwin" in $platforms} {
```
jmroot, line 269: `[llength $platforms] == 1 && "darwin" in $platforms` could be simplified to `$platforms eq "darwin"`.

src/port1.0/portlint.tcl
```diff
@@ -562,15 +590,12 @@ proc portlint::lint_main {args} {
...
-                incr errors
-            } else {
-                ui_info "OK: Found platform: $platname"
-            }
+    if {$seen_platforms && [info exists platforms]} {
+        set platforms_lint [portlint::lint_platforms $platforms]
+
+        foreach err $platforms_lint {
+            ui_error $err
+            incr errors
```
jmroot, line 598: Redundantly setting `platforms` to the default value should not be an error, just a warning.

pmetzger:
How close is this to done now? Can we commit it?

pguyot:
@pmetzger re-reading the code I noticed an [unnecessary global declaration](https://github.com/macports/macports-base/compare/2fa37c2d7944e8c7835d0ce0d7b9e91143d29ac4..87dee4f4b0d3d2e16653290a0e2e3e67cb227673) that I left in the code . Otherwise I did take @jmroot comments into account and I believe it's ready.

PR #330, opened 2024-05-10
macports.tcl: fix build_arch choice for ppc

barracuda156:
This is a bit less code with a better result: unsupported 10.6 ppc will have the right build_arch being set.
While this change does not affect supported systems, it is just a better logic.


barracuda156:
@ryandesign @kencu @mascguy Could someone take a look please?

barracuda156:
@ryandesign @kencu @mascguy Bump

PR #331, opened 2024-05-21
Check for more warnings in configure logs

ryandesign:
Check for `-Wimplicit-int` and `-Wincompatible-function-pointer-types` warnings in configure logs because llvm.org clang 16 and the latest Apple clang consider those to be errors.

Closes: https://trac.macports.org/ticket/70017

-----

This is an initial attempt that works but it could possibly be improved. There is some code duplication with the implicit function declaration checks. And since function names are not involved in these warnings, no information other than the configuration log file the warning appeared in is shown. Maybe showing how many occurrences were found in each log file would be useful, though that improvement might also be helpful in the implicit function declaration checks.

### DIFF
--- a/src/port1.0/portconfigure.tcl
+++ b/src/port1.0/portconfigure.tcl
@@ -1902,6 +1902,50 @@ proc portconfigure::configure_main {args} {
     return 0
 }
 
+proc portconfigure::check_warnings {warning_flag} {
+    global \
+        workpath
+
+    set files [list]
+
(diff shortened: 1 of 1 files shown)


ryandesign:
Any objection to merging this? We need to get *something* out there so that people can start being notified of problems in configure scripts and fixing them.

jmroot (approved):
I guess this is OK as long as it performs reasonably.

PR #333, opened 2024-05-27
Migration: snapshot installed files and use to resolve path-based deps

jmroot:
Adding the lists of installed files to registry snapshots allows us to resolve path-based depspecs in the same way that a normal upgrade does. While this increases the size of the snapshots considerably, I think it's worth it for the correctness.

Lightly tested and confirmed to behave correctly with curl installed with its `path:share/curl/curl-ca-bundle.crt:curl-ca-bundle` dep satisfied by certsync.

### DIFF
--- a/src/cregistry/sql.c
+++ b/src/cregistry/sql.c
@@ -141,7 +141,7 @@ int create_tables(sqlite3* db, reg_error* errPtr) {
 
         /* metadata table */
         "CREATE TABLE registry.metadata (key UNIQUE, value)",
-        "INSERT INTO registry.metadata (key, value) VALUES ('version', '1.212')",
+        "INSERT INTO registry.metadata (key, value) VALUES ('version', '1.213')",
         "INSERT INTO registry.metadata (key, value) VALUES ('created', strftime('%s', 'now'))",
 
         /* ports table */
@@ -223,6 +223,16 @@ int create_tables(sqlite3* db, reg_error* errPtr) {
             ", FOREIGN KEY(snapshots_id) REFERENCES snapshots(id)"
             " ON DELETE CASCADE"
             ")",
+
+        /* Files registered to ports in snapshots.
+           Needed to resolve file-based dependencies. */
+        "CREATE TABLE registry.snapshot_files ("
+              "id INTEGER"
+            ", path TEXT"
+            ", FOREIGN KEY(id) REFERENCES snapshot_ports(id)"
+            " ON DELETE CASCADE"
+            ")",
+        "CREATE INDEX registry.snapshot_file_path ON snapshot_files(path)",
 
         "COMMIT",
         NULL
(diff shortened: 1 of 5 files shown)


src/cregistry/sql.c
```diff
@@ -224,6 +224,16 @@ int create_tables(sqlite3* db, reg_error* errPtr) {
             " ON DELETE CASCADE"
             ")",
 
+        /* Files registered to ports in snapshots.
+           Needed to resolve file-based dependencies. */
+        "CREATE TABLE registry.snapshot_files ("
+              "id INTEGER"
+            ", path TEXT"
+            ", FOREIGN KEY(id) REFERENCES snapshot_ports(id)"
+            " ON DELETE CASCADE"
```
neverpanic, line 233: I realize the code above already uses `FOREIGN KEY` and `ON DELETE CASCADE`, but unfortunately I don't think that currently works. https://www.sqlite.org/foreignkeys.html says this needs sqlite >= 3.6.19 (which I don't think is a huge issue), but it also says in [section 2](https://www.sqlite.org/foreignkeys.html#fk_enable) that foreign key support must be explicitly enabled using
```
PRAGMA foreign_keys = ON;
```
for each connection.

It seems that without this, the `ON DELETE CASCADE` just does nothing:

```
$ sqlite3 no-fk.db
-- Loading resources from /Users/clemens/.sqliterc
.headers ON
.separator "\t"
.nullvalue "NULL"
.mode box
SQLite version 3.46.0 2024-05-23 13:25:27
Enter ".help" for usage hints.
sqlite> CREATE TABLE ports (id INTEGER PRIMARY KEY, name TEXT COLLATE NOCASE);
CREATE TABLE ports (id INTEGER PRIMARY KEY, name TEXT COLLATE NOCASE);
sqlite> CREATE TABLE files (id INTEGER, path TEXT, FOREIGN KEY(id) REFERENCES ports(id) ON DELETE CASCADE);
CREATE TABLE files (id INTEGER, path TEXT, FOREIGN KEY(id) REFERENCES ports(id) ON DELETE CASCADE);
sqlite> INSERT INTO ports (id, name) VALUES (1, 'port1'), (2, 'port2');
INSERT INTO ports (id, name) VALUES (1, 'port1'), (2, 'port2');
sqlite> INSERT INTO files (id, path) VALUES (1, '/path/port1/foo'), (1, '/path/port1/bar'), (2, '/path/port2/foo'), (2, '/path/port2/bar');
INSERT INTO files (id, path) VALUES (1, '/path/port1/foo'), (1, '/path/port1/bar'), (2, '/path/port2/foo'), (2, '/path/port2/bar');
sqlite> DELETE FROM ports WHERE id = 1;
DELETE FROM ports WHERE id = 1;
sqlite> SELECT * FROM files WHERE id = 1;
SELECT * FROM files WHERE id = 1;
┌────┬─────────────────┐
│ id │      path       │
├────┼─────────────────┤
│ 1  │ /path/port1/foo │
│ 1  │ /path/port1/bar │
└────┴─────────────────┘
sqlite> ^D
```

It does work with it enabled:
```
$ sqlite3 fk.db
-- Loading resources from /Users/clemens/.sqliterc
.headers ON
.separator "\t"
.nullvalue "NULL"
.mode box
SQLite version 3.46.0 2024-05-23 13:25:27
Enter ".help" for usage hints.
sqlite> PRAGMA foreign_keys = ON;
PRAGMA foreign_keys = ON;
sqlite> CREATE TABLE ports (id INTEGER PRIMARY KEY, name TEXT COLLATE NOCASE);
CREATE TABLE ports (id INTEGER PRIMARY KEY, name TEXT COLLATE NOCASE);
sqlite> CREATE TABLE files (id INTEGER, path TEXT, FOREIGN KEY(id) REFERENCES ports(id) ON DELETE CASCADE);
CREATE TABLE files (id INTEGER, path TEXT, FOREIGN KEY(id) REFERENCES ports(id) ON DELETE CASCADE);
sqlite> INSERT INTO ports (id, name) VALUES (1, 'port1'), (2, 'port2');
INSERT INTO ports (id, name) VALUES (1, 'port1'), (2, 'port2');
sqlite> INSERT INTO files (id, path) VALUES (1, '/path/port1/foo'), (1, '/path/port1/bar'), (2, '/path/port2/foo'), (2, '/path/port2/bar');
INSERT INTO files (id, path) VALUES (1, '/path/port1/foo'), (1, '/path/port1/bar'), (2, '/path/port2/foo'), (2, '/path/port2/bar');
sqlite> DELETE FROM ports WHERE id = 1;
DELETE FROM ports WHERE id = 1;
sqlite> SELECT * FROM files WHERE id = 1;
SELECT * FROM files WHERE id = 1;
sqlite> ^D
```

neverpanic (changes requested):
Great idea! Unfortunately I completely missed that our foreign key support doesn't actually work.

src/cregistry/sql.c
jmroot: I guess I would have discovered this when trying to implement snapshot deletion, which is still on the todo list. :)

jmroot:
Well, that introduces some new failures because `registry::entry delete` apparently doesn't handle foreign keys properly. Fixing that is outside the scope of this PR, so I might just handle it in separate changes on master.

@neverpanic Do you think this is otherwise OK?

neverpanic (approved):
Yeah, looks OK otherwise. This is a pre-existing problem, so no need to fix it in this PR. I believe if I saw the code correctly, this also affects non-migration code already, for the portgroups table.

We may even need some cleanup code for that.

PR #334, opened 2024-06-01
Remove unnecessary seen_platforms variable

ryandesign:
Perhaps I misunderstand something but why did a `seen_platforms` variable have to be created in 5297f1cebe34b6e635fee0aeed507e54330fbe25? It seems to work without that.


ryandesign:
I see: if this is removed, then it also warns for ports that don't set `platforms`.

PR #335, opened 2024-06-03
CI: Remove macos-11

ryandesign:
The macos-11 runner is slated for removal by the end of the month with brownouts scheduled to occur in the weeks leading up to that.


pmetzger:
This seems hard to object to given that the facility won't even exist soon. I will merge it.

jmroot:
> This seems hard to object to given that the facility won't even exist soon. I will merge it.

I mean, I mildly object to doing it earlier than necessary, but whatever.

PR #337, opened 2024-06-04
Escape special regex characters in depspecs

ryandesign:
It seems like a bug, rather than an intended feature, that "foo" in depspecs like `bin:foo:bar`, `lib:foo.1:bar`, and `path:a/b/foo:bar` was interpreted as a regular expression rather than as a simple string. The MacPorts Guide has always documented these to be simple strings, not regexes.

This fixes handling of depspecs that contain special regex characters -- at present that's `path:lib/pkgconfig/gtk+-2.0.pc:gtk2` and `path:lib/pkgconfig/gtk+-3.0.pc:gtk3` -- so that gtk2-devel and gtk3-devel, respectively, can satisfy them.

Closes: https://trac.macports.org/ticket/69585

### DIFF
--- a/src/port1.0/portutil.tcl
+++ b/src/port1.0/portutil.tcl
@@ -3186,12 +3186,13 @@ proc _libtest {depspec {return_match 0}} {
     if {$i < 0} {set i [string length $depline]}
     set depname [string range $depline 0 $i-1]
     set depversion [string range $depline $i end]
-    regsub {\.} $depversion {\.} depversion
     if {${os.platform} eq "darwin"} {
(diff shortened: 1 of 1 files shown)


jmroot:
It might be possible to not even use `regexp` to match these, which could simplify things further. I'll investigate.

ryandesign:
Yes I wondered about that. This code has always used regexps ever since it was added in 2002. I wasn't sure if there might be other uses of `_mportsearchpath` that still needed regexp capability (I didn't try to find out) so I went with the minimally invasive fix.

jmroot:
These do appear to be the only places `_mportsearchpath` is used.

PR #339, opened 2024-06-20
Add a `curl version` Tcl command to Pextlib.

RJVB:
This command can be useful for debugging connectivity issues  when Pextlib becomes upgradeable via a dedicated port and uses a libcurl that is also installed through its own dedicated port.

The command returns a Tcl dict containing information about the libcurl version that Pextlib was built against as well as the libcurl it is currently running against; a simplified equivalent of `curl --version`.
Obtaining a list of names of available feature is possible only since curl v7.87.0 so this information will probably be replaced with a simple feature mask in the Pextlib build installed by the MacPorts installer on most older Darwin versions.

See trac ticket #51516

### DIFF
--- a/src/pextlib1.0/curl.c
+++ b/src/pextlib1.0/curl.c
@@ -91,6 +91,7 @@ int CurlFetchCmd(Tcl_Interp* interp, int objc, Tcl_Obj* const objv[]);
 int CurlIsNewerCmd(Tcl_Interp* interp, int objc, Tcl_Obj* const objv[]);
 int CurlGetSizeCmd(Tcl_Interp* interp, int objc, Tcl_Obj* const objv[]);
 int CurlPostCmd(Tcl_Interp* interp, int objc, Tcl_Obj* const objv[]);
+int CurlVersionCmd(Tcl_Interp* interp, int objc, Tcl_Obj* const objv[]);
 
 typedef struct {
 	Tcl_Interp *interp;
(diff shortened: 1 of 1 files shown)


RJVB:
(FWIW, a version command for Pextlib itself might be useful too!)

src/pextlib1.0/curl.c
```diff
@@ -1435,6 +1436,61 @@ CurlPostCmd(Tcl_Interp* interp, int objc, Tcl_Obj* const objv[])
...
+#endif
+		{
+			// it would be possible to generate a list of feature names from the feature mask
+			// but tedious because it would involve checking for the existence of many of the 
+			// feature tokens; probably not justified for the intended use of the `curl version`
+			// command (primarily debugging).
+			Tcl_DictObjPut(interp, tcl_result, Tcl_NewStringObj("FEATURE_MASK", -1), Tcl_NewIntObj(theVersionInfo->features));
+		}
+		Tcl_SetObjResult(interp, tcl_result);
+	} while (0);
```
neverpanic, line 1489: Looks good to me, but why the `do { … } while (0)`? It doesn't seem like you're using `break` for control flow, and if you just want to scope the variables, just use a bare scope (although I wouldn't consider using a scope necessary here).

neverpanic (changes requested):

src/pextlib1.0/curl.c
RJVB: Oh, *that*'s why the construct is used in other functions... I completely overlooked that aspect and just copied the non-loop as something that was apparently required ... :-/

neverpanic (approved):

neverpanic:
Now.

RJVB:
>Now.

:)

PR #343, opened 2024-07-24
snapshot: UX improvements

neverpanic:
- snapshot: Report match in --diff if there are no changes  
  The user experience of `sudo port snapshot --diff $id --all` just printing nothing is not great. Improve this by explicitly printing a success message if there are no changes.
- snapshot: Handle errors with unknown snapshot IDs


jmroot (approved):
Looks fine, thanks.

PR #344, opened 2024-07-25
snapshot: Add JSON export & import functionality

neverpanic:
This will eventually allow users to move to a new machine by running

```
old> sudo port snapshot
old> sudo port snapshot --export $ID >snapshot.json
new> sudo port snapshot --import snapshot.json
new> sudo port restore --latest
```

Switch the progress bars to stderr, so that `snapshot --export` can use stdout for data while still displaying a progress bar.


pmetzger:
One thing I'll note: I often use having to rebuild my ports as an opportunity to clean things up and/or rethink what I have installed. It would be nice to make sure that one can do things like only exporting the requested ports (and then editing them), and saving / restoring the state of the `port select` stuff.

neverpanic:
> One thing I'll note: I often use having to rebuild my ports as an opportunity to clean things up and/or rethink what I have installed. It would be nice to make sure that one can do things like only exporting the requested ports (and then editing them), and saving / restoring the state of the `port select` stuff.

`port restore` only restores requested ports and their dependencies.

You can of course edit the exported JSON file, although I'm not sure how comfortable this is going to be considering the file will be pretty large (around 25M for my installation of some ~600 ports).

What this will do is provide a way to reinstall things with the default variants, since `port restore` will only preserve requested variants, too.

pmetzger:
Certainly it will be easier to edit the JSON if it is readable, though one can always use jq.

PR #346, opened 2024-08-13
allow force activation of already active ports

RJVB:
Allows to use `port -f activate foo` to activate (reactivate) ports that are already active.
Useful as a shortcut to `port deactivate foo && port activate foo` when that operation is required, but also when activating multiple ports e.g. when one or more already turn out to have the desired variant/version activated.


RJVB:
The selection list was shown twice if multiple versions/variants of a port are installed and the active one selected. Fixed.

PR #347, opened 2024-08-13
Add the most common git commands to the port driver

RJVB:
This adds `port {history|diff|status|add|commit [--amend]} <portname|portdir>`

Simple but practical shorthand for `git -C `port dir foo` <command> .`

Summary documentation is provided in the DEVELOPER ACTIONS section of the main man page.

Notes:
1. `port diff foo` saves the output to a tempfile because it can potentially be very large and thus less suitable to store in a temp. variable
2. `port commit` requires the `GIT_EDITOR` or `EDITOR` env. variable to be set (or the equivalent in .gitconfig). I decided not to set it from `MP_EDITOR` because that variable is documented as the one for editing Portfiles and might thus be set to something like `open -a xcode` or another command that spawns an advanced GUI editor *and returns immediately*. The error message printed on failure points out that a .gitconfig file should be present in the macport_user's HOME directory (the full path is given).
3. `port add foo` adds all untracked files in `port dir foo` that are not blacklisted via .gitignore
4. in case port trees not distributed via git are still a thing, should there be a dedicated mechanism to check if the specified port lives in such a tree, rather than letting git report the fact?

### DIFF
--- a/doc/port.1
+++ b/doc/port.1
@@ -1268,6 +1268,29 @@ livecheck
 .RS 4
 Check if the software hasn\(cqt been updated since the Portfile was last modified\&.
 .RE
+.PP
+history
+.RS 4
+Show the git commit history for \fIportname\fR if its portdir lives in a git repository\&.
+.RE
+diff
+.RS 4
+Show non-committed changes made to \fIportname\fR (if its portdir lives in a git repository)\&.
+.RE
+status
+.RS 4
+Show the git status of \fIportname\fR (if its portdir lives in a git repository)\&.
+.RE
+add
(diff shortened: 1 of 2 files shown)


raimue:
If you want something like this, it should be more generic. For example it could be a `port git <portname> ...` command to just execute a git command inside the port directory. That leaves more freedom to use the version control tool freely without having to reimplement all the functionality as a pass-through.

RJVB:
Yeah, no, thanks. If I wanted to type the extra words I would probably already have written a `git-port` script that should allow me to call `git port <whatever>`.

PR #349, opened 2024-08-18
Set a user-agent for curl isnewer.

thelastlin:
Fix sync warning about checking update failed due to rejected connection without a user-agent on some mirrors.

The user-agent for curl isnewer is conform to https://trac.macports.org/ticket/40319.


ryandesign:
There's now [a ticket for this](https://trac.macports.org/ticket/70973) so when this is merged the commit message should be amended to indicate that it closes the ticket.

PR #350, opened 2024-09-25
cp: NULL-terminate new_argv and use simpler execv instead of execve

ryandesign:
This is intended to fix https://trac.macports.org/ticket/70897 but I could not reproduce the problem so I don't know if this fixes it.


jmroot (approved):
Looks fine, thanks.

PR #354, opened 2024-11-03
Fix trace mode on arm64

neverpanic:
Fix trace mode on arm64 by stripping the CPU subtype that enables pointer authentication in the Mach-O headers of Apple's binaries.

Re-enable darwintrace proc tests on arm64.


neverpanic:
Actually, this code should probably be updated with bounds checks for all the access into the `mmap(2)`'d file.

neverpanic:
> Actually, this code should probably be updated with bounds checks for all the access into the `mmap(2)`'d file.

Done.

neverpanic:
Note that this doesn't correctly work on macOS 14 yet. See the discussion in the ticket.

mascguy:
Clemens, is this effort still ongoing, or...?

PR #356, opened 2024-11-18
programs/cp: Fix pointer cast warning

neverpanic:
`execve()` expects the args to be in writable memory, so copy them using `strdup(3)`.

### DIFF
--- a/src/programs/cp/cp.c
+++ b/src/programs/cp/cp.c
@@ -2,15 +2,16 @@
 
 #include <stdio.h>
 #include <stdlib.h>
-#include <unistd.h>
+#include <string.h>
 #include <sys/errno.h>
+#include <unistd.h>
 
 int main(int argc, char *argv[])
 {
-    const char *cp_path = "/bin/cp";
-    const char *clone_arg = "-c";
-    const char **new_argv = malloc(sizeof(char *) * (argc+2));
-    if (new_argv) {
+    char *cp_path = strdup("/bin/cp");
+    char *clone_arg = strdup("-c");
+    char **new_argv = malloc(sizeof(char *) * (argc+2));
+    if (cp_path && clone_arg && new_argv) {
         new_argv[0] = cp_path;
         new_argv[1] = clone_arg;
         for (int i = 1; i <= argc; i++) {


saagarjha:
(Note: drive-by comment)

> execve() expects the args to be in writable memory,

It doesn't, actually: https://pubs.opengroup.org/onlinepubs/9799919799/ (scroll down to the "Rationale" section). I would suggest just marking the variables as `char *` and letting it be rather than heap allocating them for no reason.

neverpanic:
> > execve() expects the args to be in writable memory,
> 
> It doesn't, actually: https://pubs.opengroup.org/onlinepubs/9799919799/ (scroll down to the "Rationale" section). I would suggest just marking the variables as `char *` and letting it be rather than heap allocating them for no reason.

Thanks, that's helpful. However, the macOS headers declare `execve` in `/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/unistd.h` as:

```
int  execve(const char * __file, char * const * __argv, char * const * __envp) __WATCHOS_PROHIBITED __TVOS_PROHIBITED;
```

This matches the third row in the table of the rationale section of https://pubs.opengroup.org/onlinepubs/9799919799/functions/exec.html. We also don't actually know whether Apple's implementation of `execve(2)` conforms to this spec — AFAIK Apple doesn't make any claims about macOS' compliance to POSIX.1-2024, and the manpage doesn't clarify either whether `args` can be in read-only memory, or whether it conforms to the standard.

So we either need to live with the compiler warning, silence the compiler warning, or add a cast. The allocation fixes the warning regardless of what Apple's implementation does, and the cost of two small allocations compared to the following `execve(2)` is negligible.

src/programs/cp/cp.c
```diff
@@ -2,15 +2,16 @@
...
 int main(int argc, char *argv[])
 {
-    const char *cp_path = "/bin/cp";
-    const char *clone_arg = "-c";
-    const char **new_argv = malloc(sizeof(char *) * (argc+2));
-    if (new_argv) {
+    char *cp_path = strdup("/bin/cp");
+    char *clone_arg = strdup("-c");
+    char **new_argv = malloc(sizeof(char *) * (argc+2));
+    if (cp_path && clone_arg && new_argv) {
```
raimue, line 14: Putting all into one condition will not work well with the `perror()` below. The `errno` could have been overwritten by the next library function.

raimue:
Couldn't you also allocate the static strings on the stack with `char clone_arg[] = "-c";`?

saagarjha:
> This matches the third row in the table of the rationale section of https://pubs.opengroup.org/onlinepubs/9799919799/functions/exec.html.

That's the source, for a `char *const[]` destination we need to check the columns. That accepts `char *[]` as expected. Also, macOS is one of the few certified UNIX OSes, which means that it actually does need to conform to the standards. Not the 2024 version of course, but this one for SUSv3: https://pubs.opengroup.org/onlinepubs/009695399/functions/exec.html. It has the same language in it.

neverpanic:
> Couldn't you also allocate the static strings on the stack with `char clone_arg[] = "-c";`?

Good catch, in 1b43b563cf14d7b1b2e3d6d43e1c76de577f218a.

src/programs/cp/cp.c
3343746491:
```suggestion
    if (cp_path && clone_arg && new_argv) {
```

PR #359, opened 2024-11-23
Do not lint platforms for default on subports

pguyot:
The check that platforms isn't equal to darwin is not performed on a subport as there is a specific logic for subports.

Closes: https://trac.macports.org/ticket/70997


jmroot (commented):
Unfortunately this doesn't fix the problem in all cases. See for example `swig`, which sets `platforms any` in the subports, but leaves it at the default of `darwin` in the top level port.

PR #361, opened 2025-01-05
portindex: spoof compiler version

mohd-akram:
Fixes: https://trac.macports.org/ticket/71755

Versions from https://gist.github.com/yamaya/2924292

### DIFF
--- a/src/port/portindex.tcl
+++ b/src/port/portindex.tcl
@@ -44,6 +44,43 @@ set worker_init_script {
 package require macports
 package require Thread
 
+# Spoof the compiler version
+rename macports::get_compiler_version macports::__get_compiler_version
+proc macports::get_compiler_version {compiler developer_dir} {
+    global port_options
+    if {![dict exists $port_options os.major]} {
+        return [macports::__get_compiler_version $compiler $developer_dir]
+    }
+    set os_major [dict get $port_options os.major]
+    switch -- [file tail ${compiler}] {
+        clang {
+            if {${os_major} >= 22} {
+                return "1500"
+            } elseif {${os_major} >= 17} {
+                return "1000.11.45.2"
+            } elseif {${os_major} >= 15} {
+                return "800.0.38"
+            } elseif {${os_major} >= 14} {
+                return "602.0.49"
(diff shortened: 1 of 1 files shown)


jmroot (changes requested):
We don't want to do this all the time, but only in pretend mode. This is also only one factor that can affect the output. If we're going to go down this path, I'd like to see a general mechanism for overriding variables, with the policy of what to set them to left to the caller.

mohd-akram:
> We don't want to do this all the time, but only in pretend mode.

Currently the code is checking `port_options` to only do this in pretend mode.

> This is also only one factor that can affect the output.

What are some others?

> If we're going to go down this path, I'd like to see a general mechanism for overriding variables, with the policy of what to set them to left to the caller.

Within the scope of this PR or in the future? Do you mean the caller should specify the compiler version? I refrained from that because the caller would have to have knowledge of which Xcode version is supported on which macOS version, and which compiler comes with that Xcode. `portindex` is also called from a shell script, so the user has no easy access to the MacPorts API to aid with information like this.

jmroot:
> > We don't want to do this all the time, but only in pretend mode.
> 
> Currently the code is checking `port_options` to only do this in pretend mode.

The interposing proc is always added.

> > This is also only one factor that can affect the output.
> 
> What are some others?

Potentially anything, which is why a general mechanism is needed. At least `xcodeversion` and `xcodecltversion`.

> > If we're going to go down this path, I'd like to see a general mechanism for overriding variables, with the policy of what to set them to left to the caller.
> 
> Within the scope of this PR or in the future?

Do it once, do it right.

> Do you mean the caller should specify the compiler version? I refrained from that because the caller would have to have knowledge of which Xcode version is supported on which macOS version, and which compiler comes with that Xcode. `portindex` is also called from a shell script, so the user has no easy access to the MacPorts API to aid with information like this.

"The caller" will in practice virtually always be `mprsyncup`, which can just as easily hardcode a list like this, or do something smarter with a Tcl script if desired. It can be updated whenever we want, whereas `portindex` requires a base release to update.

PR #364, opened 2025-03-06
livecheck: add git support

mohd-akram:


src/port1.0/portlivecheck.tcl
```diff
@@ -234,6 +236,21 @@ proc portlivecheck::livecheck_main {args} {
                 }
             }
         }
+        "git" {
+            if {![string length ${livecheck.branch}]} {
```
jmroot, line 240: I know there are a few places in base where this idiom is used, but there's not really any point to it. You can just write `${livecheck.branch} eq {}`.

jmroot (changes requested):
Looks good for the most part.

PR #365, opened 2025-03-16
Give Xcode more permissions in the sandbox.

cdavis5e:
Grant access to Xcode's caches, as well as the `worksrcpath`. If the ports DB or the build directory[1] are on other volumes, grant access to the `TemporaryItems` directory of that volume as well. Grant permission to create jobs.

[1]: Since I have 96 GiB of RAM, I have placed the build directory on a tmpfs to avoid wear and tear on my SSD. I just hope that no port needs more than 48 GiB to build...

### DIFF
--- a/src/port1.0/portsandbox.tcl
+++ b/src/port1.0/portsandbox.tcl
@@ -43,9 +43,10 @@ default portsandbox_profile {}
 # command line usage would be:
 # sandbox-exec -p '(version 1) (allow default) (deny file-write*) (allow file-write* <filter>)' some-command
 proc portsandbox::set_profile {target} {
-    global os.major portsandbox_profile workpath distpath \
-        package.destpath configure.ccache ccache_dir \
-        sandbox_network configure.distcc porttrace prefix_frozen
+    global os.major portsandbox_profile workpath worksrcpath distpath \
+        package.destpath configure.ccache ccache_dir portdbpath \
+        sandbox_network configure.distcc porttrace prefix_frozen \
+        build.type
 
     switch $target {
         activate -
@@ -82,6 +83,16 @@ proc portsandbox::set_profile {target} {
     if {${configure.ccache}} {
         lappend allow_dirs $ccache_dir
     }
+    if {${build.type} eq "xcode"} {
+        # whitelist directories used by Xcode tools
+        lappend allow_dirs $portdbpath/home/Library/Developer/Xcode \
+                           $portdbpath/home/Library/Caches \
+                           $portdbpath/home/Library/org.swift.swiftpm \
+                           $portdbpath/home/.swiftpm \
+        # explicitly whitelist source dir to work around problems building
+        # Xcode projects in-source
+        lappend allow_dirs $worksrcpath
+    }
 
     set portsandbox_profile "(version 1) (allow default) (deny file-write*) \
 (allow file-write-data (literal \"/dev/null\") (literal \"/dev/zero\") \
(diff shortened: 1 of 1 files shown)


neverpanic (approved):

neverpanic:
Let me re-run the test failure to see if it is caused by this change.

cdavis5e:
I can't explain this. Tests are all green on my machine. Why is `portindex(1)` aborting?
```console
$ ag 'signal interp lost'
vendor/tclx-8.6.3/generic/tclXsignal.c
1543:        Tcl_Panic ("signal interp lost");
```

So something's going wrong deep inside Tcl... Does this happen under the same conditions as the test with my change reverted?

src/port1.0/portsandbox.tcl
```diff
@@ -82,6 +83,16 @@ proc portsandbox::set_profile {target} {
     if {${configure.ccache}} {
         lappend allow_dirs $ccache_dir
     }
+    if {${build.type} eq "xcode"} {
+        # whitelist directories used by Xcode tools
+        lappend allow_dirs $portdbpath/home/Library/Developer/Xcode
+        lappend allow_dirs $portdbpath/home/Library/Caches
+        lappend allow_dirs $portdbpath/home/Library/org.swift.swiftpm
+        lappend allow_dirs $portdbpath/home/.swiftpm
```
jmroot, line 91: This could all be one lappend call (still across multiple lines for readability).

src/port1.0/portsandbox.tcl
```diff
@@ -82,6 +83,16 @@ proc portsandbox::set_profile {target} {
...
+        # explicitly whitelist source dir to work around problems building
+        # Xcode projects in-source
+        lappend allow_dirs $worksrcpath
```
jmroot, line 94: Does this really make a difference? `$worksrcpath` is a subdir of `$workpath` which is already allowed.

src/port1.0/portsandbox.tcl
```diff
@@ -101,19 +112,21 @@ proc portsandbox::set_profile {target} {
...
     }
 
-    # If ${prefix} is on its own volume, grant access to its
+    # If ${prefix} or the ports DB is on its own volume, grant access to its
     # temporary items directory, used by Xcode tools
-    if {[catch {get_mountpoint ${prefix_frozen}} mountpoint]} {
-        ui_debug "get_mountpoint failed: $mountpoint"
-        set mountpoint /
-    }
+    foreach dir [list ${prefix_frozen} ${portdbpath} ${portdbpath}/build] {
```
jmroot, line 117: I guess this is OK, but it also seems like someone could potentially mount pretty much any path on a different volume.

src/port1.0/portsandbox.tcl
```diff
@@ -142,4 +155,10 @@ proc portsandbox::set_profile {target} {
             }
         }
     }
+
+    if {${build.type} eq "xcode"} {
+        # let Xcode create jobs (FIXME: narrow allowed commands)
+        append portsandbox_profile "\
+(allow job-creation)"
```
jmroot, line 162: Do you know if this syntax works on all macOS versions we run on?

jmroot:
> I can't explain this. Tests are all green on my machine. Why is `portindex(1)` aborting?

We sometimes get inexplicable SIGABRTs on CI. My only hypothesis is that maybe xprotect is being overzealous for some reason.

ryandesign:
> I just hope that no port needs more than 48 GiB to build...

We have a small number of ports (nodejs22 for example) that need a bit more than 15 GB, but I don't think I've seen any needing as much as 20 GB.

src/port1.0/portsandbox.tcl
cdavis5e: I thought so, too, but it actually does. Without it, Xcode builds in my setup fail with "You don't have permission to save the file." I can provide a log showing this if you need proof.

src/port1.0/portsandbox.tcl
jmroot: The buildbot tests should tell us if it causes errors on 10.6. If it works there I'll make a note to check 10.5 before releasing this.

src/port1.0/portsandbox.tcl
jmroot: I believe you, it just makes you wonder how you can know in general when allowing a directory does and doesn't allow its subdirs.

src/port1.0/portsandbox.tcl
cdavis5e: Done.

src/port1.0/portsandbox.tcl
cdavis5e: I've expanded the check to include all immediate subdirectories of the ports DB.

mervyd1993 (commented):
K

src/port1.0/portsandbox.tcl
ryandesign: User reported this fails on Mac OS X 10.5: https://trac.macports.org/ticket/72940

PR #366, opened 2025-04-18
src/port: Replace port-tclsh symlink with script

neverpanic:
Tcl attempts to determine the path of its standard library relative to the path of its binary, but does not expand symlinks when doing that. `port-tclsh` thus inadvertently uses libraries from a MacPorts-installed Tcl rather than the copies shipped with MacPorts.

See: https://trac.macports.org/ticket/64507
Closes: https://trac.macports.org/ticket/72359


mascguy:
Thanks for the quick fix Clemens!

PR #369, opened 2025-05-25
src/port: Replace port-tclsh with binary

neverpanic:
It seems like using scripts as interpreters in shebangs leads to unexpected effects on macOS, so let's just use a binary instead, which seems to work fine. Fortunately, this particular binary is very simple, too.

Closes: https://trac.macports.org/ticket/72530


Wlaeed13:
ماهاذي الرساله

في أحد، 25 مايو، 2025 في 10:39 م، كتب Clemens Lang ***@***.***

PR #372, opened 2025-09-18
Restore non-error if 'outdated' expands to nothing

neverpanic:
'sudo port upgrade outdated' specifically printed "Nothing to upgrade" due to a special case in port/port.tcl. A recent change using macports::upgrade_multi broke this, and now causes an error because $status is never set (the loop never runs).

Fix this by handling the empty portnames case explicitly and restore the previous message.


jmroot:
The message is actually still printed (by `require_portlist`) before the error happens.

PR #380, opened 2026-03-13
port: add 'prefix' action to display the configured prefix

herbygillot:
Adds a new `port prefix` command that prints the installation prefix MacPorts was configured with. Follows the same pattern as `port version` and `port platform`. Quiet mode (-q) suppresses the "Prefix: " label.

Also adds a dedicated port-prefix(1) man page and registers it in the doc build system.

Closes: https://trac.macports.org/ticket/72961

### DIFF
--- a/src/port/port.tcl
+++ b/src/port/port.tcl
@@ -2680,6 +2680,13 @@ proc action_platform { action portlist opts } {
 }
 
 
+proc action_prefix { action portlist opts } {
+    global macports::prefix
+    puts ${prefix}
+    return 0
+}
+
+
 proc action_dependents { action portlist opts } {
     if {[require_portlist portlist]} {
         return 1
@@ -4152,6 +4159,7 @@ set action_array [dict create \
     \
     version     [list action_version        [ACTION_ARGS_NONE]] \
     platform    [list action_platform       [ACTION_ARGS_NONE]] \
+    prefix      [list action_prefix         [ACTION_ARGS_NONE]] \
     \
     uninstall   [list action_uninstall      [ACTION_ARGS_PORTS]] \
     \
--- a/src/port1.0/tests/portmain.test
+++ b/src/port1.0/tests/portmain.test
@@ -35,6 +35,10 @@ set portpath .
 set portbuildpath ./build
 set macports::portdbpath .
 
+set test_tclsh [file normalize [file join $macports::autoconf::top_srcdir tests test-tclsh]]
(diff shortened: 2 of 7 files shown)


jmroot:
I always found this a bit absurd since you can't run `port` in the first place without using some knowledge of its installation prefix.

src/port/port.tcl
```diff
@@ -2681,6 +2681,16 @@ proc action_platform { action portlist opts } {
 }
 
 
+proc action_prefix { action portlist opts } {
+    global macports::prefix
+    if {![macports::ui_isset ports_quiet]} {
+        puts -nonewline "Prefix: "
+    }
```
neverpanic: While I agree with @jmroot that this is perhaps a bit superfluous, there have been questions for this, homebrew has it, it doesn't cost us much to support it, and I don't think this will need a lot of maintenance in the future. I also see value in having this in environments where somebody has symlinked `$prefix/bin/port` or done mount point shenanigans.

Given that its primary use case will be automated usage, though, I don't think we'd ever need or want the human-readable "Prefix: " header, and I'd just drop it both here an in the manpage.

Otherwise, I bet we'll just see users skip reading the manpage and constructing `sed`, `awk`, or `cut` statements to extract the value rather than passing `-q`.

neverpanic (changes requested):

src/port/port.tcl
herbygillot: Yep, change made - now it just prints the prefix without that header.

src/port1.0/tests/portmain.test
```diff
@@ -27,6 +27,18 @@ set portpath .
 set portbuildpath ./build
 set macports::portdbpath .
 
+set port_client_source [file normalize [file join $pwd .. .. port port.tcl]]
+set port_client_fd [open $port_client_source r]
+set port_client_data [read $port_client_fd]
+close $port_client_fd
+
+if {![regexp {(?s)(proc action_prefix \{ action portlist opts \} \{.*?\n\})} \
```
neverpanic: Sorry, didn't look at the test in the last iteration — extracting the definition from a file, evaluating it and then running it feels extremely contrived way to test this. Can we not instead maybe fork, run `port prefix` and check its output instead?

herbygillot: You're 100% right - resolved.

neverpanic (approved):

PR #381, opened 2026-03-13
configure: add --with-git option

herbygillot:
Add MP_TOOL_PATH(GIT, [git]) to configure.ac so that users can supply an alternate git binary via --with-git=PATH at configure time, matching the pattern used for rsync, openssl, cvs, and other tools.

Fixes: https://trac.macports.org/ticket/72451

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -122,6 +122,7 @@ dnl named with a _BIN suffix here.
 MP_TOOL_PATH(BSDMAKE, [bsdmake/pmake])
 MP_TOOL_PATH(BZIP2_BIN, [bzip2])
 MP_TOOL_PATH(CVS, [cvs])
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -122,6 +122,7 @@ dnl named with a _BIN suffix here.
 MP_TOOL_PATH(BSDMAKE, [bsdmake/pmake])
 MP_TOOL_PATH(BZIP2_BIN, [bzip2])
 MP_TOOL_PATH(CVS, [cvs])
+MP_TOOL_PATH(GIT, [git])
```
jmroot: Should the `AC_PATH_PROG(GIT, [git], [])` on line 153 go away?

jmroot:
Ah nope never mind, it works as is.

PR #394, opened 2026-03-18
Add project-wide agent instructions and expected standards

herbygillot:
These Markdown files help agents better understand the standards they need to follow for any work done against this codebase.


pmetzger:
Didn't notice this earlier. Certainly an AGENTS.md is important, and probably one is needed for macports-ports as well.

Is it actually necessary to have the CLAUDE.md if there's an AGENTS.md? I also note that you don't say much about testing process etc. That's probably a matter of some importance as well.

PR #395, opened 2026-03-20
port: add 'source' subcommand for managing sources.conf

herbygillot:
Add `port source` to list, add, remove, and manage default-flag state
of entries in sources.conf.

--add supports --first (insert before existing sources), --no-sync
(append [nosync] flag), and --default (mark as the explicit default,
clearing it from other sources).

--set-default and --unset-default toggle the [default] flag on
existing entries.

--remove warns if the removed source was the only explicit default.

Bare local paths starting with / or ~ are automatically normalized and
converted to file:// URLs; local paths with spaces are supported.

Write operations are atomic (write to temp file, then rename), check
for write permission, and emit a sudo hint on failure.

Business logic uses a structured read-parse-modify-write pipeline in
macports1.0/macports.tcl: source_read_conf / source_parse_line /
source_write_conf / source_render_record handle round-tripping the
file while preserving comments and blank lines. port.tcl contains the
thin UI wrapper (action_source).

Uses file tildeexpand before file normalize for Tcl 9 compatibility,
where file normalize no longer performs tilde expansion.

Includes 22 tcltest unit tests (sources.test), a new port-source(1)
man page (port-source.1.txt), and updates to port(1) and
sources.conf(5).

Fixes: https://trac.macports.org/ticket/73683

### DIFF
--- a/doc/Makefile.in
+++ b/doc/Makefile.in
@@ -79,6 +79,7 @@ MAN1=	\
 		port-setrequested.1 \
 		port-setunrequested.1 \
 		port-snapshot.1 \
+		port-source.1 \
 		port-space.1 \
 		port-sync.1 \
 		port-test.1 \
--- a/doc/port-source.1
+++ b/doc/port-source.1
@@ -1,0 +1,265 @@
+'\" t
+.TH "PORT\-SOURCE" "1" "2\&.12\&.99" "MacPorts 2\&.12\&.99" "MacPorts Manual"
+.\" -----------------------------------------------------------------
+.\" * Define some portability stuff
+.\" -----------------------------------------------------------------
+.\" ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+.\" http://bugs.debian.org/507673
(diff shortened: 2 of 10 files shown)


herbygillot:
@jmroot, I’m looking to merge this soon. Please take a look - thanks.

jmroot:
I'm not really convinced that this is worth doing.

herbygillot:
> I'm not really convinced that this is worth doing.

I think you're likely looking at this from the perspective of an experienced veteran of the tool.

A command like this just increases `port`'s newbie friendliness. Rather than tell people to edit a text config that they could mess up with typos, our documentation can just tell people to run a simple command (`port source ...`).

pmetzger:
I think this stuff is worth doing. I'm also really sad that my attempt to get something done recently that has *dramatically* increased the speed with which I can repair broken ports has been languishing.

Increasing the ease of use of MacPorts is, I think, a significant interest of the project.

jmroot:
That's the thing, I can't see how learning this CLI and the mental model behind it (which is still based on the config file structure) is any easier than learning the config file format. Feel free to start a discussion on the mailing list about the merits of this approach and ways of improving ease of use in general.

herbygillot:
Giving up and closing.

pmetzger:
I feel kind of demoralized here. What's the point?

PR #402, opened 2026-03-30
portextract: add file_extract command

herbygillot:
Add a new file_extract command that extracts archive files with automatic format detection from file suffixes. Supports gzip, bzip2, xz, lzma, lzip, zstd, compress, tar, zip, 7z, and dmg. Accepts -dirname to override the extraction directory and -type to override suffix-based detection. Relative filenames are resolved in filespath then distpath, and distfile tags are stripped before lookup.

This is intended to eventually replace the use_* extraction switches internally, while the use_* options remain available to Portfile authors.

Includes unit and integration tests, and portfile(7) man page documentation.

Fixes: https://trac.macports.org/ticket/50969

### DIFF
--- a/doc/portfile.7
+++ b/doc/portfile.7
@@ -2668,6 +2668,64 @@ Behaves identically to
 .Ic file delete -force .
 .It Ic touch
 Built-in command mimicking the BSD touch command.
+.It Xo
+.Ic file_extract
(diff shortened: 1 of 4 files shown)


herbygillot:
Closing this as I assume @jmroot’s changes around extraction are sufficient now.

PR #404, opened 2026-04-11
Fix trace mode on arm64

neverpanic:
I've been running this for months, if not years at this point, and while there are things that could be improved, this is certainly better than a completely broken trace mode, so let's finally get it merged.

- pextlib1.0: re-copy to SIP workaround if pextlib changed  
  This is required because the binaries will now be modified while copying, and it allows us to make future changes if necessary.
- pextlib1.0: Fix trace mode on arm64  
  The actual change, we're changing the CPU subtype on Apple's binaries and just remove the pointer authentication. Luckily the code stays valid and binaries don't crash. The `ln` binary as shipped on macOS 14 doesn't work with this because it uses uncommon relocations, but at this point macOS 14 is so old that I don't think this should still be a blocker. Other than that, I haven't encountered any binary that doesn't work with this hack.
- darwintracelib1.0: re-enable proc tests on arm64  
  Pretty simple bugfix on the side
- pextlib1.0: Fix corner case in trace mode re-signing  
  Addresses a crash I've occasionally seen that shouldn't happen, but does.
- tracelib: allow access to trace socket by any user  
  Fixes attempts to determine SOURCE_DATE_EPOCH and the port tree git commit if the port tree is owned by a different user (which it is on my system).


neverpanic:
A few of the tests can't handle not having MacPorts installed already, so I'll have to look into fixing those to  run in the source tree.

neverpanic:
Remaining failure is caused by the `ln` binary on macOS 14. We should probably just mark trace mode on arm64 as unsupported on that platform and move on.

PR #405, opened 2026-04-11
tests: consolidate duplicated test setup code and config files

herbygillot:
Extract the shared boilerplate (macports init, tmp prefix setup, mportinit, macports_worker_init) from 27 test files into a single test_setup.tcl sourced by all unit test suites.  Merge the three divergent copies of macports_worker_init into one consolidated superset.

Eliminate the duplicate sources.conf files in port1.0/tests/ and package1.0/tests/ by having init_tmp_prefix derive the path from top_srcdir, leaving a single copy in macports1.0/tests/.

Replace the hardcoded /tmp/macports-tests path in the integration test infrastructure with a unique temporary directory created via `file tempfile`, which respects TMPDIR and avoids collisions between parallel or repeated runs.  Move macports.conf and sources.conf content into static template files with @TEST_ROOT@ substitution so the config structure is visible on disk rather than generated inline.

Replace hardcoded /tmp paths in portutil.test with $pwd-relative paths.

Add configure~ to .gitignore.

Fixes: https://trac.macports.org/ticket/73828
Fixes: https://trac.macports.org/ticket/73827


herbygillot:
@neverpanic, @jmroot, can we merge this in?

herbygillot:
Just a reminder that we should merge this to clean up the tests.

herbygillot:
Actually let me update it for recent changes to master...

herbygillot:
OK, it's once again ready to merge.

herbygillot:
Will rebase/clean this up soon and merge in if there are no objections.

jmroot:
> Will rebase/clean this up soon and merge in if there are no objections.

I don't think there is anything seriously wrong with the changes merged, but procedurally, inviting objections and then merging 5 hours later is not helpful. That's not even enough time to allow members of a global project to see the initial message, let alone review the changes and formulate objections.

herbygillot:
> > Will rebase/clean this up soon and merge in if there are no objections.
> 
> I don't think there is anything seriously wrong with the changes merged, but procedurally, inviting objections and then merging 5 hours later is not helpful. That's not even enough time to allow members of a global project to see the initial message, let alone review the changes and formulate objections.

You're 100% right, my apologies.

PR #407, opened 2026-04-16
selfupdate: preserve all configure arguments

herbygillot:
During selfupdate, only a few configure arguments (--prefix, --with-install-user, --with-install-group, --with-directory-mode) were reconstructed and passed to configure. Other arguments such as --with-curlprefix, --with-applications-dir, --with-frameworks-dir, and tool paths were silently lost.

Save the original configure arguments to
$prefix/share/macports/configure_args during make install, and read them back during selfupdate. Fall back to the previous reconstruction logic when upgrading from an older version that lacks the saved file.

Closes: https://trac.macports.org/ticket/73670


neverpanic (approved):
Looks good to me, feel  free to merge

herbygillot:
Thanks @neverpanic

PR #409, opened 2026-04-22
portlint: pkgconfig dependency path-style

trustbyte:
Add to port lint a warning if port:pkgconfig is used instead of path-style dependency


ryandesign:
It's unusual to single out pkgconfig for this honor. There are over two hundred *foo*/*foo*-devel port pairs whose dependencies should also be written using `path:` style to allow either of them to be used.

trustbyte:
They bug me (and others) for this one.

The rest I did not see questioned, maybe I missed those but this pkgconfig is of abundance pointed out.

PR #410, opened 2026-05-10
Print environment in verbose mode

mohd-akram:
Debug prints too much, but this is useful to see.

### DIFF
--- a/src/port1.0/portutil.tcl
+++ b/src/port1.0/portutil.tcl
@@ -448,7 +448,7 @@ proc command_exec {args} {
     }
 
     # Debug that.
-    ui_debug "Environment: [environment_array_to_string ${varprefix}.env_array]"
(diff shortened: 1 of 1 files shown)


mohd-akram:
I've been using this for the past two weeks and it's been incredibly helpful. Thoughts?

pmetzger:
Might also be nice to have a debug flag that let  you chain a list of the things you wanted to see in the debug output...

jmroot:
Personally I very rarely need to look at the environment for a build even as a port developer. IMO it's fine for it to just be in the log if needed.

> Might also be nice to have a debug flag that let you chain a list of the things you wanted to see in the debug output...

Fine-grained control of debug output would be nice.

markemer:
Maybe allow like -d or -dd or -ddd or maybe just like -d 3? But I agree, fine grained control would rule.

PR #411, opened 2026-05-20
Add  unprivileged --local / -L Portfile development build mode

pmetzger:
## Summary

This adds a new `--local` / `-L` mode to `port` for Portfile development for doing development builds of ports without root privileges.

Local mode runs non-installing phases (and only non-installing phases) using the normal MacPorts phase machinery, but redirects the build state into a user-owned `./work` directory under the effective Portfile directory instead of to a symlink into a root-owned directory. It is intended for workflows such as:

```sh
cd devel/foo
port --local build
port --local destroot
port --local clean
```

or:

```sh
port -D ~/src/macports-ports/devel/foo -L build
```

## Behavior

In local mode, MacPorts now:

- allows only `fetch`, `checksum`, `extract`, `patch`, `configure`, `build`, `test`, `destroot`, and `clean`
- rejects installing or registry-mutating actions before phase work begins, you need to have all the dependencies for a port installed already.
- refuses to run as root
- uses `${portdir}/work` for build state
- uses local paths under `work` for distfiles, logs, `HOME`, `TMPDIR`, and destroot output
- avoids taking the registry write lock (and can't because it doesn't run as root).
- requires dependencies be installed and active
- fails clearly rather than installing missing dependencies

This mode is not intended to install software, create archives, etc. It's not a substitute for a real mechanism for doing unprivileged builds, which MacPorts could of course use. This is just a developer/sandbox mode for validating Portfile build machinery. When you're ready to actually produce a package or install locally, you need to do that as root.

## Tests

Added `tests/test/local-mode` coverage for:

- local destroot/build output under the Portfile directory
- local logs
- local clean removing only local `./work`
- rejected `install`
- missing dependency failure
- satisfied `bin:` dependency behavior against a read-only registry file

Validation run locally:

```sh
../tests/test-tclsh test.tcl -t local-mode -nocolor
make test
git diff --check
tests/test-tclsh src/port/port -D ~/src/macports-ports/sysutils/tree -L clean
```

### DIFF
--- a/doc/port.1
+++ b/doc/port.1
@@ -24,7 +24,7 @@ port \- Command line interface for MacPorts
 .SH "SYNOPSIS"
 .sp
 .nf
-\fBport\fR [\fB\-bcdfknNopqRstuvy\fR] [\fB\-D\fR \fIportdir\fR|\fIportname\fR] [\fB\-F\fR \fIcmdfile\fR] [\fIaction\fR] [\fIactionflags\fR]
+\fBport\fR [\fB\-bcdfkLnNopqRstTuvy\fR] [\fB\-D\fR \fIportdir\fR|\fIportname\fR] [\fB\-F\fR \fIcmdfile\fR] [\fIaction\fR] [\fIactionflags\fR]
      [[\fIportname\fR | \fIpseudo\-portname\fR | \fIport\-expressions\fR | \fIport\-url\fR]]
      [[\fI@version\fR] [+/\-variant \&...] \&... [option=value \&...]]
 .fi
@@ -524,6 +524,20 @@ Despite any errors encountered, proceed to process multiple ports and commands\&
 .RE
 .PP
 \fBDevelopment\fR
+.PP
+\-L, \-\-local
+.RS 4
+Run developer local build mode from a Portfile directory\&. This mode is intended for non\-root testing of edited Portfiles\&. It stores mutable build state, fetched distfiles, and logs under ${portpath}/work, uses the registry only to verify that dependencies are already installed and active, and does not install missing dependencies or modify the MacPorts installation\&. Supported actions are
+\fBfetch\fR,
+\fBchecksum\fR,
+\fBextract\fR,
+\fBpatch\fR,
+\fBconfigure\fR,
+\fBbuild\fR,
+\fBtest\fR,
+\fBdestroot\fR, and
+\fBclean\fR\&.
+.RE
 .PP
 \-o
 .RS 4
--- a/ChangeLog
+++ b/ChangeLog
@@ -3,6 +3,11 @@
 ###
 
 Release 2.13.0 (unreleased)
+    - Added --local (-L) developer build mode, which runs non-installing
+      phases as a regular user using work, distfiles, and logs under the
+      Portfile directory without modifying the MacPorts installation.
+      (perry)
+
     - Added a -T flag to prepend ISO 8601 timestamps to each line of
       output, and elapsed time logging for each phase.
       (#2020, herbygillot in 037d168)
--- a/doc/port.1.txt
+++ b/doc/port.1.txt
@@ -9,7 +9,7 @@ port - Command line interface for MacPorts
 SYNOPSIS
 --------
 [cmdsynopsis]
-*port* [*-bcdfknNopqRstuvy*] [*-D* 'portdir'|'portname'] [*-F* 'cmdfile'] ['action'] ['actionflags']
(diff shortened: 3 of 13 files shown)


pmetzger:
Just to point a few things out:

0. This small change is a big quality of life improvement for me at least. It means being able to do work on ports without having to constantly do a sudo dance. It also means being able to allow AI agents to help diagnosing port problems without needing to run them as root, which would be a pretty bad security problem.
1. I am not an expert at Tcl or MacPorts internal implementation, I've previously only really worked on maintaining the ports themselves. Also I had a bunch of help from Codex CLI. I've left this as a draft for now because I really want to review this with a true expert, ideally @jmroot, before we consider it good enough.
2. Ideally, MacPorts should be able to do all builds unprivileged etc. and should only require root for installing packages that have been prebuilt. It should probably even be able to sandbox the builds. However, that's much more work. This is a much more limited scope change.

pmetzger:
Any thoughts from anyone? I'm finding this really really useful day to day...

pmetzger:
Having worked with the tool more, there's an obvious change that's needed, which is to make "work" look more like "work" under normal circumstances because renames etc. fail. I think I'll make a few changes for that and push a new version.

mohd-akram:
I believe MacPorts uses root precisely to do unprivileged builds, as that's the only way to do `setuid`. You can probably solve your immediate problem with some sudoers configuration to let it run specific `port` commands without having to enter a password. There's also the `~/.macports/macports.conf` file which can be used to specify a different prefix so that it doesn't require root I believe. I previously used a hack like that to run `port` with HOME set to a temporary directory containing `.macports/macports.conf` to test some things. At that time it would have been nice to have a more specific environment variable specifying the location of the conf file rather than having to use the HOME hack.

pmetzger:
@mohd-akram I don't want to do an unprivileged build as the macports user id, that would not serve my purpose. I want to just do a build locally in the git checkout as me. It's working well, FWIW, though there are some nits I still want to fix in the code.

pmetzger:
@jmroot @herbygillot It seems that no one is interested in reviewing this; should I just close it?

jmroot:
This is still marked as a draft. It looks like largely a reinvention of the altprefix stuff which was removed in b5b549a. I'm not particularly convinced by the rationale either.

pmetzger:
I know of no other package manager that makes it absolutely impossible to test building a package without having a privileged account. Without this, you need to be able to su/sudo in order to build a port under development. Building ports while you are playing with them should not require privileges, either for humans, or AIs. Whether we're talking about Debian or Node or Arch or Python, everything lets you do unprivileged builds, at least while testing, and often full stop; MacPorts may not be unique, but it's certainly unusual in this regard.

Now, is this actually important? Yes. Several times lately, I've successfully fixed a port *only* because I could safely let Codex play with repeatedly building and testing via "port -L" on my machine. For example, I was waiting for months for various Haskell issues to get fixed, only no one bothered to fix them, and as a result I couldn't rebuild Pandoc, which I needed. "port -L", Codex, and a couple of hours, and I managed to have the Haskell toolchain issues with iconv fixed. I could not, however, afford to sit there typing "sudo" and hand testing it for the robot over and over again the entire time, and I could not have run the robot with root permissions. If I hadn't had "port -L", I might have finally given up and moved to another package manager.

pmetzger:
Oh, and @jmroot it is marked "draft" because I want a real discussion about the implementation (and any flaws it may have) before it gets committed.

oytech:
I only looked briefly at the code, but on concept level it looks like a good idea. I could imagine using it for testing if a change to PortGroup is working as expected and that a set of Portfiles still builds successfully without risking some undesired system wide modifications.

I used Arch Linux long time ago but I'm pretty sure that `makepkg` allows building packages without sudo (assuming installing build dependencies is not needed). I agree with @pmetzger that option of doing unprivileged package build seems like a standard feature for package manager.

doc/port.1
```diff
@@ -24,7 +24,7 @@ port \- Command line interface for MacPorts
 .SH "SYNOPSIS"
 .sp
 .nf
-\fBport\fR [\fB\-bcdfknNopqRstuvy\fR] [\fB\-D\fR \fIportdir\fR|\fIportname\fR] [\fB\-F\fR \fIcmdfile\fR] [\fIaction\fR] [\fIactionflags\fR]
+\fBport\fR [\fB\-bcdfkLnNopqRstTuvy\fR] [\fB\-D\fR \fIportdir\fR|\fIportname\fR] [\fB\-F\fR \fIcmdfile\fR] [\fIaction\fR] [\fIactionflags\fR]
```
oytech, line 27: PR discusses adding -L option but here -T is also added. Looks like unrelated change.

PR #414, opened 2026-06-15
Parse ar files for Mach-O code

pguyot:
This should fix false positive warnings about OCaml ports that do ship MachO code and are not noarch, and also warnings about ports that do but are declared noarch.


jmroot (commented):
Thanks, looks reasonable to me, but I'll request a review from Clemens as well. I think you can add `Closes: https://trac.macports.org/ticket/72437` to the commit message.

neverpanic (approved):
Looks good to me, thanks for the contribution!

neverpanic:
(You had really good timing on this one.)

pguyot:
I didn't amend the commit message before the merge, so the trac ticket needs a manual close?

neverpanic:
> I didn't amend the commit message before the merge, so the trac ticket needs a manual close?

Done.

PR #415, opened 2026-06-25
Fix possible PortIndex rsync update failure.

fhgwright:
When attempting to update the local PortIndex from the precomputed version for the current OS/arch, there is sometimes a preexisting version with the correct mtime and size, but not the correct content. With the default options, rsync fails to update it.  This change suppresses skipping based on size and time (for the index-related files, not the tarball).

This bug has no obvious user-visible effect, since the bad PortIndex fails the signature check and is then regenerated, costing extra time but not malfunctioning.  It's visible in the debug output.

TESTED:
Now correctly uses the downloaded index in a formerly failing case.

### DIFF
--- a/src/macports1.0/macports.tcl
+++ b/src/macports1.0/macports.tcl
@@ -3968,7 +3968,10 @@ proc mportsync {{options {}}} {
                         set index_source $source
                     }
                     set remote_indexdir "${index_source}PortIndex_${os_platform}_${os_major}_${os_arch}/"
-                    set rsync_commandline "$rsync_path $rsync_options $include_option $remote_indexdir $destdir"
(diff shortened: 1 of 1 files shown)


fhgwright:
@jmroot 
I don't know if you get automatically notified or not; I'm not familiar with this "workflow approval" thing.

jmroot (approved):
Thanks, looks fine to me.

fhgwright:
I have no idea where the bad version of the PortIndex was coming from.  After manually updating it to the correct version and then rerunning `port selfupdate` it would somehow get reverted to the bad one and then not get updated, with nothing in the debug output to show its generation.  So there's probably another bug, but basing the `rsync` condition on content alone seems more robust in general, and not very expensive for files of these sizes.

PR #421, opened 2026-08-17
portlib: make delete handle read-only directories

herbygillot:
file delete -force clears the read-only bit on files but not on
directories, and unlinking an entry requires write permission on the
directory containing it. It also chmods only the top-level path, not
nested ones. A tree with read-only subdirectories therefore cannot be
removed except by root, which bypasses the check.

Go's module cache creates its directories mode 0555, so this breaks the
automatic work directory wipe: when a Portfile changes, open_statefile
drops privileges and deletes workpath as the macports user, which fails.
The clean target survives only because it elevates to root first.

Take the needed permissions and retry once. Doing this rather than
elevating keeps it working for unprivileged installs, where there is no
root to fall back on, and covers any source of read-only directories
rather than one ecosystem.

Symlinks are not followed, so permissions outside the tree are untouched,
and child paths come from glob rather than file join, which would turn an
entry named ~foo into a home directory reference under Tcl 8.

Assisted-by: Claude:claude-opus-5

### DIFF
--- a/src/portlib1.0/portlib.tcl
+++ b/src/portlib1.0/portlib.tcl
@@ -1669,9 +1669,39 @@ namespace eval portlib {
             file copy {*}$args
         }
 
+        # make_deletable
+        # Add write permission to $path and every directory below it, so their
+        # entries can be unlinked. Symlinks are not followed.
+        proc make_deletable {path} {
+            # lstat, so a symlink reports as one rather than as its target.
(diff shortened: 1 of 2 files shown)


pmetzger:
The idea seems sane enough to me?

jmroot:
Would this count as a bug in `file delete`?

